PR #12, opened 2023-07-27
feat: add go-libp2p Elixir bindings

MegaRedHand:


lib/libp2p.ex
```diff
@@ -0,0 +1,111 @@
...
+  Creates a new Host.
+  """
+  def host_new() do
+    raise "NIF host_new not implemented"
+  end
+
+  @doc """
+  Deletes a Host.
+  """
+  def host_close(_host) do
```
mpaulucci, line 27: can you add `spec` to all functions?

MegaRedHand: Sure!

Makefile
```diff
@@ -1,4 +1,25 @@
...
+
+libp2p.so: $(GO_ARCHIVES) $(GO_HEADERS) libp2p/libp2p.c libp2p/utils.c
+	gcc -Wall -Werror -dynamiclib -undefined dynamic_lookup -I $(ERLANG_INCLUDES) -o libp2p.so \
+		libp2p/libp2p.c libp2p/utils.c $(GO_ARCHIVES)
+
+clean:
+	-rm $(GO_ARCHIVES) $(GO_HEADERS) libp2p.so
+
+# Compile C and Go artifacts.
+compile: libp2p.so
```
mpaulucci, line 22: the idea is that this command will compile everything (elixir code also?)?

mpaulucci (approved):

Makefile
MegaRedHand: Just the non-elixir parts (I forgot that Elixir code needs compilation 😅). We can add the Elixir parts too if needed.

mpaulucci: Maybe rename to `compile-native` or something like that to avoid confusion?

mpaulucci (approved):

lib/libp2p.ex
```diff
@@ -0,0 +1,149 @@
+defmodule Libp2p do
```
mpaulucci, line 1: I saw that rustler uses a separate `native` folder at the top to have the code that is not elixir. For example: https://github.com/rusterlium/NifIo

What do you think of following this approach?

mpaulucci: It can be seen in this blog post too: https://mainmatter.com/blog/2020/06/25/writing-rust-nifs-for-elixir-with-rustler/

MegaRedHand: Done! I moved it so it uses a structure like this:

```
lib
|-- lambda_ethereum_consensus
|   |-- application.ex
|   `-- libp2p.ex
`-- lambda_ethereum_consensus.ex
native
`-- libp2p_nif
    |-- go.mod
    |-- go.sum
    |-- libp2p.c
    |-- main.a
    |-- main.go
    |-- main.h
    |-- utils.c
    `-- utils.h
priv
`-- native
    `-- libp2p_nif.so
```

PR #14, opened 2023-08-02
docs: add consensus basic notes

### LINKED ISSUE #13: Add theory basics for fork-choice

Arkenan:
**Description**

This PR adds some notes for the basic theory of fork-choice in ethereum

Closes #13

### DIFF
--- a/docs/fork_choice.md
+++ b/docs/fork_choice.md
@@ -53,6 +53,25 @@ In general:
 
 $$W_N = B_N + \sum_i^{i \in \text{children}[N]}W_i$$
 
+## Slashing
+
+In the previous scheme, there are two rewards:
+
+- Proposer rewards, given to a proposer when their block is included in the chain. This also adds an incentive for them to try to predict the most-likely branch to be the canonical one.
+- Attester rewards, which are smaller. These are in effect if the blocks they attest to are included.
+
+These incentives, however, are not enough. To maximize their likelyhood of getting rewards, they may missbehave:
+
+- Proposers may propose a block for every current fork.
+- Attesters may attest to every current head in their local chains.
+
+These missbehaviors debilitate the protocol (they give weight to all forks) and no honest node running fork-choice would take part on them. To prevent them, nodes that are detected while doing them are slashed (punished), which means that they are excluded from the validator set and a portion of their stake is burned.
+
+In turn, nodes that provide proof of offending nodes are given a whistleblower reward. Proofs are:
+
+- For proposer slashing: two block headers in the same slot signed by the same signature.
+- For attester slashing: two attestations signed in the same slot by the same signature.
+
 ## Guarantees
 
 - Majority honest progress: if the network has over 50% nodes running this algorithm honestly, the chain progresses and each older block is exponentially more unlikely to be reverted.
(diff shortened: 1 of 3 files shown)


docs/fork_choice.md
```diff
@@ -0,0 +1,79 @@
...
+$$W_N = B_N + \sum_i^{i \in \text{children}[N]}W_i$$
+
+## Slashing
+
+In the previous scheme, there are two rewards:
+
+- Proposer rewards, given to a proposer when their block is included in the chain. This also adds an incentive for them to try to predict the most-likely branch to be the canonical one.
+- Attester rewards, which are smaller. These are given if the blocks they attest to are included.
+
+These incentives, however, are not enough. To maximize their likelyhood of getting rewards, they may missbehave:
```
MegaRedHand, line 63: Typo
```suggestion
These incentives, however, are not enough. To maximize their likelyhood of getting rewards, they may misbehave:
```

docs/fork_choice.md
```diff
@@ -0,0 +1,79 @@
...
+
+- Proposers may propose a block for every current fork.
+- Attesters may attest to every current head in their local chains.
+
+These missbehaviors debilitate the protocol (they give weight to all forks) and no honest node running fork-choice would take part on them. To prevent them, nodes that are detected while doing them are slashed (punished), which means that they are excluded from the validator set and a portion of their stake is burned.
```
MegaRedHand, line 68:
```suggestion
These misbehaviors debilitate the protocol (they give weight to all forks) and no honest node running fork-choice would take part on them. To prevent them, nodes that are detected while doing them are slashed (punished), which means that they are excluded from the validator set and a portion of their stake is burned.
```

docs/fork_choice.md
```diff
@@ -0,0 +1,79 @@
...
+
+In turn, nodes that provide proof of offending nodes are given a whistleblower reward. Proofs are:
```
MegaRedHand, line 70: This isn't actually true from what I've read. The whistleblower reward is _meant to_ be given to the node that provides the proof, but it's _actually_ given to the block proposer (source: https://github.com/ethereum/consensus-specs/issues/1631)

MegaRedHand (commented):
Left some comments.
Also, I found the main consensus doc confusing, because of the mention of the EVM and EL transactions. As I understand, in the consensus layer the beacon operations would serve as txs, while the state transition function would be the state machine. I think at some point the text should mention this.

docs/fork_choice.md
```diff
@@ -0,0 +1,79 @@
+# Fork-choice: LMD GHOST
+
+Let's separate the two parts of the name.
+
+- GHOST: Greediest, heaviest-observed sub-tree. The algorithm recognizes that by voting block $B$, we also vote its parent, block $A$, so the amount of votes of A represents the "weight" of that whole tree.
```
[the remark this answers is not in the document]
Arkenan: You're right. I changed this paragraph to explain it in a more straight-forward way. Let me know what you think

docs/fork_choice.md
Arkenan: Fixed.

docs/fork_choice.md
Arkenan: Unexpected! I went through the specs in more detail and you're right. I thought that would be a phase-0 issue, but the rationale behind this is that "offending proofs are very easy to steal by validators, so there's no point in splitting the reward".

MegaRedHand (approved):

PR #20, opened 2023-08-07
feat: add discv5 bindings

MegaRedHand:


test/libp2p_test.exs
```diff
@@ -71,4 +89,73 @@ defmodule Libp2pTest do
...
+
+    {:ok, iterator} = Libp2p.listener_random_nodes(listener)
+
+    true = Libp2p.iterator_next(iterator)
+    {:ok, node} = Libp2p.iterator_node(iterator)
+
+    {:ok, id} = Libp2p.node_id(node)
+    {:ok, addrs} = Libp2p.node_multiaddr(node)
+
+    :ok = Libp2p.peerstore_add_addrs(peerstore, id, addrs, Libp2p.ttl_permanent_addr())
```
Arkenan, line 109: Not for this PR, but it would b

test/libp2p_test.exs
```diff
@@ -71,4 +89,73 @@ defmodule Libp2pTest do
...
+
+    :ok = Libp2p.host_close(host)
+  end
```
Arkenan, line 112: Not for this PR, as what you did is perfectly fine for a think-wrapper. We should add a different layer of wrapping were a `listener_random_nodes` returns an [Stream](https://hexdocs.pm/elixir/1.12/Stream.html) elixir type that we can map/flatmap, etc over.

test/libp2p_test.exs
```diff
@@ -71,4 +89,73 @@ defmodule Libp2pTest do
...
+
+    if Libp2p.node_tcp(node) != nil do
+      {:ok, id} = Libp2p.node_id(node)
+      {:ok, addrs} = Libp2p.node_multiaddr(node)
+
+      :ok = Libp2p.peerstore_add_addrs(peerstore, id, addrs, Libp2p.ttl_permanent_addr())
+
+      case Libp2p.host_new_stream(host, id, protocol_id) do
+        {:ok, stream} ->
+          case Libp2p.stream_read(stream) do
```
Arkenan, line 129: Same here, this could be an actual elixir stream.

Arkenan (approved):
Awesome PR! I left some comments, but none of them are to resolve here. Can you start a ticket to tackle them in the future?

test/libp2p_test.exs
MegaRedHand: Thanks for the comment! I didn't know about [`Stream.resource`](https://hexdocs.pm/elixir/1.12/Stream.html#resource/3) 😮. It makes it really easy to implement the wrapper!

MegaRedHand:
Thanks for the review! Opened issue #23 to address the comments

PR #22, opened 2023-08-08
feat: add Snappy (de)compression

MegaRedHand:
This PR adds Snappy "stream" (de)compression via Rust's [_snap_](https://crates.io/crates/snap) library. Along with that, it also adds an `Enumerable` wrapper over `Libp2p.stream`, and a method to close its writing side. It's left for following PRs to expand on the wrapper, allowing a better way to write onto the stream; and to implement _true_ streaming (de)compression (currently, you need to read the whole stream before processing).


lib/lambda_ethereum_consensus/libp2p_stream.ex
```diff
@@ -0,0 +1,32 @@
+defmodule Libp2p.Stream do
```
mpaulucci, line 1: maybe we want al libp2p code in the same folder?

native/snappy/src/lib.rs
```diff
@@ -0,0 +1,33 @@
...
+    let mut encoder = read::FrameEncoder::new(&bin[..]);
+    let mut buffer = Vec::with_capacity(bin.len());
+    encoder
+        .read_to_end(&mut buffer)
+        .map_err(|e| e.to_string())?;
+    Ok(bytes_to_binary(env, &buffer))
+}
+
+#[rustler::nif]
+fn decompress<'env>(env: Env<'env>, bin: Binary) -> Result<Binary<'env>, String> {
```
mpaulucci, line 24: very nice! BTW have you looked into the different types of Binary there is? https://docs.rs/rustler/latest/rustler/types/binary/index.html I don't completely understand the differences between them but just wanted to show this in case you haven't seen it before

MegaRedHand: Yeah. AFAIK the `Binary` would be a read-only string that's owned by the calling process, the `OwnedBinary` is a read-write string that's owned by the NIF and isn't associated with a VM process (an "env"), and `NewBinary` is just a convenience, which works like an `OwnedBinary` (it's read-write) but already associated to a process.

mpaulucci (approved):

PR #26, opened 2023-08-10
feat: add EL <> CL Engine API Wrapper

### LINKED ISSUE #25: Create EL <> CL communication wrapper for Engine API
https://github.com/ethereum/execution-apis/tree/main

phklive:
**Motivation**

The goal is to create a wrapper to enable EL <> CL communication through the Engine API. This would be the interface of the Elixir client communicating with the Engine API.

**Description**

- Implemented generation of JWT ( needed for auth )
- Implemented creation of generic RPC call
- Created 2 endpoints ( 1 functional, 1 non-functional as example ) 
- Added response validation / error-handling 

Closes #25


lib/lambda_ethereum_consensus/engine/execution.ex
```diff
@@ -0,0 +1,35 @@
...
+  alias LambdaEthereumConsensus.RPC
+
+  # Default Execution Layer endpoint
+  @execution_engine_endpoint "http://0.0.0.0:8551"
+
+  # Default Execution Layer RPC version
+  @execution_engine_rpc_version "2.0"
+
+  @spec engine_exchange_capabilities(map) :: {:error, any} | {:ok, any}
+  def engine_exchange_capabilities(params \\ %{"capabilities" => []}) do
```
mpaulucci, line 11: I wouldn't add a default params, instead we should validate that the params have the format that we expect.

phklive: Is it by adding guard clauses OR function definition params OR other ?

Guard clauses like this: 
```
def call(method, endpoint, version, params) when is_binary(method) do
end
```

Function definitions like this: 
```
def call({:ok, data}, data) do
end
```

These are examples

lib/lambda_ethereum_consensus/jwt.ex
```diff
@@ -0,0 +1,25 @@
...
+
+  @spec generate_token :: {:error, atom | keyword} | {:ok, binary, %{optional(binary) => any}}
+  def generate_token() do
+    claim = %{"iat" => Joken.current_time()}
+
+    signer =
+      Joken.Signer.create(
+        "HS256",
+        Base.decode16!(
+          String.upcase("c4224413b9f85af9a9d1f07962493e646cf7d79f76333bd0a8b901affb558f84")
```
Arkenan, line 19: Instead of hardcoding this token, let's:

1. In `config.ex`, let's write the contents of the JWT file and put it in a config.
2. In here, let's just `Application.get_env(LambdaEthereumConsensus, JWT)` to get it. 

This way, we de-hardcode it, but also we don't read the file each time we generate the token.

lib/lambda_ethereum_consensus/engine/execution.ex
mpaulucci: You can deconstruct them like
`{"capabilities" => capabilities } = params` so you ensure that the params map has the fields that you are interesed in and you also don't do a passthrough

Arkenan: We may just go

```elixir
@typespec engine_exchange_capabilities(list) :: {:ok, any} | {:error, any} 
def engine_exchange_capabilities(capabilities \\ []) do
```

ONLY IF an empty array is a valid param. If not:

```elixir
@typespec engine_exchange_capabilities(list) :: {:ok, any} | {:error, any} 
def engine_exchange_capabilities(capabilities) do
```

And then we can just build the map in the call. In general, when building an API wrapper it's a good idea to pass actual parameters instead of a map. Specially because the idea is to make this functions easier to understand and call from other parts of the project and arguments make that clear.

Arkenan (changes requested):
Thanks Paul! Awesome work on kickstarting this API! Added a couple of small comments.

lib/lambda_ethereum_consensus/engine/execution.ex
phklive: In this instance an empty map would be a valid param for `engine_exchangeCapabilities` but wouldn't be right for other methods.

Do you recommend to not use default params at all then ?

lib/lambda_ethereum_consensus/jwt.ex
```diff
@@ -0,0 +1,25 @@
...
+  @spec generate_token :: {:error, atom | keyword} | {:ok, binary, %{optional(binary) => any}}
+  def generate_token() do
+    claim = %{"iat" => Joken.current_time()}
+
+    signer =
+      Joken.Signer.create(
+        "HS256",
+        Base.decode16!(
+          String.upcase(Application.fetch_env!(:lambda_ethereum_consensus, :jwt_secret))
+        )
```
Arkenan, line 20: nit and fully optional: a more readable version of this using pipes would be:

```elixir
Application.fetch_env!(:lambda_ethereum_consensus, :jwt_secret)
|> String.upcase()
|> Base.decode16!()
```

Arkenan (approved):
Thanks Paul, awesome PR! I added an optional comment, but feel free to merge it as is.

lib/lambda_ethereum_consensus/engine/execution.ex
```diff
@@ -0,0 +1,26 @@
...
+
+    case RPC.call(
+           "engine_exchangeCapabilities",
+           @execution_engine_endpoint,
+           @execution_engine_rpc_version,
+           params
+         ) do
+      {:ok, result} -> RPC.validate_response(result)
+      {:error, error} -> {:error, error}
+    end
```
MegaRedHand, line 24: Today I learned that we can use [`with`](https://elixir-lang.org/getting-started/mix-otp/docs-tests-and-with.html#with) for mapping `:ok` results:

```suggestion
    with {:ok, result} <- RPC.call(
           "engine_exchangeCapabilities",
           @execution_engine_endpoint,
           @execution_engine_rpc_version,
           params
         ) do
      RPC.validate_response(result)
    end
```

phklive: But then how do i manage the error case ?

MegaRedHand: You can't use it for that, but in this case, you aren't actually handling the error but just propagating it upwards. With `with`, if the match fails, the non-matching result is returned, and so the behaviour is _mostly_ the same as with `case`.

Issue #11, opened 2023-07-27
Add CI task that runs linter (credo) type checks (dialyzer) and tests


Arkenan:
Depends on #9 and #10.

mpaulucci:
I think we can close this one, reopen if needed

PR #29, opened 2023-08-11
ci: check format and run credo

### LINKED ISSUE #9: Add credo
We need linting with credo and a make target that runs it.

MegaRedHand:
Closes #9 #11

This PR adds credo to the Makefile and CI (+ a `mix format` check), and fixes a warning.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,8 +42,8 @@ jobs:
         if: steps.output-cache.outputs.cache-hit != 'true'
         run: make compile-native
 
-  build:
-    name: Build
+  dialyzer:
+    name: Dialyzer
     needs: compile-native
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -42,8 +42,8 @@ jobs:
         if: steps.output-cache.outputs.cache-hit != 'true'
         run: make compile-native
 
-  build:
-    name: Build
+  dialyzer:
```
mpaulucci, line 45: I think it makes sense to keep dializer as part of the "build" since you could think of it as a compile time check. So build = mix compile + dialyzer.

WDYT?

MegaRedHand: We aren't currently running `mix compile` here, and I'm not sure I understand the difference. Isn't it the same as `mix dialyzer --plt`?

Makefile
```diff
@@ -47,3 +47,7 @@ deps:
 # Run tests
 test: compile-native
 	mix test
+
+lint: compile-native
```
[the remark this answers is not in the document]
MegaRedHand: My bad. Thought credo would need to load the NIFs

.github/workflows/ci.yml
mpaulucci: Ah ok. My point is that I think we should keep the naming of the checks more high level, like "build" or "compile" instead of dialyzer, and "lint " instead of credo. Credo, dialyzer are just specific tools

MegaRedHand: I see. That should simplify later changes to the tools we use.

mpaulucci (approved):

Issue #16, opened 2023-08-02
Add ability to run EF spec tests


mpaulucci:
Language agnostic consensus spec tests can be downloaded from https://github.com/ethereum/consensus-spec-tests/releases . We can start with the `minimal` preset

Here is the information about how to interpret them: https://github.com/ethereum/consensus-specs/tree/dev/tests/formats

I would start by adding support for the `SSZ static types`: https://github.com/ethereum/consensus-specs/blob/dev/tests/formats/ssz_static/core.md which serializes/deserializes ssz files.

Here is an example of how Lighthouse does it: https://github.com/sigp/lighthouse/blob/stable/testing/ef_tests/src/cases/ssz_static.rs

PR #32, opened 2023-08-11
test: add ability to run EF spec tests

MegaRedHand:
Closes: #16


test/test_helper.exs
```diff
@@ -1 +1,60 @@
...
+          config == pinned_config do
+        test_name = "c:#{config} f:#{fork} r:#{runner} h:#{handler} s:#{suite} -> #{cse}"
+
+        test_runner = Map.get(SpecTestUtils.runner_map(), runner)
+
+        @tag :spectest
+        @tag config: config
+        @tag fork: fork
+        @tag runner: runner
+        @tag suite: suite
```
mpaulucci, line 46: very cool!

test/test_helper.exs
```diff
@@ -1 +1,60 @@
...
+        if test_runner == nil do
+          @tag :skip
```
mpaulucci, line 48: This is really cool, but I think we need a more fine grained option to skip/include tests, so we can start adding them one by one, or runner by runner.

mpaulucci (approved):
👏

test/spec-test/runners/ssz_test.exs
```diff
@@ -0,0 +1,22 @@
+defmodule SSZTestRunner do
```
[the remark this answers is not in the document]
MegaRedHand: I like the idea!

test/test_helper.exs
mpaulucci: sounds good! Would it work with the @skip tag like the tests that don't have a runner?

MegaRedHand: Yeah!

mpaulucci: nice!

PR #33, opened 2023-08-15
feat: add NIF to serialize and deserialize ssz objects

mpaulucci:
**Motivation**
We need a way to (de)serialize SSZ objects which is the main serialization tool in the ethereum client

**Description**
This is a proof of concept and has a lot of room for improvement. One thing that we need to do is making it more generic so we can serialize and deserealize many types of objects.


Arkenan (approved):
As a PoC for rustler use this looks great, although I think we should find a way to move away from lighthouse types.

MegaRedHand (approved):

PR #39, opened 2023-08-16
refactor(ssz): reduce SSZ boilerplate by using macros

MegaRedHand:
This PR modifies the code in the SSZ nif to use declarative macros, that reduce the amount of manual code needed to add a new type. It also adds the `Fork` and `ForkData` types as usage examples.


test/ssz_test.exs
```diff
@@ -1,17 +1,39 @@
...
       epoch: 12_345,
-      root:
-        <<1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
-          0, 0, 1>>
+      root: Base.decode16!("0100000000000000000000000000000000000000000000000000000000000001")
     }
 
-    {:ok, encoded} = LambdaEthereumConsensus.Ssz.to_ssz(value)
-    {:ok, decoded} = LambdaEthereumConsensus.Ssz.from_ssz(encoded)
+    assert_roundtrip(Checkpoint, value)
```
mpaulucci, line 17: TIL: you can specify an atom like that, and that modules are atoms.

That being said, I don't think this is very idiomatic, it could be confusing. Is it difficult to be able to write an atom in lower case (i.e :fork_choice) that is then mapped to a pascal case struct in Rust?

mpaulucci:
awesome work!

test/ssz_test.exs
mpaulucci: Ah that would be amazing!

mpaulucci (approved):

PR #40, opened 2023-08-16
test(ssz): add runner for `ssz_static` spec tests

MegaRedHand:


test/spec/runners/ssz_static.ex
```diff
@@ -0,0 +1,55 @@
...
+    assert {:ok, decompressed} = :snappyer.decompress(compressed)
+
+    expected =
+      YamlElixir.read_from_file!(case_dir <> "/value.yaml")
+      |> Stream.map(&parse_yaml/1)
+      |> Map.new()
+
+    expected_root = YamlElixir.read_from_file!(case_dir <> "/roots.yaml")
+
+    assert_ssz(testcase, decompressed, expected, expected_root)
```
mpaulucci, line 33: Can we include the Elixir struct here somewhere?

like: decompressed = %Checkpoint{}

mpaulucci: So that we ensure that the Elixir structs are respecting the schema.

MegaRedHand: I don't know if this is what you mean, but I intend to add the Elixir-side structs and integrate that with the SSZ lib and the test runner in another PR.

mpaulucci: sounds good

mpaulucci (approved):

PR #45, opened 2023-08-17
feat: add beacon-chain spec containers

MegaRedHand:


test/spec/utils/testcase.ex
```diff
@@ -6,7 +6,7 @@ defmodule SpecTestCase do
   defstruct [:config, :fork, :runner, :handler, :suite, :case]
 
   def new([config, fork, runner, handler, suite, cse]) do
```
mpaulucci, line 8: do we use this function anywhere?

mpaulucci (approved):

PR #48, opened 2023-08-22
test: failing indexed attestation

MegaRedHand:
** DO NOT MERGE **

This PR adds a failing test for `IndexedAttestation` de/serialization.


lib/ssz_types/historical_batch_minimal.ex
```diff
@@ -0,0 +1,20 @@
+defmodule SszTypes.HistoricalBatchMinimal do
```
mpaulucci, line 1: should we have one file "historical_batch" per type and add the two modules there?

MegaRedHand: Yeah, I think that would be better.

PR #84, opened 2023-08-24
test: add gossipsub real node network example

MegaRedHand:
This PR adds an example (intended as a PoC, not as a CI test) that uses gossipsub to print attestations published by peers.


test/unit/libp2p_test.exs
```diff
@@ -2,6 +2,7 @@ defmodule Libp2pTest do
   use ExUnit.Case
   doctest Libp2p
 
+  # Mainnet
   @bootnodes [
```
mpaulucci, line 6: If this is an external ip then this should probably go into "test/integration", right?

MegaRedHand: Those are the Mainnet bootnodes' addresses.

test/integration/libp2p_test.exs
```diff
@@ -0,0 +1,172 @@
+defmodule Libp2pTest do
+  use ExUnit.Case
+  doctest Libp2p
+
+  # Mainnet
+  @bootnodes [
```
mpaulucci, line 6: maybe we want to move these to config? https://hexdocs.pm/elixir/main/Config.html

We can address this later though.

mpaulucci (approved):

PR #97, opened 2023-08-29
refactor: drop lighthouse types

MegaRedHand:
Since working with lighthouse containers was more trouble than is worth, we are dropping them and duplicating struct definition (one for Rustler traits, and another for SSZ traits), and using the existing `FromElx` and `FromLH` (now `FromSsz`) traits for conversion.


mpaulucci (approved):
🎉

PR #107, opened 2023-08-29
feat: initiate peer discovery in the application

MegaRedHand:


mpaulucci (approved):

phklive:
I am curious by adding the module to the children of the application.ex, it automatically starts the peer discovery module as a process right ? 

Which means that running iex -S mix launches it right ?

MegaRedHand:
Yes, that's right

PR #119, opened 2023-08-30
build: ignore running ci when changing only markdown files.

mpaulucci:
**Motivation**
We don't want to waste resources and time by running the CI when only .md files change

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,8 +3,12 @@ name: CI
 on:
   push:
     branches: [ main ]
+    paths-ignore:
+      - '**/*.md'
   pull_request:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -3,8 +3,12 @@ name: CI
 on:
   push:
     branches: [ main ]
+    paths-ignore:
+      - '**/*.md'
```
MegaRedHand, line 7: I think we should leave it on for `main`, since it doesn't really speed things up.

mpaulucci: ok

MegaRedHand (approved):

PR #123, opened 2023-08-30
feat: add gossip consumer component

MegaRedHand:


lib/lambda_ethereum_consensus/p2p/subscriber.ex
```diff
@@ -0,0 +1,33 @@
+defmodule LambdaEthereumConsensus.Subscriber do
```
mpaulucci, line 1: I think in Elixir we tend to follow the folder structure when naming modules. So this one would be `LambdaEthereumConsensus.P2P.Subscriber`

lib/lambda_ethereum_consensus/p2p/gossip_consumer.ex
```diff
@@ -0,0 +1,55 @@
+defmodule LambdaEthereumConsensus.P2P.GossipConsumer do
+  @moduledoc """
+  This module consumes events created by Subscriber.
+  """
+  use Broadway
+
+  def start_link(%{gsub: gsub, topic: topic_name, type: _, handler: _} = opts)
```
mpaulucci, line 7: nit: `s/type/ssz_type` ?

mpaulucci (approved):

PR #129, opened 2023-08-31
feat: add BLS `sign` function.

### LINKED ISSUE #128: Implement BLS `sign` function
We want to just use an existing implementation for now.

I suggest that we create a `bls_nif` with `rustler` and we use the same Rust lib that lighthouse is using for now.

Here is the `bls` spec test for reference: https://github.com/sigp/lighthouse/blob/stable/testing/ef_tests/src/cases/bls_sign_msg.rs

Acceptance criteria:`sign` test case on bls spec should pass: https://github.com/lambdaclass/lambda_ethereum_consensus/blob/main/test/spec/runners/bls.ex#L63-L65

mpaulucci:
**Description**
Add bls `sign` function by using the lighthouse implementation

Closes #128


test/spec/runners/bls.ex
```diff
@@ -60,8 +60,18 @@ defmodule BLSTestRunner do
...
+    %{message: message, privkey: private_key} = input
+
+    case output do
+      nil ->
+        {result, _error_msg} = Bls.sign(private_key, message)
+        assert result == :error
+
+      output ->
+        {:ok, signature} = Bls.sign(private_key, message)
+        assert signature == output
```
MegaRedHand, line 73: Prepending `assert` on match assignments improves error messages.
```suggestion
      nil ->
        assert {result, _error_msg} = Bls.sign(private_key, message)
        assert result == :error

      output ->
        assert {:ok, signature} = Bls.sign(private_key, message)
        assert signature == output
```

native/bls_nif/src/lib.rs
```diff
@@ -0,0 +1,31 @@
...
+    binary.into()
+}
+
+#[rustler::nif]
+fn sign<'env>(
+    env: Env<'env>,
+    private_key: Binary,
+    message: Binary,
+) -> Result<Binary<'env>, String> {
+    let sk = match SecretKey::deserialize(&private_key.as_slice()) {
```
MegaRedHand, line 20: Nit
```suggestion
    let sk = match SecretKey::deserialize(private_key.as_slice()) {
```

MegaRedHand (approved):

MegaRedHand:
We should also add the crate to the `fmt` target in the _Makefile_

lib/bls.ex
```diff
@@ -0,0 +1,27 @@
+defmodule Bls do
+  @moduledoc """
+  BLS signature verification.
+  """
+  use Rustler, otp_app: :lambda_ethereum_consensus, crate: "bls_nif"
+
+  @spec sign(binary(), binary()) :: binary()
```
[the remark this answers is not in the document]
mpaulucci: good catch

Issue #131, opened 2023-09-01
Implement first endpoint `/eth/v1/beacon/states/{state_id}/root` for Beacon API


phklive:
Following the Beacon API specs we need to implement the first endpoint enabling us to query the root of the latest Beacon state using `/eth/v1/beacon/states/{state_id}/root`

mpaulucci:
Closed by https://github.com/lambdaclass/lambda_ethereum_consensus/pull/218

PR #139, opened 2023-09-01
feat: added phoenix boilerplate + first endpoint

### LINKED ISSUE #130: Implement phoenix framework server boilerplate
https://hexdocs.pm/phoenix/overview.html

Phoenix is an Elixir framework enabling to build MVC servers. 

We will use it to create the Beacon API for the consensus client.

phklive:
**Motivation**

The aim is to create the boilerplate for a phoenix server that we will use to implement our Beacon API, furthermore I have added the first endpoint that we will use in phase 1 of our roadmap to query the root of our latest beaconState.

**Description**

Added phoenix, removed boilerplate code, implemented first endpoint

Closes #130 #131


lib/beacon_api/controllers/v1/beacon_controller.ex
```diff
@@ -0,0 +1,11 @@
+defmodule BeaconApi.V1.BeaconController do
+  use BeaconApi, :controller
+
+  @spec get_state_root(Plug.Conn.t(), any) :: Plug.Conn.t()
```
mpaulucci, line 4: we should handle 400, 404, etc.. for this scenario

lib/beacon_api/router.ex
```diff
@@ -0,0 +1,29 @@
...
+  # Ethereum API Version 1
+  scope "/eth/v1", BeaconApi.V1 do
+    pipe_through(:api)
+
+    scope "/beacon" do
+      get("/states/:state_id/root", BeaconController, :get_state_root)
+    end
+  end
+
+  # Ethereum API Version 2
```
mpaulucci, line 17: don't think we need this now

lib/beacon_api/controllers/v1/beacon_controller.ex
```diff
@@ -0,0 +1,11 @@
...
+  def get_state_root(conn, params) do
+    json(conn, %{
+      id: Map.get(params, "state_id"),
+      root: "0xbe41d3394dc8c461ab7049dbedb8944613ada8bd1743e091948f4d7b5ca8af36"
+    })
+  end
```
Arkenan, line 10: I don't think this is the format the response expects. The example in the beacon API swagger is:

```json
{
  "execution_optimistic": false,
  "finalized": false,
  "data": {
    "root": "0xcf8e0d4e9587369b2301d0790347320302cc0943d5a1884560367e8208d920f2"
  }
}
```

It's true that we don't have much of that info available yet, but even if we're hardcoding an answer, we should make the format compliant. That is, for example:

```elixir
json(conn, %{
  execution_optimistic: true,
  finalized: false,
  data: %{
    root: "0xbe41d3394dc8c461ab7049dbedb8944613ada8bd1743e091948f4d7b5ca8af36"
  }
})
```

Arkenan (changes requested):

lib/beacon_api/controllers/v1/beacon_controller.ex
Arkenan: If we add this in this PR, we may take the following strategy:
1. If asking for `state_id in ["head", "genesis", "finalized", "justified"]` then return the hardcoded block root.
3. If asking for the hardcoded block using the root, return the hardcoded block root.
4. If asking for any other block using a root or a number, return a 404.
5. If state_id is not a number, not in block root format (0x and 32 bytes) or one of the checkpoints/head/genesis, then return 400.

Regarding 5, I think it may be easier to implement it outside of this PR, after integrating with OpenApiSpex (I'm adding an issue for that right now).

lib/beacon_api/router.ex
phklive: We will need it to implement the v2 endpoints of the beacon api. 
Should I:
- Remove it for now and add back when we implement the endpoints
- Leave it empty for boilerplate so we know how and where to implement

lib/beacon_api/controllers/v1/beacon_controller.ex
phklive: Modified

lib/beacon_api/controllers/v1/beacon_controller.ex
phklive: What is 5 in "Regarding 5" ?

mpaulucci:
I think https://github.com/lambdaclass/lambda_ethereum_consensus/issues/165 should probably be part of this task?

lib/beacon_api/router.ex
mpaulucci: I would just keep the endpoints that we implement and remove boilerplate. I think it's trivial where to implement.

phklive: Ok

phklive (commented):
Added requested changes

mpaulucci:
@Arkenan was talking to someone about the fact that maybe Phoenix is an overkill and we should just use `plug` instead. I would check with him before merging this

phklive:
> @Arkenan was talking to someone about the fact that maybe Phoenix is an overkill and we should just use `plug` instead. I would check with him before merging this

Could definitely be, will wait for your expertise

Arkenan (approved):
LGTM, left a couple of nit comments.

lib/beacon_api/error_json.ex
```diff
@@ -0,0 +1,10 @@
+defmodule BeaconApi.ErrorJSON do
+  use BeaconApi, :controller
+
+  @spec render(any, any) :: %{message: <<_::184>>}
```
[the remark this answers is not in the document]
phklive: modified

lib/beacon_api/controllers/v1/beacon_controller.ex
```diff
@@ -1,11 +1,43 @@
...
   @spec get_state_root(Plug.Conn.t(), any) :: Plug.Conn.t()
   def get_state_root(conn, params) do
-    json(conn, %{
-      id: Map.get(params, "state_id"),
-      root: "0xbe41d3394dc8c461ab7049dbedb8944613ada8bd1743e091948f4d7b5ca8af36"
-    })
+    state_id =
+      Map.get(params, "state_id")
+
+    case state_id in [
```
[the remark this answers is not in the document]
phklive: modified, makes a lot of sense.

PR #143, opened 2023-09-04
refactor: move special encoding/decoding rules to the types themselves.

mpaulucci:
**Motivation**
We want the custom rules to be as closer to the type definition so that we don't bloat the Ssz module. Addionally, if we have a type A with special encoding/decoding rules that is contained inside type B, we shouldn't have to define the rule for type B, it should work out of the box

### DIFF
--- a/lib/ssz.ex
+++ b/lib/ssz.ex
@@ -31,15 +31,13 @@ defmodule Ssz do
 
   # Ssz types can have special decoding rules defined in their optional encode/1 function,
   defp encode(%name{} = struct) do
-    case function_exported?(name, :encode, 1) do
-      true ->
-        name.encode(struct)
-
-      false ->
-        struct
-        |> Map.from_struct()
-        |> Enum.map(fn {k, v} -> {k, encode(v)} end)
-        |> then(&struct!(name, &1))
+    if function_exported?(name, :encode, 1) do
+      name.encode(struct)
+    else
+      struct
+      |> Map.from_struct()
+      |> Enum.map(fn {k, v} -> {k, encode(v)} end)
+      |> then(&struct!(name, &1))
     end
   end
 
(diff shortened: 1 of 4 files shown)


test/spec/utils/generator.ex
```diff
@@ -52,8 +52,11 @@ defmodule SpecTestGenerator do
         test_runner = Map.get(SpecTestGenerator.runner_map(), testcase.runner)
 
         @tag :spectest
-        @tag config: testcase.config, fork: testcase.fork
-        @tag runner: testcase.runner, suite: testcase.suite
+        @tag config: testcase.config,
+             fork: testcase.fork,
+             runner: testcase.runner,
+             handler: testcase.handler,
```
mpaulucci, line 58: added handler filter

MegaRedHand: Good catch

lib/ssz.ex
```diff
@@ -29,31 +29,43 @@ defmodule Ssz do
...
+    case function_exported?(name, :encode, 1) do
+      true ->
+        name.encode(struct)
+
+      false ->
+        struct
+        |> Map.from_struct()
+        |> Enum.map(fn {k, v} -> {k, encode(v)} end)
+        |> then(&struct!(name, &1))
+    end
```
MegaRedHand, line 43: Since the matched value is a bool, we can use an `if-else` here instead of a `case`.

lib/ssz.ex
```diff
@@ -29,31 +29,43 @@ defmodule Ssz do
...
+    case function_exported?(name, :decode, 1) do
+      true ->
+        name.decode(struct)
+
+      false ->
+        struct
+        |> Map.from_struct()
+        |> Enum.map(fn {k, v} -> {k, decode(v)} end)
+        |> then(&struct!(name, &1))
+    end
```
MegaRedHand, line 59: Same

MegaRedHand (approved):

PR #145, opened 2023-09-04
feat: add BLS `fast_aggregate_verify` function

### LINKED ISSUE #133: Implement `Bls.fast_aggregate_verify` function
Implementation can be inspired in: https://github.com/lambdaclass/lambda_ethereum_consensus/pull/129

Acceptance criteria: Spec tests for bls `fast_aggregate_verify` should pass.

f3r10:
Description
Add bls fast_aggregate_verify function by using the lighthouse implementation

Closes #133


native/bls_nif/src/lib.rs
```diff
@@ -28,4 +28,30 @@ fn sign<'env>(
...
+        .iter()
+        .map(|pkb| match PublicKey::deserialize(pkb.as_slice()) {
+            Ok(pk) => Ok(pk),
+            Err(e) => return Err(format!("{:?}", e)),
+        })
+        .collect::<Result<Vec<PublicKey>, _>>();
+    let pubkeys = match pubkeys_result {
+        Ok(pubkeys) => pubkeys,
+        Err(e) => return Err(format!("{:?}", e)),
+    };
```
MegaRedHand, line 51: This can be done with a `map_err` and a `?`:

```suggestion
    let pubkeys = pubkeys_result.map_err(|e| format!("{:?}", e))?;
```

native/bls_nif/src/lib.rs
```diff
@@ -28,4 +28,30 @@ fn sign<'env>(
...
+    let aggregate_sig = match AggregateSignature::deserialize(signature.as_slice()) {
+        Ok(aggregate_sig) => aggregate_sig,
+        Err(e) => return Err(format!("{:?}", e)),
+    };
+    let pubkeys_result = public_keys
+        .iter()
+        .map(|pkb| match PublicKey::deserialize(pkb.as_slice()) {
+            Ok(pk) => Ok(pk),
+            Err(e) => return Err(format!("{:?}", e)),
+        })
```
MegaRedHand, line 46: This match shouldn't be needed, since the error is later converted to a `String` anyways.

```suggestion
        .map(|pkb| PublicKey::deserialize(pkb.as_slice()))
```

test/spec/runners/bls.ex
```diff
@@ -76,8 +76,14 @@ defmodule BLSTestRunner do
...
 
-  defp assert_fast_aggregate_verify(_input, _output) do
-    assert false
+  defp assert_fast_aggregate_verify(
+         %{message: message, pubkeys: pubkeys, signature: signature},
+         output
+       ) do
+    case Bls.fast_aggregate_verify(pubkeys, message, signature) do
+      {:ok, confirmation} -> assert confirmation == output
+      {result, _error_msg} -> assert result == :error
```
MegaRedHand, line 85: We should add the error message as the assert's message, to improve debuggability:
```suggestion
      {result, error_msg} -> assert result == :error, error_msg
```

native/bls_nif/src/lib.rs
```diff
@@ -28,4 +28,30 @@ fn sign<'env>(
...
+#[rustler::nif]
+fn fast_aggregate_verify<'env>(
+    public_keys: Vec<Binary>,
+    message: Binary,
+    signature: Binary,
+) -> Result<bool, String> {
+    let aggregate_sig = match AggregateSignature::deserialize(signature.as_slice()) {
+        Ok(aggregate_sig) => aggregate_sig,
+        Err(e) => return Err(format!("{:?}", e)),
+    };
```
MegaRedHand, line 40: This can be turned into a `map_err`

MegaRedHand (commented):
Good work! Left some comments about possible improvements.

native/bls_nif/src/lib.rs
f3r10: thanks, you are right with a `map_err` is much better.

test/spec/runners/bls.ex
```diff
@@ -76,8 +76,14 @@ defmodule BLSTestRunner do
...
-  defp assert_fast_aggregate_verify(_input, _output) do
-    assert false
+  defp assert_fast_aggregate_verify(
+         %{message: message, pubkeys: pubkeys, signature: signature},
+         output
+       ) do
+    case Bls.fast_aggregate_verify(pubkeys, message, signature) do
+      {:ok, confirmation} -> assert confirmation == output
+      {result, error_msg} -> assert result == :error, error_msg
+    end
```
MegaRedHand, line 86: Since we only care about the _happy path_, this can be simplified using plain matches inside an assert. Example from the `assert_sign` function:
```elixir
assert {:ok, signature} = Bls.sign(private_key, message)
assert signature == output
```

f3r10: oh ok :+1:

f3r10: I tried, but there were four failed spec tests. 
For example, for `fast_aggregate_verify_tampered_signature_5e745ad0c6199a6c` the output is `false`  but the test fails with an `{:error, "BlstError(BLST_POINT_NOT_ON_CURVE)"}`. 
I think this error is generated by the `deserialize` methods used in the rust native part.

MegaRedHand: My bad! Then we should check depending on the test's `output`. If the expected output is `false` we should check it returns either `{:ok, false}` or `{:error, reason}`, but if it's `true`, then it should always return `{:ok, true}`.

f3r10: maybe something  like this?:  
```
if output do
      assert {:ok, confirmation} = Bls.fast_aggregate_verify(pubkeys, message, signature)
      assert confirmation == output
else
  case Bls.fast_aggregate_verify(pubkeys, message, signature) do
    {:ok, failed_confirmation} -> assert failed_confirmation == false
    {result, error_msg} -> assert result == :error, error_msg
  end
end
```

MegaRedHand: I think it'll be simpler if we keep the single `case`, but change the match arms:
```elixir
{:ok, true} -> assert output
{:ok, false} -> assert not output
{:error, reason} -> assert not output, reason
```

f3r10: fantastic, much clearer.

For the error case, would it be also necessary to check that is indeed an :error?
``` elixir
{result, reason} ->
  assert not output, reason
  assert result == :error
```

MegaRedHand: No need, since we match against `{:error, reason}`, and not `{result, reason}`. Then if we try to match against a `{:non_error, "something"}`, for example, the test will fail because no match arm matched inside the `case`.

MegaRedHand (approved):
🚀 🚀 🚀

PR #162, opened 2023-09-05
feat: add component for handling Req/Resp requests

MegaRedHand:


lib/lambda_ethereum_consensus/p2p/req_resp.ex
```diff
@@ -0,0 +1,89 @@
+defmodule LambdaEthereumConsensus.P2P.ReqRespHandler do
```
mpaulucci, line 1: this is for incoming requests, right?

MegaRedHand: Yes. Do you think `RequestHandler` would be better?

mpaulucci: IncomingRequestHandler maybe?

mpaulucci (approved):

PR #168, opened 2023-09-06
refactor: parameterize SSZ Rust types to allow mainnet vs minimal config.

mpaulucci:
**Motivation**
Simplify the ssz types that have different configurations (minimal vs mainnet)

**Description**
Created `Mainnet` and `Minimal ` configs and parametrized the struct that use the config values.


MegaRedHand (approved):
🔥

PR #169, opened 2023-09-06
chore: add all configurations for mainnet / minimal in SSZ crate.

mpaulucci:


native/ssz_nif/src/ssz_types/beacon_chain.rs
```diff
@@ -100,13 +100,11 @@ pub(crate) struct HistoricalSummary {
 }
 
 #[derive(Encode, Decode)]
-pub(crate) struct DepositBase<C: Config> {
-    pub(crate) proof: FixedVector<Bytes32, C::DepositContractTreeDepth>,
+pub(crate) struct Deposit {
+    pub(crate) proof: FixedVector<Bytes32, typenum::U33>,
```
mpaulucci, line 104: this is the way lighouse does it, because the number is DepositContractTreeDepth + 1.

MegaRedHand, line 104: We should add a comment

MegaRedHand (approved):

native/ssz_nif/src/ssz_types/beacon_chain.rs
mpaulucci: added

PR #179, opened 2023-09-07
feat: add BLS `eth_fast_aggregate_verify` function

### LINKED ISSUE #137: Implement `Bls.eth_fast_aggregate_verify` function
Implementation can be inspired in: https://github.com/lambdaclass/lambda_ethereum_consensus/pull/129

Acceptance criteria: Spec tests for bls `eth_fast_aggregate_verify` should pass.

f3r10:
Description
Add bls eth_fast_aggregate_verify function by using the lighthouse implementation

Closes #137

### DIFF
--- a/lib/bls.ex
+++ b/lib/bls.ex
@@ -18,6 +18,11 @@ defmodule Bls do
     :erlang.nif_error(:nif_not_loaded)
   end
 
+  @spec eth_fast_aggregate_verify([binary()], binary(), binary()) :: {:ok, boolean} | {:error, any()}
+  def eth_fast_aggregate_verify(_public_keys, _message, _signature) do
+    :erlang.nif_error(:nif_not_loaded)
+  end
+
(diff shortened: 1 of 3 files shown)


lib/bls.ex
```diff
@@ -18,6 +18,12 @@ defmodule Bls do
     :erlang.nif_error(:nif_not_loaded)
   end
 
+  @spec eth_fast_aggregate_verify([binary()], binary(), binary()) ::
+          {:ok, boolean} | {:error, any()}
```
MegaRedHand, line 22: Rust's `String` is mapped into an Elixir binary
```suggestion
          {:ok, boolean} | {:error, binary()}
```

MegaRedHand (approved):
LGTM 👍

lib/bls.ex
f3r10: ok :+1: , so it is also necessary to fix `fast_aggregate_verify` function

PR #195, opened 2023-09-11
fix: missing comparison in generator's `__mix_recompile__?`

MegaRedHand:
For context, see [this example](https://hexdocs.pm/mix/1.15.5/Mix.Tasks.Compile.Elixir.html#module-__mix_recompile__-0)


test/spec/utils/generator.ex
```diff
@@ -41,7 +41,7 @@ defmodule SpecTestGenerator do
       # Recompile module only if corresponding dir layout changed
       def __mix_recompile__? do
         Path.wildcard(unquote("tests/#{pinned_config}/#{pinned_fork}/**"))
-        |> :erlang.md5()
+        |> :erlang.md5() == unquote(paths_hash)
```
mpaulucci, line 44: shouldn't it be `!=`?

MegaRedHand: Yes 😵‍💫

mpaulucci (approved):

PR #196, opened 2023-09-11
feat: add `SignedAggregateAndProof` SSZ container

MegaRedHand:
Depends on #188

This sets up the file structure for container types from _validator.md_, and enables us to deserialize aggregate signatures from the `beacon_aggregate_and_proof` gossipsub topic.


lib/lambda_ethereum_consensus/p2p/gossipsub.ex
```diff
@@ -19,9 +19,8 @@ defmodule LambdaEthereumConsensus.P2P.GossipSub do
       {"voluntary_exit", SszTypes.SignedVoluntaryExit},
       {"proposer_slashing", SszTypes.ProposerSlashing},
       {"attester_slashing", SszTypes.AttesterSlashing},
-      {"bls_to_execution_change", SszTypes.SignedBLSToExecutionChange}
-      # use type SignedAggregateAndProof
-      # {"beacon_aggregate_and_proof", SszTypes.SignedBeaconBlockHeader},
+      {"bls_to_execution_change", SszTypes.SignedBLSToExecutionChange},
+      {"beacon_aggregate_and_proof", SszTypes.SignedAggregateAndProof}
```
mpaulucci, line 23: this is nice, but should we subscribe to these topics if we're not really processing them?

mpaulucci (approved):

lib/lambda_ethereum_consensus/p2p/gossipsub.ex
MegaRedHand: I left them to try out the SSZ library, and the most common gossip we are receiving is from the `beacon_aggregate_and_proof` topic. Should I comment/remove them?

mpaulucci: I guess we can leave it for now so that we can test that we're actually receiving things.

PR #197, opened 2023-09-12
feat: add Nix development environment

### LINKED ISSUE #149: Add a nix file that can bring up the node
The nix file should be, if possible, pure. It should contain:
- A valid elixir+erlang+phoenix installation
- Rust installation
- Go installation

This would allow anyone with nix to just run `nix shell` and not need to install anything to run the node.

h3lio5:
**Motivation**

Adds support to Nix development environment. 

**Description**

Steps to setup the environment.
1. Install Nix if you don't already have it. You can install it from [here](https://zero-to-nix.com/start/install)
2. Check if Nix has been successfully installed: `nix --version`.
3. To launch the environment: `nix develop`.

Closes #149


h3lio5:
Sorry, I forgot to add phoenix installation to the script. I'll fix that later tonight.

Arkenan (changes requested):
Hi! Thanks a lot for the PR! Some feedback:

1. We should add the official nix installation site in README.md: https://nixos.org/download
2. We should add any extra steps needed to run this in the README as well. For instance I had to:

```bash

mkdir ~/.config/nix
echo "experimental-features = nix-command flakes " > ~/.config/nix/nix.conf

``` 

to be able to run the `nix` command and for `flakes` support.

File: native/libp2p_nif/main.h [hunk 1]
```diff
@@ -0,0 +1,117 @@
+/* Code generated by cmd/cgo; DO NOT EDIT. */
```
MegaRedHand, line 1: This file can be deleted

flake.nix
```diff
@@ -0,0 +1,46 @@
+{
+  description = "Nix development environment for LambdaClass Ethereum Consensus Client.";
+
+  # Flake inputs
+  inputs = {
+    nixpkgs.url      = "github:NixOS/nixpkgs/nixos-unstable";
```
MegaRedHand, line 6: Why is this unstable?

h3lio5: I understand that it is like the nightly version and also it was used in most tutorials. I can change it to the stable version if needed.

native/libp2p_nif/main.h
h3lio5: sure

flake.nix
MegaRedHand: I defer the decision to @Arkenan, but I'd prefer stable if none of the packages require otherwise.

Arkenan (approved):
LGTM, we can merge it after changing the nix version to a stable one. If unstable is required, please add a comment on why and we can merge this as is.

h3lio5:
> LGTM, we can merge it after changing the nix version to a stable one. If unstable is required, please add a comment on why and we can merge this as is.

I think we may we fine with the stable version too as we are anyways keeping the tooling (rust, golang, elixir) version fixed. I found this comment relevant though: https://discourse.nixos.org/t/what-do-i-risk-by-using-nixos-unstable/16195/4

Arkenan:
Hey, that's a good source. Feel free to solve the conflicts and merge to main

PR #201, opened 2023-09-13
chore: move bootnodes to the configuration

MegaRedHand:


mpaulucci (approved):
nice

config/config.exs
```diff
@@ -15,5 +16,27 @@ config :lambda_ethereum_consensus, BeaconApi.Endpoint,
     layout: false
   ]
 
+# Configures peer discovery
```
mpaulucci, line 19: These are the ones corresponding to Sepolia, right? If that's the case, maybe a comment would help?

MegaRedHand: Those are for mainnet, and it's commented on the bootnodes option.

PR #203, opened 2023-09-13
feat: add BlockStore and store raw blocks coming from gossip.

mpaulucci:
**Motivation**

**Description**

Closes #issue_number


lib/lambda_ethereum_consensus/event_handlers/block_handler.ex
```diff
@@ -0,0 +1,15 @@
+defmodule LambdaEthereumConsensus.Handlers.BlockHandler do
```
mpaulucci, line 1: Alternatively, we could also have one single event handler with one function per topic like:
`def handle_message("beacon_block", %SignedBeaconBlock{message: block}) do`

lib/lambda_ethereum_consensus/event_handlers/block_handler.ex
```diff
@@ -0,0 +1,15 @@
...
+  Module that handles blocks that are received through the gossipsub topic.
+  """
+  alias LambdaEthereumConsensus.Store.BlockStore
+  alias SszTypes.SignedBeaconBlock
+
+  require Logger
+
+  @spec handle_message(String.t(), SszTypes.SignedBeaconBlock.t()) :: :ok
+  def handle_message(_topic_name, %SignedBeaconBlock{message: block}) do
+    Logger.notice("Block decoded: '#{inspect(block.slot)}'")
```
MegaRedHand, line 12: We should lower this to `debug`

MegaRedHand (approved):

Issue #204, opened 2023-09-13
Move LibP2P implementation to ports


Arkenan:
NIFs are not great for blocking operations like reading through a socket. In our case, subscriptions and discovery through libp2p are a problem. We want to make a go engine that interacts with libp2p and use ports to communicate to it through elixir.

Steps:
[] A PoC should contain at least the port owner on the elixir side (GenServer), goroutines issuing notifications from the go side and a serialization mechanism for communication between them. 
[] Move discovery from NIFs to ports.
[] Move gossipsub from NIFs to ports.

PR #205, opened 2023-09-13
feat: ports for libp2p (PoC)

Arkenan:
**Description**

General changes:
- Added a protobuf spec representing commands and notifications for a port interface. Commands are subscribe and unsubscribe from topics. Notifications are gossipsub notifications or responses to commands.
- Added a proto make target for code generation both in go and elixir.
- Communication is size-delimited both in elixir and go.

Go changes:
- Added a main file that listens to commands, handles them and issues responses as notifications.
- Added a library (Port) that can read and write protobuf from a port (stdin and stdout).
- Added a library (Subscription) that launches goroutines mocking subscriptions, sending async messages once every second.

Elixir changes:
- Added a Genserver that can issue subscribe and unsubscribe messages to the port.
- The ports notifications are handled by the genserver as well. Unused bytes are stored in the genserver state to be concatenated with messages that might be arriving next.

Tackles the first step of #204


libp2p_port/internal/port/port.go
```diff
@@ -0,0 +1,49 @@
+package port
```
mpaulucci, line 1: maybe move to `native/libp2p_port`?

Arkenan: As this is not a NIF (Native implemented function) I added the "port" reference to the directory instead. This code is not dynamically linked with our code, so I thought calling it "native" was a bit confusing. I can move it to the native folder if you still think that's more appropriate though!

Makefile
```diff
@@ -68,17 +68,22 @@ compile-native: $(OUTPUT_DIR)/libp2p_nif.so
...
 	iex -S mix
 
 # Install mix dependencies.
 deps:
 	cd $(LIBP2P_DIR)/go_src; \
 	go get && go install
+	cd libp2p_port; \
+	go get && go install
+	mix escript.install hex protobuf
+	go install google.golang.org/protobuf/cmd/protoc-gen-go@latest
```
MegaRedHand, line 85: `go get && go install` was failing for me back when I didn't have the protobuf-generated files. The order of operations should be:
1. `go install google.golang.org/protobuf/cmd/protoc-gen-go@latest`
2. generate protobuf files
3. `go get && go install`

lib/libp2p_port.ex
```diff
@@ -0,0 +1,140 @@
...
+  def handle_cast({:send, data}, %{port: port} = state) do
+    send_delimited(port, data)
+    {:noreply, state}
+  end
+
+  @impl GenServer
+  def handle_info({_port, {:data, data}}, %{buffer: buffer, waiting_for: size} = state) do
+    new_state = handle_buffer(buffer <> data, size)
+    {:noreply, state |> Map.merge(new_state)}
+  end
```
MegaRedHand, line 76: This could be simplified using the `{:packet, 4}` option in `Port.open`. That way the port driver handles the size header parsing and segmenting of messages (Elixir-side only).

MegaRedHand (approved):
Left some comments, but are non-blocking.

Makefile
Arkenan: Done, thanks!

lib/libp2p_port.ex
Arkenan: It's insane that elixir handles all of this automatically. Simplified the code as you said!

PR #207, opened 2023-09-14
feat: add `operations` test runner

phklive:
**Motivation**

Added the complete`operations` test runner

**Description**

Added the `operations` test runner that enables us to test the functionality of our core internal functions as `process_execution_payload` and others.


native/libp2p_nif/main.h
```diff
@@ -0,0 +1,117 @@
+/* Code generated by cmd/cgo; DO NOT EDIT. */
+
+/* package command-line-arguments */
+
+
```
MegaRedHand, line 5: Please remove this file.

test/spec/runners/operations.ex
```diff
@@ -0,0 +1,221 @@
...
+  end
+
+  @doc """
+  Each handler has associated types, hence we resolve each type that has to be used from each handler
+  """
+  def resolve_type_from_handler(handler) do
+    case Map.get(@type_map, handler) do
+      nil -> raise "Unknown case #{handler}"
+      type -> Module.concat(SszTypes, type)
+    end
```
MegaRedHand, line 209: [`Map.fetch!`](https://hexdocs.pm/elixir/1.12.3/Map.html#fetch!/2) can also be used for this

MegaRedHand (approved):

native/libp2p_nif/main.h
phklive: Removed

test/spec/runners/operations.ex
phklive: Would it be better using Map.fetch! ?

MegaRedHand: I think it'd be worse, because the message is less clear (i.e. "key ... not found in: %{...}").

PR #211, opened 2023-09-14
feat: add `BeaconState` persistence

### LINKED ISSUE #209: [Sync] Save current `BeaconState` to memory
Since we can get the state from a finalized checkpoint through checkpoint sync, we need to persist it in memory. For now, we can use an Elixir `Agent` to store it.

MegaRedHand:
Closes #209


lib/lambda_ethereum_consensus/store/state_store.ex
```diff
@@ -0,0 +1,49 @@
...
+  alias LambdaEthereumConsensus.Store.Db
+  alias LambdaEthereumConsensus.Store.Utils
+  alias SszTypes.BeaconState
+
+  @state_prefix "beacon_state"
+  @slot_prefix "slot"
+
+  @spec store_state(BeaconState.t()) :: :ok
+  def store_state(%BeaconState{} = state) do
+    {:ok, state_root} = Ssz.hash_tree_root(state)
```
Arkenan, line 14: Not sure how we're using this yet, but we should consider using the block roots as storage keys for beacon states, as those are very tightly coupled. For example, in `on_block`, in the specs, we have:

```python
block_root = hash_tree_root(block)

# ...

# Add new block to the store
store.blocks[block_root] = block
# Add new state for this block to the store
store.block_states[block_root] = state
```

Arkenan (approved):
Great PR! Added a comment on the block roots as candidate for state store. Let me know what you think. Otherwise, feel free to merge this.

lib/lambda_ethereum_consensus/store/state_store.ex
MegaRedHand: I didn't know this 😮. I'll make this change, although, in the future, we might need another "table" mapping the state root to the state for [serving API requests](https://ethereum.github.io/beacon-APIs/#/Beacon/getStateRoot).

PR #218, opened 2023-09-16
feat: add Beacon API endpoints to query block's root

MegaRedHand:


lib/lambda_ethereum_consensus/store/block_store.ex
```diff
@@ -3,27 +3,49 @@ defmodule LambdaEthereumConsensus.Store.BlockStore do
...
+  @slothash_prefix "slothash"
 
-  @spec store_block(SszTypes.BeaconBlock.t()) :: any
+  @spec store_block(SszTypes.BeaconBlock.t()) :: :ok
   def store_block(%SszTypes.BeaconBlock{} = block) do
     {:ok, block_root} = Ssz.hash_tree_root(block)
     {:ok, encoded_block} = Ssz.to_ssz(block)
 
-    key = @prefix <> block_root
+    key = get_key(@block_prefix, block_root)
```
Arkenan, line 14: Nit and optional: now that we have two different prefixes, I'd add a `block_key(root)` and `root_by_slot_key(slot)` functions to wrap around the `get_keys` function.

lib/lambda_ethereum_consensus/store/block_store.ex
```diff
@@ -3,27 +3,49 @@ defmodule LambdaEthereumConsensus.Store.BlockStore do
...
+  @spec get_block_root_by_slot(SszTypes.slot()) ::
+          {:ok, SszTypes.root()} | {:error, String.t()} | :not_found
+  def get_block_root_by_slot(slot) do
+    key = get_key(@slothash_prefix, slot)
+    LambdaEthereumConsensus.Store.Db.get(key)
+  end
 
-      {:ok, block} ->
-        Ssz.from_ssz(block, SszTypes.BeaconBlock)
+  @spec get_block_by_slot(SszTypes.slot()) ::
```
Arkenan, line 36: I'd add a comment here explaining that this is a last-to arrive policy until we implement fork-choice. I.e. for now the last block to arrive for a certain block is the one returned, which is a temporary decision.

Arkenan (approved):
LGTM with small comments.

lib/lambda_ethereum_consensus/store/block_store.ex
MegaRedHand: That's a good idea!

lib/lambda_ethereum_consensus/store/block_store.ex
MegaRedHand: Done

PR #219, opened 2023-09-16
fix: ensure module is loaded before encoding/decoding

### LINKED ISSUE #217: ArgumentError: `1st argument: not an integer`
Every now and then, we get an error message like this:

```
11:11:37.580 [error] ** (ArgumentError) errors were found at the given arguments:

  * 1st argument: not an integer

    (stdlib 4.3) :binary.encode_unsigned(<<126, 48, 193, 62, 5, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0>>, :little)
    (lambda_ethereum_consensus 0.1.0) lib/ssz.ex:73: Ssz.encode_u256/1
    (elixir 1.15.4) lib/map.ex:916: Map.update!/3
    (lambda_ethereum_consensus 0.1.0) lib/ssz.ex:49: anonymous fn/1 in Ssz.encode/1
    (elixir 1.15.4) lib/enum.ex:1701: anonymous fn/3 in Enum.map/2
    (stdlib 4.3) maps.erl:411: :maps.fold_1/3
    (elixir 1.15.4) lib/enum.ex:2522: Enum.map/2
    (lambda_ethereum_consensus 0.1.0) lib/ssz.ex:49: Ssz.encode/1
```

MegaRedHand:
Closes #217

From the `Kernel.function_exported?` documentation:

> Returns true if module **is loaded** and contains a public function with the given arity, otherwise false.

Since the module wasn't loaded at first, we weren't encoding U256s properly.

### DIFF
--- a/lib/ssz.ex
+++ b/lib/ssz.ex
@@ -84,9 +84,7 @@ defmodule Ssz do
 
   # Ssz types can have special decoding rules defined in their optional encode/1 function,
   defp encode(%name{} = struct) do
-    Code.ensure_loaded!(name)
-
-    if function_exported?(name, :encode, 1) do
+    if exported?(name, :encode, 1) do
       name.encode(struct)
     else
       struct
@@ -105,9 +103,7 @@ defmodule Ssz do
 
   # Ssz types can have special decoding rules defined in their optional decode/1 function,
   defp decode(%name{} = struct) do
-    Code.ensure_loaded!(name)
-
-    if function_exported?(name, :decode, 1) do
+    if exported?(name, :decode, 1) do
       name.decode(struct)
     else
       struct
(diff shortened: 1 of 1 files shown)


lib/ssz.ex
```diff
@@ -84,6 +84,8 @@ defmodule Ssz do
 
   # Ssz types can have special decoding rules defined in their optional encode/1 function,
   defp encode(%name{} = struct) do
+    Code.ensure_loaded!(name)
```
Arkenan, line 87: As we always use function_exported paired with ensure_loaded, I'd create a small function to do both:

```elixir
def exported?(module, function, arity) do
  Code.ensure_loaded!(name)
  function_exported?(name, :encode, 1)
end
```

and then call that one directly.

Arkenan (approved):
Approved with small comment.

lib/ssz.ex
MegaRedHand: Good idea!

Issue #220, opened 2023-09-18
[SSZ] Support composite type de/serialization and hashing


MegaRedHand:

MegaRedHand:
This was/is being tackled in multiple other issues

PR #222, opened 2023-09-18
feat(ssz): list serialization and hashing

MegaRedHand:
Addresses #220 (doesn't fully solve it, as only lists are supported)


Arkenan (approved):
LGTM as this is a positive PR, having the refactor and the struct list implementation. We do need a second PR with lists of other types that can be serialized, specially binaries.

PR #225, opened 2023-09-20
feat: request blocks from peers

### LINKED ISSUE #208: [Sync] Request blocks from a peer
Since we can get the current head via checkpoint sync, the next step is getting subsequent blocks. We can do that using the `BeaconBlocksByRange` (v1/2) P2P request.

MegaRedHand:
Closes #208

Currently, the process requests a random block from a random peer and stores it in the database.

### DIFF
--- a/lib/lambda_ethereum_consensus/p2p/block_downloader.ex
+++ b/lib/lambda_ethereum_consensus/p2p/block_downloader.ex
@@ -14,19 +14,19 @@ defmodule LambdaEthereumConsensus.P2P.BlockDownloader do
 
   @impl true
   def handle_demand(incoming_demand, host) do
-    # Capella fork slot
     # TODO: get missing slots from DB
-    start_slot = 6_209_536
+    start_slot = 7_374_077
 
     # TODO: handle no-peers asynchronously?
     peer_id = get_some_peer()
 
-    payload = %SszTypes.BeaconBlocksByRangeRequest{
-      start_slot: start_slot,
-      # TODO: we need to refactor the Snappy library to return
-      # the remaining buffer when decompressing
-      count: 1
-    }
+    payload =
+      %SszTypes.BeaconBlocksByRangeRequest{
+        start_slot: start_slot,
+        # TODO: we need to refactor the Snappy library to return
+        # the remaining buffer when decompressing
+        count: 1
+      }
 
     # This should never fail
     {:ok, encoded_payload} = payload |> Ssz.to_ssz()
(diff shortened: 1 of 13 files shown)


lib/lambda_ethereum_consensus/p2p/peerbook.ex
```diff
@@ -0,0 +1,50 @@
...
+  def add_peer(peer_id) do
+    GenServer.cast(__MODULE__, {:new_peer, peer_id})
+  end
+
+  @doc """
+  Get some peer from the peerbook.
+  """
+  def get_some_peer do
+    GenServer.call(__MODULE__, :get_some_peer)
+  end
```
Arkenan, line 49: Nit: Let's move this public api calls to between start_link and the callback implementations.

lib/lambda_ethereum_consensus/p2p/block_downloader.ex
```diff
@@ -0,0 +1,132 @@
...
+      for _ <- 0..incoming_demand do
+        request_block(host) |> wrap_message()
+      end
+
+    {:noreply, blocks, host}
+  end
+
+  defp request_block(host) do
+    # TODO: get missing slots from DB
+    start_slot = 7_270_097 + :rand.uniform(100_000)
```
Arkenan, line 27: What is this magic number? can we document it and/or create a constant for it ?

lib/lambda_ethereum_consensus/p2p/block_downloader.ex
```diff
@@ -0,0 +1,132 @@
...
+    {:ok, compressed_payload} = encoded_payload |> Snappy.compress()
+
+    with {:ok, stream} <- Libp2p.host_new_stream(host, peer_id, @protocol_id),
+         :ok <- Libp2p.stream_write(stream, size_header <> compressed_payload),
+         :ok <- Libp2p.stream_close_write(stream),
+         {:ok, chunk} <- read_response(stream),
+         {:ok, block} <- decode_response(chunk) do
+      block
+    else
+      # we just ignore the error and continue
```
Arkenan, line 58: Let's also log the error. If it's too verbose, we can add it as `debug` level for now.

lib/lambda_ethereum_consensus/p2p/block_downloader.ex
```diff
@@ -0,0 +1,132 @@
...
+  defp read_response(stream) do
+    result =
+      stream
+      |> Libp2p.Stream.from()
+      |> Enum.reduce({:ok, ""}, fn
+        {:ok, chunk}, {:ok, acc} -> {:ok, acc <> chunk}
+        {:error, reason}, _ -> {:error, reason}
+      end)
+
+    fork_context = "BBA4DA96" |> Base.decode16!()
```
Arkenan, line 72: What is this magic string? Let's document it and/or create a constant.

lib/lambda_ethereum_consensus/p2p/block_downloader.ex
```diff
@@ -0,0 +1,132 @@
...
+
+  defp get_some_peer do
+    case P2P.Peerbook.get_some_peer() do
+      nil ->
+        Process.sleep(1000)
+        get_some_peer()
+
+      peer_id ->
+        peer_id
+    end
```
Arkenan, line 130: This is an ok patch, but let's start an issue to think about what to do when discovery hasn't yet reached a reasonable amount of peers.

Arkenan (approved):
Approved with comments

lib/lambda_ethereum_consensus/p2p/block_downloader.ex
MegaRedHand: It's a random number close to the HEAD minus 100K, to mock the slot choosing aspect of the function. This was replaced by a proper algorithm in #228

lib/lambda_ethereum_consensus/p2p/block_downloader.ex
MegaRedHand: 👍  Created #233

PR #228, opened 2023-09-23
feat: request missing blocks only

MegaRedHand:
Depends on #211 #225

This PR adds missing block detection to our DB, and uses that to request blocks from peers. It also refactors the module to only serve the DB handle and allowing concurrent access to it, since _levelDB_ is already thread-safe.

### DIFF
--- a/lib/lambda_ethereum_consensus/p2p/block_downloader.ex
+++ b/lib/lambda_ethereum_consensus/p2p/block_downloader.ex
@@ -23,6 +23,9 @@ defmodule LambdaEthereumConsensus.P2P.BlockDownloader do
     blocks = request_blocks(host, demand)
     remaining = demand - length(blocks)
 
+    # Since request_blocks always returns the demanded amount,
+    # this can only happen if database is empty, or we don't
+    # have any blocks to request
     if remaining > 0 do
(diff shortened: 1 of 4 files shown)


lib/lambda_ethereum_consensus/p2p/block_downloader.ex
```diff
@@ -13,29 +14,43 @@ defmodule LambdaEthereumConsensus.P2P.BlockDownloader do
...
-      for _ <- 0..incoming_demand do
-        request_block(host) |> wrap_message()
-      end
+  def handle_demand(incoming_demand, {host, remaining_demand}) do
+    demand = incoming_demand + remaining_demand
+    blocks = request_blocks(host, demand)
+    remaining = demand - length(blocks)
 
-    {:noreply, blocks, host}
+    if remaining > 0 do
```
mpaulucci, line 26: please comment why we might need to retry

mpaulucci (approved):

lib/lambda_ethereum_consensus/p2p/block_downloader.ex
MegaRedHand: Done

PR #244, opened 2023-09-27
feat: integrate checkpoint sync with block requesting

MegaRedHand:
Depends on #228

### DIFF
--- a/lib/lambda_ethereum_consensus/application.ex
+++ b/lib/lambda_ethereum_consensus/application.ex
@@ -2,6 +2,7 @@ defmodule LambdaEthereumConsensus.Application do
   # See https://hexdocs.pm/elixir/Application.html
   # for more information on OTP Applications
   @moduledoc false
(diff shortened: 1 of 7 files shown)


lib/lambda_ethereum_consensus/application.ex
```diff
@@ -39,7 +31,23 @@ defmodule LambdaEthereumConsensus.Application do
     # See https://hexdocs.pm/elixir/Supervisor.html
     # for other strategies and supported options
     opts = [strategy: :one_for_one, name: LambdaEthereumConsensus.Supervisor]
-    Supervisor.start_link(children, opts)
+
+    with res = {:ok, _} <- Supervisor.start_link(children, opts) do
+      # Check for the --checkpoint-sync argument and act accordingly
+      # TODO: this could be done in an async task
```
mpaulucci, line 37: can you remove the async comment now?

PR #246, opened 2023-09-28
feat: make nix work

Arkenan:
**Description**

Changes:
- Added protobuf nix package.
- Added erlang 25 nix package.
- Added go packages and mix escripts folders to the path so that they can be accessed after compiling.
- Moved the installation and compilation of protos to scripts so that they can be reused and can include conditionals.
- The elixir script is only installed if not present.
- If using asdf, there's a reshim after installing the protos.


flake.nix
```diff
@@ -19,8 +19,7 @@
           };
 
           rustToolchain = pkgs.rust-bin.stable."1.71.1".default;
-
-          elixir = pkgs.beam.packages.erlang_25.elixir_1_15;
+          otp = pkgs.beam.packages.erlang_25;
```
MegaRedHand, line 22: The Erlang version should be 26

MegaRedHand (approved):

PR #248, opened 2023-09-28
test: add epoch processing spec test runner.

mpaulucci:
**Motivation**
Adding the test runner will unblock implementing transition functions

### DIFF
--- a/test/spec/runners/epoch_processing.ex
+++ b/test/spec/runners/epoch_processing.ex
@@ -1,0 +1,61 @@
+defmodule EpochProcessingRunner do
+  use ExUnit.CaseTemplate
+  use TestRunner
+
+  @moduledoc """
+  Runner for Epoch Processing test cases. See: https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/epoch_processing
+  """
+
+  # Remove handler from here once you implement the corresponding functions
+  @disabled_handlers [
+    "justification_and_finalization",
+    "inactivity_updates",
+    "rewards_and_penalties",
+    "registry_updates",
+    "slashings",
+    "eth1_data_reset",
+    "effective_balance_updates",
+    "slashings_reset",
+    "randao_mixes_reset",
+    "historical_summaries_update",
+    "participation_record_updates",
+    "participation_flag_updates",
+    "sync_committee_updates"
+  ]
+
+  @deprecated_handlers [
+    "historical_roots_update"
+  ]
+
+  @impl TestRunner
+  def skip?(%SpecTestCase{} = testcase) do
+    Enum.member?(@disabled_handlers ++ @deprecated_handlers, testcase.handler)
+  end
+
+  @impl TestRunner
+  def run_test_case(%SpecTestCase{} = testcase) do
+    case_dir = SpecTestCase.dir(testcase)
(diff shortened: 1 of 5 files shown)


test/spec/runners/epoch_processing.ex
```diff
@@ -0,0 +1,61 @@
+defmodule EpochProcessingRunner do
```
MegaRedHand, line 1: To be consistent, this should end with `TestRunner`:

```suggestion
defmodule EpochProcessingTestRunner do
```

test/spec/utils/generator.ex
```diff
@@ -5,7 +5,8 @@ defmodule SpecTestGenerator do
   @runner_map %{
     "ssz_static" => SSZStaticTestRunner,
     "bls" => BLSTestRunner,
-    "operations" => OperationsTestRunner
+    "operations" => OperationsTestRunner,
+    "epoch_processing" => EpochProcessingRunner
```
MegaRedHand, line 9: Here too:

```suggestion
    "epoch_processing" => EpochProcessingTestRunner
```

test/spec/utils/utils.ex
```diff
@@ -27,4 +27,15 @@ defmodule SpecTestUtils do
...
+
+  def get_config("minimal"), do: MinimalConfig
+  def get_config("mainnet"), do: MainnetConfig
+
+  def read_ssz_from_file(file_path, ssz_type, config) do
+    with {:ok, compressed} <- File.read(file_path),
+         {:ok, decompressed} <- :snappyer.decompress(compressed),
+         {:ok, ssz_object} <- Ssz.from_ssz(decompressed, ssz_type, config) do
+      {:ok, ssz_object}
+    end
```
MegaRedHand, line 39: Since this is a test util, I think we should use plain matches and just let it crash™
That way errors mention the specific line that caused the error.

test/spec/runners/operations/runner.ex
```diff
@@ -30,17 +30,29 @@ defmodule OperationsTestRunner do
   def run_test_case(%SpecTestCase{} = testcase) do
     case_dir = SpecTestCase.dir(testcase)
     handler = testcase.handler
+    config = SpecTestUtils.get_config(testcase.config)
 
-    {:ok, pre} = OperationsTestUtils.prepare_test(case_dir, handler, "pre")
+    {:ok, pre} =
+      SpecTestUtils.read_ssz_from_file(
```
MegaRedHand, line 36: This one should be `read_ssz_from_file!` since the _pre_ file should always exist, right?

test/spec/runners/epoch_processing.ex
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  @impl TestRunner
+  def run_test_case(%SpecTestCase{} = testcase) do
+    case_dir = SpecTestCase.dir(testcase)
+
+    config = SpecTestUtils.get_config(testcase.config)
+
+    {:ok, _pre} =
+      SpecTestUtils.read_ssz_from_file(
```
MegaRedHand, line 42: Same here

MegaRedHand (approved):
🚀 🚀 🚀

PR #268, opened 2023-10-03
feat: added runtime presets

phklive:
**Motivation**

Addition of runtime presets


config/runtime.exs
```diff
@@ -0,0 +1,8 @@
+# General application configuration
+import Config
+
+config :lambda_ethereum_consensus, :mainnet_capella_presets,
+  YamlElixir.read_from_file!("config/presets/mainnet/capella.yaml") |> SpecTestUtils.parse_yaml()
```
phklive, line 5: @MegaRedHand Here we do not need the parse yaml its only for the tests right ?

phklive: Then i would let you or someone takeover considering I will go off for this week.

mpaulucci (approved):

PR #276, opened 2023-10-05
Remove proto generated files from credo checks

Arkenan:
**Motivation**

Proto generated files might contain linter errors like not having newlines at the end of the file, but we can't and won't change them manually, plus they are not in source control either way.

**Description**

This PR removes the generated elixir protos from the linter checks.

### DIFF
--- a/.credo.exs
+++ b/.credo.exs
@@ -25,7 +25,7 @@
           "lib/",
           "test/"
         ],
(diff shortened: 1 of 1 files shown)


mpaulucci:
I think it was fixed with this? https://github.com/lambdaclass/lambda_ethereum_consensus/pull/273/files#diff-aad92a030ac915c80a2e0d0204d240fc952893bc87b372dc1b6d485609f27260R3-R6

PR #277, opened 2023-10-05
docs: update consensus doc and add architecture

Arkenan:
Changes:

- Updated basic consensus and EVM basics doc to be simpler and clearer.
- Added clients doc that explains how execution and consensus interact.
- Added architecture doc that explains a bit of how that is implemented in this project. This is a very in-progress document and will be expanded in future PRs.


docs/architecture.md
```diff
@@ -0,0 +1,126 @@
...
+```
+
+## Networking
+
+The main entry for new events is the gossip protocol, which is the way in which our consensus node communicates with other consensus nodes. This includes:
+
+1. Discovery: our node has a series of known `bootnodes` hardcoded. We request them a list of the nodes the know about and we add them to our list. We save them locally and now can use those too to request new nodes.
+2. Message propagation. When a proposer sends a new block, or validators attest for a new block, they sent those to other known nodes. Those, in turn propagate the messages sent to other nodes. This process is repeated until, ideally, the whole network receives the messages.
+
+We use a library called `libp2p`, implemented in `go`, for the networking primitives.
```
MegaRedHand, line 39: This is a bit misleading. To clarify, and quoting [the libp2p repo's readme](https://github.com/libp2p/libp2p):

> libp2p is a modular networking framework bundled together as a full stack of protocols for peer-to-peer systems.

The actual implementation we are using is `go-libp2p` (+ other `go-libp2p-*` modules), and the discovery part of `go-ethereum`'s P2P stack.

I'd rephrase this as:

```
We use the `go-libp2p` library for the networking primitives, which is an implementation of the `libp2p`networking stack.
```

docs/clients.md
```diff
@@ -0,0 +1,83 @@
+# Clients
+
+As we talked previously, Ethereum is a network made by different computers. Its workflow would work, in theory, as follows:
+
+- A user of the network sends a transaction (e.g. an eth transfer) to the network.
```
MegaRedHand, line 5: nit
```suggestion
- A user of the network sends a transaction (e.g. an ETH transfer) to the network.
```

MegaRedHand (commented):
Great explanation! I left some comments, mainly about typos and things related to the networking spec

docs/clients.md
```diff
@@ -0,0 +1,83 @@
...
+    2. Save the attestations and update the weights of the blocks on the tree.
+    3. Apply fork choice to recalculate, if necessary, what blocks are included in the canonical chain.
+    4. The changes + the new block are propagated to the execution node so transactions are applied and state is changed and up to date.
+
+## Caveats
+
+This is the ideal happy path. There are other ways in which this can work, and ways in which something can go wrong:
+
+- Not all users run their own node. They use services like infura or alchemy, that take JSON-RPC requests and process them with their own nodes. Wallets like metamask, for instance, use these.
+- There are intermediaries like flashbots, and MEV, which are outside of the scope of this documentations, that cause variations in the "who talks to who" diagram.
```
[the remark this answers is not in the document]
Arkenan: True! I meant that for MEV, execution clients send their transactions for external intermediaries to build blocks, but I'll leave that for another time.

MegaRedHand (approved):

PR #283, opened 2023-10-08
feat: add process_eth1_data_reset

### LINKED ISSUE #259: [epoch processing] Implement `process_eth1_data_reset`
https://github.com/lambdaclass/lambda_ethereum_consensus/blob/66059de3653c17ae620ce2aca2db61c558057f6a/docs/specs/beacon-chain.md?plain=1#L2485-L2491

Acceptance criteria:
- Run corresponding spec tests by removing https://github.com/lambdaclass/lambda_ethereum_consensus/blob/66059de3653c17ae620ce2aca2db61c558057f6a/test/spec/runners/epoch_processing.ex#L16

Godspower-Eze:
**Description**

Added process_eth1_data_reset

Closes #259


lib/lambda_ethereum_consensus/state_transition/epoch_processing.ex
```diff
@@ -0,0 +1,33 @@
...
+  alias ChainSpec
+  alias SszTypes.BeaconState
+
+  @spec process_eth1_data_reset(BeaconState.t()) :: BeaconState.t()
+  def process_eth1_data_reset(state) do
+    next_epoch = get_current_epoch(state) + 1
+    epochs_per_eth1_voting_period = ChainSpec.get("EPOCHS_PER_ETH1_VOTING_PERIOD")
+
+    if rem(next_epoch, epochs_per_eth1_voting_period) == 0 do
+      %BeaconState{state | eth1_data_votes: []}
```
mpaulucci, line 15: you are not doing anything on this line. In Elixir, values are immutable. So you should assign the result to state

lib/lambda_ethereum_consensus/state_transition/epoch_processing.ex
```diff
@@ -0,0 +1,33 @@
+defmodule LambdaEthereumConsensus.StateTransition.EpochProcessing do
+  @moduledoc """
+  This module contains utility functions for handling epoch processing
+  """
+
+  alias ChainSpec
+  alias SszTypes.BeaconState
+
+  @spec process_eth1_data_reset(BeaconState.t()) :: BeaconState.t()
```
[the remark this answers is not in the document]
Godspower-Eze: Yeah, it makes sense.

I mainly did it this way because this function shouldn't fail, so I didn't see the need to add an exception handler and return an error.

lib/lambda_ethereum_consensus/state_transition/epoch_processing.ex
Godspower-Eze: Got it! Thanks

lib/lambda_ethereum_consensus/state_transition/epoch_processing.ex
```diff
@@ -0,0 +1,24 @@
...
+  def process_eth1_data_reset(state) do
+    next_epoch = Accessors.get_current_epoch(state) + 1
+    epochs_per_eth1_voting_period = ChainSpec.get("EPOCHS_PER_ETH1_VOTING_PERIOD")
+
+    if rem(next_epoch, epochs_per_eth1_voting_period) == 0 do
+      ^state = %BeaconState{state | eth1_data_votes: []}
+    end
+
+    {:ok, state}
+  rescue
```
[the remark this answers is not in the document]
Godspower-Eze: Okay. Done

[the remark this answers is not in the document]
mpaulucci: https://hexdocs.pm/elixir/main/try-catch-and-rescue.html this is a good read and explains why we generally want to avoid `rescue`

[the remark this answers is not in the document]
Godspower-Eze: Should I remove it completely?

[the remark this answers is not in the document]
mpaulucci: Try doing a `case` on the post first 

case post do
   nil -> assert {:error, reason } == result
   post -> assert {:ok, post } = result

something like this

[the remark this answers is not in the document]
Godspower-Eze: Interesting. Thanks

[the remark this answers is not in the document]
Godspower-Eze: Hmmm...this doesn't work either. You know `reason` is undefined and when I replace it with `_`, it doesn't work either.

test/spec/runners/epoch_processing.ex
```diff
@@ -55,7 +57,12 @@ defmodule EpochProcessingTestRunner do
...
   end
 
-  # def handle_case("eth1_data_reset", pre, post) do
-  #   assert process_eth1_data_reset(process_eth1pre.eth1_data) == post.eth1_data
-  # end
+  def handle_case("eth1_data_reset", pre, post) do
+    result = EpochProcessing.process_eth1_data_reset(pre)
+
+    case post do
+      nil -> elem(result, 0) == :error
```
mpaulucci, line 64: you are not asserting anything in this line

Godspower-Eze: Corrected that. Thanks

Godspower-Eze: I think the issue with the build from dialyzer is that we are checking for the error but it is not returned explicitly in the function.

What if I rescue it and reraise it?

mpaulucci: Ah, gotcha. Is there a spec test where post is actually `nil`? Otherwise you just need to handle the `:ok` scenario

mpaulucci: It seems this function always succeeds so we can update the @spec removing the `:error` possibility. That was my bad, sorry about that

Godspower-Eze: Yeah, it never fails

lib/lambda_ethereum_consensus/state_transition/epoch_processing.ex
```diff
@@ -0,0 +1,21 @@
+defmodule LambdaEthereumConsensus.StateTransition.EpochProcessing do
+  @moduledoc """
+  This module contains utility functions for handling epoch processing
+  """
+
+  alias ChainSpec
+  alias LambdaEthereumConsensus.StateTransition.Accessors
+  alias SszTypes.BeaconState
+
+  @spec process_eth1_data_reset(BeaconState.t()) :: BeaconState.t()
```
mpaulucci, line 10: Sorry, this should still return `{:ok, BeaconState.t()}` just to be consistent across the different processing functions

Godspower-Eze: Okay. Done

mpaulucci (approved):

PR #284, opened 2023-10-09
chore: clean up spec-tests code and move vectors

### LINKED ISSUE #146: Downloaded test vectors should go inside `spec/test` folder
We're currently downloading the spec test vectors to the root of the project. I think it would be better to download them and uncompress them inside a special folder, i.e `test/spec/vectors`. The main reason is not to pollute the root of the project and avoid confusion (for example, we have `test` and `tests`).

Acceptance criteria:
- Download test vectors to a special folder inside `test/spec`. See `make download-vectors`
- Update spec test generator to look for tests inside that folder: https://github.com/lambdaclass/lambda_ethereum_consensus/blob/main/test/spec/utils/generator.ex
- Optional: do we want to remove `*.tar.gz` files after decompressing them?
- Update `.gitignore` if applicable

MegaRedHand:
Closes #146

This PR moves the vector download directory to `test/spec/vectors`, and moves the config setting on the spec-tests to the test generator, cleaning up some of the code.

### DIFF
--- a/test/spec/utils/generator.ex
+++ b/test/spec/utils/generator.ex
@@ -9,12 +9,14 @@ defmodule SpecTestGenerator do
     "epoch_processing" => EpochProcessingTestRunner
   }
 
+  @vectors_dir SpecTestUtils.get_vectors_dir()
+
   def all_cases do
-    ["tests"]
+    [@vectors_dir]
     |> Stream.concat(["*"] |> Stream.cycle() |> Stream.take(6))
     |> Enum.join("/")
     |> Path.wildcard()
-    |> Stream.map(&Path.relative_to(&1, "tests"))
+    |> Stream.map(&Path.relative_to(&1, @vectors_dir))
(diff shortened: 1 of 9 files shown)


test/spec/utils/generator.ex
```diff
@@ -47,10 +50,16 @@ defmodule SpecTestGenerator do
...
       # Recompile module only if corresponding dir layout changed
       def __mix_recompile__? do
-        Path.wildcard(unquote("tests/#{pinned_config}/#{pinned_fork}/**"))
+        Path.wildcard(unquote("#{vectors_dir}/#{pinned_config}/#{pinned_fork}/**"))
         |> :erlang.md5() != unquote(paths_hash)
       end
 
+      config = SpecTestUtils.get_config(pinned_config)
+
+      setup_all do
```
mpaulucci, line 59: nice!

mpaulucci (approved):

PR #287, opened 2023-10-10
feat: add GenServer that deals with pending blocks

### LINKED ISSUE #253: Downloaded blocks should be moved to pending before being processed
When we get a block from a peer, we should do this:

1. Add to a "pending blocks" collection (which should be in memory).
2. If the parent block is in fork choice (processed), we can start processing the block. If the parent is not in fork choice, we should download the parent. Once the parent is downloaded, repeat step 1 with the parent block.
3. If the processing is successful, we add the block to fork choice and remove it from the pending blocks collection. We then process all children of the processed blocks that are in the pending blocks collection. On the other hand, if the processing is invalid, we remove the block and all the descendants from the pending collection

mpaulucci:
**Motivation**
When we get blocks from gossip, there might be several reasons why we can't add the block to fork choice. The main reason is that the parent block might be unknown, but there might be others.

**Description**
We create a new GenServer that stores pending blocks, asks for parent block if necesary, and eventually adds blocks to the fork choice.

Closes #253

### DIFF
--- a/test/fixtures/utils.ex
+++ b/test/fixtures/utils.ex
@@ -34,7 +34,7 @@ defmodule Fixtures.Random do
   end
 
   @spec uint64 :: pos_integer
-  def uint64() do
+  def uint64 do
     :rand.uniform(2 ** 64 - 1)
   end
 end


lib/lambda_ethereum_consensus/p2p/block_downloader.ex
```diff
@@ -3,54 +3,25 @@ defmodule LambdaEthereumConsensus.P2P.BlockDownloader do
   This module requests blocks from peers.
```
mpaulucci, line 3: This module was a GenServer, but since I changed it to download blocks only when asked, I removed that. The problem is that now we have to pass the `host` every time. Would it make sense to store the host in a centralized place, maybe PeerBook?

test/fixtures/block.ex
```diff
@@ -0,0 +1,94 @@
+defmodule Fixtures.Block do
```
mpaulucci, line 1: TODO: change integers for random integers?

test/fixtures/block.ex
```diff
@@ -0,0 +1,73 @@
...
+      body: beacon_block_body()
+    }
+  end
+
+  @spec beacon_block_body :: SszTypes.BeaconBlockBody.t()
+  def beacon_block_body do
+    %SszTypes.BeaconBlockBody{
+      randao_reveal: Random.bls_signature(),
+      eth1_data: eth1_data(),
+      graffiti: Random.hash(),
```
MegaRedHand, line 24: Since _hash_ is not the specific type, we should call it `Random.bytes32()` or `Random.bytes(32)` to avoid confusion.

MegaRedHand (approved):

lib/lambda_ethereum_consensus/p2p/gossip_handler.ex
```diff
@@ -5,16 +5,17 @@ defmodule LambdaEthereumConsensus.P2P.GossipHandler do
...
 
-  alias LambdaEthereumConsensus.Store.BlockStore
+  alias LambdaEthereumConsensus.Beacon.PendingBlocks
   alias SszTypes.SignedBeaconBlock
 
   @spec handle_message(String.t(), SszTypes.SignedBeaconBlock.t()) :: :ok
   def handle_message("/eth2/bba4da96/beacon_block/ssz_snappy", %SignedBeaconBlock{message: block}) do
     Logger.debug(
-      "[Gossip] Block decoded for slot #{block.slot}. Root: #{Base.encode16(block.state_root)}"
+      "[Checkpoint sync] Block decoded for slot #{block.slot}. Root: #{Base.encode16(block.state_root)}"
```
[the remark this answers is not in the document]
mpaulucci: Bad merge, I will revert

lib/lambda_ethereum_consensus/p2p/gossip_handler.ex
```diff
@@ -5,16 +5,17 @@ defmodule LambdaEthereumConsensus.P2P.GossipHandler do
...
     )
 
-    BlockStore.store_block(block)
+    PendingBlocks.add_block(block)
+    :ok
```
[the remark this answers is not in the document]
mpaulucci: I know, but this caused a weird behaviour when suddenly add_block was returning something different. I think this makes it more explicit. Maybe I should pattern match the return from add_block though.

test/fixtures/block.ex
mpaulucci: Could be, but it's something we could work on as a follow up.

lib/lambda_ethereum_consensus/p2p/block_downloader.ex
mpaulucci: nice!

test/fixtures/block.ex
MegaRedHand: If it's just generating random numbers, you can just use `:rand.uniform/1`. Should be a small change, if you're interested.

lib/lambda_ethereum_consensus/beacon/pending_blocks.ex
```diff
@@ -0,0 +1,122 @@
...
+              {:ok, signed_block} ->
+                Logger.debug("Block downloaded: #{signed_block.message.slot}")
+                block = signed_block.message
+                add_block(block)
+
+              {:error, reason} ->
+                Logger.error("Block download failed: '#{reason}'")
+            end
+
+            nil
```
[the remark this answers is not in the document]
mpaulucci: well, actually Im taking the return values and assigning them to `blocks_to_remove`. If I don't return something, the last evaluation will be returned, which probably will yield the same result, but I prefer to be explicit.

PR #291, opened 2023-10-10
feat: implement peer discovery in ports

MegaRedHand:
Depends on #269


native/libp2p_port/internal/discovery/discovery.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/libp2p/go-libp2p/core/peer"
+)
+
+type Discoverer struct {
+	port           *port.Port
+	discv5_service *discover.UDPv5
+}
+
+func NewDiscoverer(p *port.Port, listener *reqresp.Listener, config *proto_helpers.Config) Discoverer {
+	if !config.UseDiscv5 {
```
Arkenan, line 25: Does this mean we only support discv5? If that's the case we should consider:
1) Documenting that.
2) Why do we need a boolean if that's the case? Are we planning on supporting something else?

native/libp2p_port/internal/discovery/discovery.go
```diff
@@ -0,0 +1,103 @@
...
+		// NetRestrict *netutil.Netlist  // list of allowed IP networks
+		Bootnodes: bootnodes, // list of bootstrap nodes
+		// Unhandled   chan<- ReadPacket // unhandled packets are sent on this channel
+		// Log         log.Logger        // if set, log messages go here
+
+		// V5ProtocolID configures the discv5 protocol identifier.
+		// V5ProtocolID *[6]byte
+
+		// ValidSchemes enr.IdentityScheme // allowed identity schemes
+		// Clock        mclock.Clock
```
Arkenan, line 59: We should move the optional parameters comment outisde of this assignment, it makes it a bit hard to read. Is this a note? A TODO?

native/libp2p_port/internal/discovery/discovery.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+		key, err := utils.ConvertToInterfacePubkey(node.Pubkey())
+		if err != nil {
+			continue
+		}
+		nodeID, err := peer.IDFromPublicKey(key)
+		if err != nil {
+			continue
+		}
+		listener.AddPeer([]byte(nodeID), addrArr, 60e9)
```
Arkenan, line 101: nit: We should document the magic number/move it to a constant.

native/libp2p_port/internal/utils/utils.go
```diff
@@ -1,11 +1,51 @@
...
+	"github.com/libp2p/go-libp2p/core/crypto"
+)
 
 func PanicIfError(err error) {
 	if err != nil {
 		panic(err)
 	}
 }
+
+// Taken from Prysm
```
Arkenan, line 19: From Prysm documentation? Code? Let's add a link to the page or github blob if we don't want to go into details.

Arkenan (approved):
LGTM with minor comments.

native/libp2p_port/internal/discovery/discovery.go
Arkenan: Oh, if that's the case:
1) I'd call it as you suggest, which is clearer.
2) I'd remove that logic from here. It's strange to even build the struct if we don't enable discovery. I'd move this if to `main`. What do you think?

Arkenan (approved):

PR #294, opened 2023-10-12
feat: add initial fork choice store.

mpaulucci:
**Motivation**
We need to keep track of the fork choice. We are starting by creating a store similar to the one provided by the canonical implementation, see: https://eth2book.info/capella/annotated-spec/#store


lib/lambda_ethereum_consensus/fork_choice/store.ex
```diff
@@ -0,0 +1,116 @@
...
+      genesis_time: genesis_time,
+      justified_checkpoint: current_justified_checkpoint,
+      finalized_checkpoint: finalized_checkpoint,
+      unrealized_justified_checkpoint: nil,
+      unrealized_finalized_checkpoint: nil,
+      proposer_boost_root: nil,
+      equivocating_indices: [],
+      blocks: %{},
+      block_states: %{
+        Ssz.hash_tree_root(initial_state) => initial_state
```
MegaRedHand, line 93: The hash should be of the block, not the state.

lib/lambda_ethereum_consensus/utils.ex
```diff
@@ -27,11 +27,11 @@ defmodule LambdaEthereumConsensus.Utils do
...
+            {:ok, struct}
 
-          {:error, _} ->
+          {:error, error} ->
             Logger.error("There has been an error syncing from checkpoint.")
-            :error
+            {:error, error}
         end
 
       _ ->
```
MegaRedHand, line 37: We should return `{:error, error}` here too.

lib/lambda_ethereum_consensus/fork_choice/store.ex
mpaulucci: nice catch!

lib/lambda_ethereum_consensus/fork_choice/supervisor.ex
```diff
@@ -0,0 +1,34 @@
...
+    # Parse command line arguments
+    {args, _remaining_args, _errors} =
+      OptionParser.parse(System.argv(), switches: [checkpoint_sync: :string])
+
+    Logger.notice("Syncing from checkpoint...")
+
+    # Check for the --checkpoint-sync argument and act accordingly
+    # TODO: this could be done in an async task
+    {:ok, %SszTypes.BeaconState{} = initial_state} =
+      Keyword.fetch!(args, :checkpoint_sync) |> Utils.sync_from_checkpoint()
```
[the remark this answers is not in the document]
mpaulucci: Yes, I added some code to make it more explicit.

lib/lambda_ethereum_consensus/application.ex
```diff
@@ -14,18 +12,23 @@ defmodule LambdaEthereumConsensus.Application do
     {args, _remaining_args, _errors} =
       OptionParser.parse(System.argv(), switches: [checkpoint_sync: :string])
 
+    if not Keyword.has_key?(args, :checkpoint_sync) do
+      Logger.error("No checkpoint sync url provided.")
+      :init.stop(0)
```
MegaRedHand, line 17: nit. This should be non-zero
```suggestion
      :init.stop(1)
```

MegaRedHand (approved):

lib/lambda_ethereum_consensus/application.ex
mpaulucci: gonna add this to the follow up.

PR #309, opened 2023-10-12
feat: implement process inactivity updates

### LINKED ISSUE #255: [epoch processing] Implement `process_inactivity_updates`
This is the function

https://github.com/lambdaclass/lambda_ethereum_consensus/blob/66059de3653c17ae620ce2aca2db61c558057f6a/docs/specs/beacon-chain.md?plain=1#L2238-L2253

See Spec tests format: https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/epoch_processing

Acceptance criteria:
- Run corresponding spec tests by removing https://github.com/lambdaclass/lambda_ethereum_consensus/blob/66059de3653c17ae620ce2aca2db61c558057f6a/test/spec/runners/epoch_processing.ex#L12.

f3r10:
**Motivation**

Closes #255 

**Description**

Added the function `process_inactivity_updates`


config/configs/minimal.yaml
```diff
@@ -147,3 +147,7 @@ BLOB_SIDECAR_SUBNET_COUNT: 6
 # Whisk
 WHISK_EPOCHS_PER_SHUFFLING_PHASE: 4
 WHISK_PROPOSER_SELECTION_GAP: 1
+
+GENESIS_EPOCH: 0
+
+TIMELY_TARGET_FLAG_INDEX: 1
```
MegaRedHand, line 153: These shouldn't be here. The difference between them and preset/config is that there can be multiple presets/configs. So the constants are true constants. We should add a `Constants` module with 0-arity functions like `genesis_epoch`, `timely_target_flag_index`, and so on, instead of reading them from a file.

lib/lambda_ethereum_consensus/state_transition/accessors.ex
```diff
@@ -13,6 +30,63 @@ defmodule LambdaEthereumConsensus.StateTransition.Accessors do
     Misc.compute_epoch_at_slot(slot)
   end
 
+  @doc """
+  Return the previous epoch (unless the current epoch is ``GENESIS_EPOCH``).
+  """
+  @spec get_previous_epoch(BeaconState.t()) :: SszTypes.epoch()
+  def get_previous_epoch(%BeaconState{} = state) do
+    current_epoch = get_current_epoch(state)
+    genesis_epoch = ChainSpec.get("GENESIS_EPOCH")
```
MegaRedHand, line 39: Then this would be
```suggestion
    genesis_epoch = Constants.genesis_epoch()
```

lib/lambda_ethereum_consensus/state_transition/accessors.ex
```diff
@@ -13,6 +30,63 @@ defmodule LambdaEthereumConsensus.StateTransition.Accessors do
...
+    end
+  end
+
+  @doc """
+  Return the set of validator indices that are both active and unslashed for the given ``flag_index`` and ``epoch``.
+  """
+  @spec get_unslashed_participating_indices(BeaconState.t(), integer, SszTypes.epoch()) ::
+          {:ok, MapSet.t()} | {:error, binary()}
+  def get_unslashed_participating_indices(%BeaconState{} = state, flag_index, epoch) do
+    if Enum.member?([get_previous_epoch(state), get_current_epoch(state)], epoch) do
```
MegaRedHand, line 54:
```suggestion
    if epoch in [get_previous_epoch(state), get_current_epoch(state)] do
```

config/presets/minimal/capella.yaml
```diff
@@ -15,3 +15,8 @@ MAX_WITHDRAWALS_PER_PAYLOAD: 4
 # ---------------------------------------------------------------
 # [customized] 2**4 (= 16) validators
 MAX_VALIDATORS_PER_WITHDRAWALS_SWEEP: 16
+
+# Time parameters
+# ---------------------------------------------------------------
+# 2**5 (= 32) slots 6.4 minutes
+SLOTS_PER_EPOCH: 32
```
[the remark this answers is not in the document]
f3r10: yes!!! I did not know why with mainnet the tests were passing but with minimal did not :sweat_smile:

config/configs/minimal.yaml
f3r10: I agree, should this be done in this same pull request, or would it be better to open an independent issue?

MegaRedHand: Doing it here should be fine since it's a small change

f3r10: I created  `lib/constants.ex`. Is that location OK?

MegaRedHand: Yes 👍

lib/lambda_ethereum_consensus/state_transition/epoch_processing.ex
```diff
@@ -80,4 +81,126 @@ defmodule LambdaEthereumConsensus.StateTransition.EpochProcessing do
...
+      validators
+      |> Stream.with_index()
+      |> Stream.filter(fn {validator, _index} ->
+        Predicates.is_active_validator(validator, previous_epoch) ||
+          (validator.slashed && previous_epoch + 1 < validator.withdrawable_epoch)
+      end)
+      |> Stream.map(fn {_validator, index} -> index end)
+      |> Enum.to_list()
+
+    {:ok, validator_indices}
```
MegaRedHand, line 204: This function cannot return an error.

lib/lambda_ethereum_consensus/state_transition/epoch_processing.ex
```diff
@@ -80,4 +81,126 @@ defmodule LambdaEthereumConsensus.StateTransition.EpochProcessing do
...
+        Accessors.get_unslashed_participating_indices(
+          state,
+          timely_target_index,
+          Accessors.get_previous_epoch(state)
+        )
+
+      state_is_in_inactivity_leak = is_in_inactivity_leak(state)
+
+      updated_eligible_validator_indices =
+        with {:ok, eligible_validator_indices} <- get_eligible_validator_indices(state) do
```
MegaRedHand, line 105: Then this `with` shouldn't be needed

MegaRedHand (commented):
Just a small comment. Once it's fixed we can merge this

MegaRedHand (approved):

PR #313, opened 2023-10-16
feat: added compute_shuffled_index

### LINKED ISSUE #242: [Building blocks] Implement shuffling algorithm
To be able to decide how to divide the validators into committees, we need a shuffling algorithm. This is the one specified in the spec:

https://github.com/lambdaclass/lambda_ethereum_consensus/blob/5e97070e691fdbc685b48d73c1e01472e3f06dd5/docs/specs/beacon-chain.md?plain=1#L1149-L1172

This is the corresponding spec test: https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/shuffling

Additional read if needed: https://eth2book.info/capella/part2/building_blocks/shuffling/

Actual test cases can be seen in `lambda_ethereum_consensus/tests/minimal/phase0/shuffling/core`

Acceptance criteria:
- Implement `compute_shuffled_index` in Elixir according to the spec
- Implement `shuffling` spec test runner
- All shuffling tests should pass

Godspower-Eze:
**Description**

Adds shuffling and test runner

Closes #242


lib/lambda_ethereum_consensus/state_transition/misc.ex
```diff
@@ -11,4 +13,57 @@ defmodule LambdaEthereumConsensus.StateTransition.Misc do
...
     div(slot, slots_per_epoch)
   end
+
+  @doc """
+     Return the shuffled index corresponding to ``seed`` (and ``index_count``).
+  """
+  @spec compute_shuffled_index(SszTypes.uint64(), SszTypes.uint64(), SszTypes.bytes32()) ::
+          SszTypes.uint64()
+  def compute_shuffled_index(index, index_count, seed) do
+    Assertions.assert(index > index_count)
```
phklive, line 23: Here use: 

`unless index > index_count`

lib/lambda_ethereum_consensus/state_transition/misc.ex
```diff
@@ -2,6 +2,8 @@ defmodule LambdaEthereumConsensus.StateTransition.Misc do
   @moduledoc """
   Misc functions
   """
+  import Bitwise
+  alias ExUnit.Assertions
```
phklive, line 6: This should only exist in test files

phklive:
If you implement this you should also implement unit test to make sure the impl is correct

lib/lambda_ethereum_consensus/state_transition/misc.ex
Godspower-Eze: Based on the implementation, it should throw that's why I used that but I would change it. Thanks

Godspower-Eze:
> If you implement this you should also implement unit test to make sure the impl is correct

Yeah, it's in draft. I am not done yet

lib/lambda_ethereum_consensus/state_transition/misc.ex
phklive: In elixir we should never throw except if we have an internal error, something that should never happen. Like in Rust you never want to unwrap except if internal failure.

Hence what you want to do in those case of errors is return something following those lines: `{:error, "[Error message of this specific error]"}` enabling us then to handle the error correctly.

Godspower-Eze:
I added it I was removed when I pulled I guess. I will add update in a bit

test/spec/runners/shuffling.ex
```diff
@@ -0,0 +1,33 @@
+defmodule ShufflingTestRunner do
```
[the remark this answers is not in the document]
Godspower-Eze: Strangely, anytime I run the spectests it excludes all and the `assert false` doesn't fail even for the existing runner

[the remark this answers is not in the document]
MegaRedHand: That's because we are only running the Capella tests. The "shuffling" tests seem to be on Phase0.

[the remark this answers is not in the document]
Godspower-Eze: I added `assert false` to bls tests and it didn't fail still. Same issue?

Also, is there I way to work around this, just to test it?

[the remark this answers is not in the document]
Godspower-Eze: They are not downloaded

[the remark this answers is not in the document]
MegaRedHand: Are you running them with `make spec-test` ? If not, you should at least run `make download-vectors`.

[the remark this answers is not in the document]
MegaRedHand: With the changes in #322 (+ downloading the vectors) you should be able to run the shuffling tests.

Godspower-Eze:
@MegaRedHand @phklive please take a look at this. There's something wrong with the logic that I can't seem to figure out.

I think it has to do with the endianness.

lib/lambda_ethereum_consensus/state_transition/misc.ex
```diff
@@ -12,6 +13,69 @@ defmodule LambdaEthereumConsensus.StateTransition.Misc do
...
+        {:error, "index not less than index count"}
+      else
+        shuffle_round_count = ChainSpec.get("SHUFFLE_ROUND_COUNT")
+
+        new_index =
+          Enum.reduce(0..(shuffle_round_count - 1), index, fn round, current_index ->
+            round_as_bytes =
+              <<round>> |> :binary.decode_unsigned() |> :binary.encode_unsigned(:little)
+
+            seed_as_bytes = seed |> :binary.decode_unsigned() |> :binary.encode_unsigned(:little)
```
MegaRedHand, line 33: The `seed` here is already a binary, so no need to reinterpret it.

lib/lambda_ethereum_consensus/state_transition/misc.ex
```diff
@@ -12,6 +13,69 @@ defmodule LambdaEthereumConsensus.StateTransition.Misc do
...
+    result =
+      unless index < index_count and index_count > 0 do
+        {:error, "index not less than index count"}
+      else
+        shuffle_round_count = ChainSpec.get("SHUFFLE_ROUND_COUNT")
+
+        new_index =
+          Enum.reduce(0..(shuffle_round_count - 1), index, fn round, current_index ->
+            round_as_bytes =
+              <<round>> |> :binary.decode_unsigned() |> :binary.encode_unsigned(:little)
```
MegaRedHand, line 31: Since round is a byte long, it has no endianness

lib/lambda_ethereum_consensus/state_transition/misc.ex
```diff
@@ -12,6 +13,69 @@ defmodule LambdaEthereumConsensus.StateTransition.Misc do
...
+
+            seed_as_bytes = seed |> :binary.decode_unsigned() |> :binary.encode_unsigned(:little)
+
+            hash_of_seed_round =
+              :crypto.hash(:sha256, seed_as_bytes <> round_as_bytes)
+              |> :binary.decode_unsigned()
+              |> :binary.encode_unsigned(:little)
+
+            first_8_bytes_of_hash_of_seed_round =
+              hash_of_seed_round |> :binary.bin_to_list({0, 8}) |> :binary.list_to_bin()
```
MegaRedHand, line 41: I suggest reading about binary pattern matching with `<<>>`: https://hexdocs.pm/elixir/1.12/Kernel.SpecialForms.html#%3C%3C%3E%3E/1
```suggestion
            <<first_8_bytes_of_hash_of_seed_round::binary-size(8), _::binary>> = hash_of_seed_round
```

MegaRedHand (commented):
Left some comments. Hope it helps!

lib/lambda_ethereum_consensus/state_transition/misc.ex
Godspower-Eze: Since, the hash function takes in binary data and the hash of value in little endian and big endian differs, doesn't it matter in this case?

Godspower-Eze: Still a bit unclear but I will read more on it. Thanks.

lib/lambda_ethereum_consensus/state_transition/misc.ex
```diff
@@ -58,4 +108,17 @@ defmodule LambdaEthereumConsensus.StateTransition.Misc do
     slots_per_epoch = ChainSpec.get("SLOTS_PER_EPOCH")
     epoch * slots_per_epoch
   end
+
+  @spec bytes_to_uint64(binary()) :: SszTypes.uint64()
+  defp bytes_to_uint64(value) do
+    # Converts a binary value to a 64-bit unsigned integer
+    <<first_8_bytes::binary-size(8), _::binary>> = value
+    first_8_bytes |> :binary.decode_unsigned(:little)
```
MegaRedHand, line 116: You can decode binaries as numbers as part of a match (and it can be moved to the function signature, too).
```suggestion
    <<first_8_bytes::unsigned-integer-little-size(64), _::binary>> = value
    first_8_bytes
```

lib/lambda_ethereum_consensus/state_transition/misc.ex
```diff
@@ -58,4 +108,17 @@ defmodule LambdaEthereumConsensus.StateTransition.Misc do
...
+  end
+
+  @spec uint_to_bytes4(integer()) :: SszTypes.bytes4()
+  defp uint_to_bytes4(value) do
+    # Converts an unsigned integer value to a bytes 4 value
+    <<value::32>> |> :binary.bin_to_list() |> Enum.reverse() |> :binary.list_to_bin()
```
MegaRedHand, line 122: You can also encode binaries using pattern matching
```suggestion
    <<value::unsigned-integer-little-size(32)>>
```

lib/lambda_ethereum_consensus/state_transition/misc.ex
```diff
@@ -12,6 +13,55 @@ defmodule LambdaEthereumConsensus.StateTransition.Misc do
...
+        flip = rem(pivot + index_count - current_index, index_count)
+        position = max(current_index, flip)
+
+        position_div_256 = uint_to_bytes4(div(position, 256))
+
+        source =
+          :crypto.hash(:sha256, seed <> round_as_bytes <> position_div_256)
+
+        byte_index = div(rem(position, 256), 8)
+        byte = source |> :binary.bin_to_list() |> Enum.fetch!(byte_index)
```
MegaRedHand, line 48: We can use pattern matching here
```suggestion
        <<_::binary-size(byte_index - 1), byte, _::binary>> = source
```
(this matches the first `byte_index - 1` bytes, binds the next byte to `byte`, and discards the rest)

test/spec/runners/shuffling.ex
```diff
@@ -0,0 +1,56 @@
...
+      result = Misc.compute_shuffled_index(index, index_count, seed)
+
+      case result do
+        {:ok, value} ->
+          assert value == Enum.fetch!(indices, index)
+
+        {:error, _} ->
+          assert index >= index_count or index_count == 0
+      end
+    end
```
MegaRedHand, line 54: What's the difference between these two cases? Wouldn't they be always equivalent since the function is pure?

MegaRedHand (commented):
Good work! Left some comments about pattern matching, and a question about the tests.

test/spec/runners/shuffling.ex
Godspower-Eze: It's specified in the spec that should be checked and it kinda shows that function is independent of the order.

lib/lambda_ethereum_consensus/state_transition/misc.ex
Godspower-Eze: Great! Had to change `byte_index - 1` to `byte_index`

test/spec/runners/shuffling.ex
MegaRedHand: Could you share a link to that? Also, we should add a code comment with this info.

Godspower-Eze: https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/shuffling. Number 2

MegaRedHand: We should add a comment with that info, for example:

> NOTE: this is the same as before, but the test-format specs suggest including this check

MegaRedHand (approved):
🔥 Excellent work! 🔥

Godspower-Eze:
> 🔥 Excellent work! 🔥

Thank you!

PR #314, opened 2023-10-17
feat: added `process_historical_summaries_update`

### LINKED ISSUE #263: [epoch processing] Implement `process_historical_summaries_update`
https://github.com/lambdaclass/lambda_ethereum_consensus/blob/66059de3653c17ae620ce2aca2db61c558057f6a/docs/specs/beacon-chain.md?plain=1#L2573-L2583

See Spec tests format: https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/epoch_processing

Acceptance criteria:
- Run corresponding spec tests by removing https://github.com/lambdaclass/lambda_ethereum_consensus/blob/66059de3653c17ae620ce2aca2db61c558057f6a/test/spec/runners/epoch_processing.ex#L20

phklive:
**Motivation**

Core function of the epoch processing

Closes #263

### DIFF
--- a/lib/lambda_ethereum_consensus/state_transition/epoch_processing.ex
+++ b/lib/lambda_ethereum_consensus/state_transition/epoch_processing.ex
@@ -4,8 +4,8 @@ defmodule LambdaEthereumConsensus.StateTransition.EpochProcessing do
   """
 
   alias LambdaEthereumConsensus.StateTransition.Accessors
-  alias SszTypes.HistoricalSummary
   alias SszTypes.BeaconState
+  alias SszTypes.HistoricalSummary
   alias SszTypes.Validator
 
   @spec process_effective_balance_updates(BeaconState.t()) ::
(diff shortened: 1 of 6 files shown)


MegaRedHand:
My guess is that the SSZ NIF doesn't have enough information to process the call. The arguments are a list of binaries and a max size, but those binaries could mean anything. To hash lists of non-structs, you should use `hash_list_tree_root_typed/3`.

phklive:
> My guess is that the SSZ NIF doesn't have enough information to process the call. The arguments are a list of binaries and a max size, but those binaries could mean anything. To hash lists of non-structs, you should use `hash_list_tree_root_typed/3`.

I am not sure what SszType to pass then considering that they are a `list(SszTypes.root())`

MegaRedHand:
As discussed in the daily: we need to add the type implementation to the NIF. Basically:
1. Add a `SszTypes.Root` module (like `SszTypes.Transaction`)
2. Add the type in the Rust side (should work just adding the existing `Root` type to the macro in `utils/mod.rs`)

phklive:
@MegaRedHand if you could give a quick look into my code:

These are my generated roots: 
SCR-20231018-ojpl

These are the required (post state) roots: 
SCR-20231018-ojui

I looked into my code and can't seem to find the error that makes the roots different.

MegaRedHand:
Continued in #370

PR #317, opened 2023-10-18
feat: download initial finalized block and add pending blocks to fork choice.

mpaulucci:
**Description**
We need to add the initial block that corresponds to the beacon state that we got from the checkpoint sync. From that point on we can rely on the pending blocks genserver to fill all the blocks from checkpoint to now

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,7 +78,7 @@ start: compile-native compile-port
 
 # Run an interactive terminal with the main supervisor setup.
 iex: compile-native compile-port
-	iex -S mix run -- --checkpoint-sync https://beaconstate-mainnet.chainsafe.io
+	iex -S mix run -- --checkpoint-sync https://sync-mainnet.beaconcha.in/
 
 # Install mix dependencies.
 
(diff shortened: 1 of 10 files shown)


Makefile
```diff
@@ -78,7 +78,7 @@ start: compile-native compile-port
 
 # Run an interactive terminal with the main supervisor setup.
 iex: compile-native compile-port
-	iex -S mix run -- --checkpoint-sync https://beaconstate-mainnet.chainsafe.io
+	iex -S mix run -- --checkpoint-sync https://sync-mainnet.beaconcha.in/
```
mpaulucci, line 81: changed this because the previous one was returning a state that finalized 3hs ago, vs the new one that returns one finalized 20 min ago.

lib/lambda_ethereum_consensus/beacon/pending_blocks.ex
```diff
@@ -103,20 +103,6 @@ defmodule LambdaEthereumConsensus.Beacon.PendingBlocks do
...
-    # add to fork choice
-    # adding this to make dialyzer happy
-    block_hash == <<>>
-  end
-
-  @spec add_to_fork_choice(SszTypes.BeaconBlock.t()) :: :ok
-  defp add_to_fork_choice(_block) do
-    # TODO
-    :ok
+    Process.send_after(self(), :process_blocks, 100)
```
MegaRedHand, line 106: This changed from 1000 to 100. Was that intended?

MegaRedHand (approved):

PR #320, opened 2023-10-18
feat: send notification when a peer is added

MegaRedHand:
Depends on #310

This is needed to have a record of available peers. The alternative would be retrieving all peers, which is unfeasible at long term.


mpaulucci (approved):

mpaulucci (approved):

Arkenan (approved):
I don't think it's reasonable to have a nil peer handler, but I'll see how this plays out after integration. As for the rest, LGTM.

PR #322, opened 2023-10-18
chore: generate all mainnet test cases

MegaRedHand:
This PR enables generation of non-capella mainnet spec-tests. That way, we can run them on cases like #313 that have vectors only on phase0.


mpaulucci (approved):
nice

PR #324, opened 2023-10-19
feat: migrate existing code to use libp2p port

### LINKED ISSUE #204: Move LibP2P implementation to ports

MegaRedHand:
~Depends on #320~
Closes #204


lib/lambda_ethereum_consensus/application.ex
```diff
@@ -12,18 +12,25 @@ defmodule LambdaEthereumConsensus.Application do
     {args, _remaining_args, _errors} =
       OptionParser.parse(System.argv(), switches: [checkpoint_sync: :string])
 
-    {:ok, host} = Libp2p.host_new()
-    {:ok, gsub} = Libp2p.new_gossip_sub(host)
+    config = Application.get_env(:lambda_ethereum_consensus, :discovery)
+    port = Keyword.get(config, :port, 9000)
+    bootnodes = Keyword.get(config, :bootnodes, [])
```
mpaulucci, line 17: In general I think it's better to crash than to add default values. If we pass an empty string as bootnodes it's not going to work anyway and it's probably going to be harder to debug.

lib/lambda_ethereum_consensus/beacon/pending_blocks.ex
```diff
@@ -11,15 +11,14 @@ defmodule LambdaEthereumConsensus.Beacon.PendingBlocks do
   alias LambdaEthereumConsensus.ForkChoice.Store
   alias LambdaEthereumConsensus.P2P.BlockDownloader
 
-  @type state :: %{host: Libp2p.host(), pending_blocks: %{}}
```
mpaulucci, line 14: nice!

lib/lambda_ethereum_consensus/p2p/block_downloader.ex
```diff
@@ -44,26 +44,16 @@ defmodule LambdaEthereumConsensus.P2P.BlockDownloader do
...
-          request_block_by_slot(slot, host, retries - 1)
-        else
-          {:error, reason}
-        end
-    end
+    protocol = @blocks_by_range_protocol_id
+    message = size_header <> compressed_payload
+    error_msg = "Retrying request for block with slot #{slot}"
+
+    retrying_request(peer_id, protocol, message, error_msg, retries)
```
mpaulucci, line 51: this seems out of scope of this PR and will conflict with https://github.com/lambdaclass/lambda_ethereum_consensus/pull/341

mpaulucci (approved):

native/libp2p_port/internal/subscriptions/subscriptions.go
```diff
@@ -21,9 +22,12 @@ type subscription struct {
 }
 
 type Subscriber struct {
-	subscriptions map[string]subscription
-	gsub          *pubsub.PubSub
-	port          *port.Port
+	// NOTE: since subscriptions is accessed only for topic un/subscription
+	// and message publishing, this lock doesn't have much contention
+	subscriptionsMutex *sync.Mutex
```
Arkenan, line 27: When would we need a mutex? AFAIK commands are handled in a sequence, so we wouldn't ever subscribe or unsubscribe concurrently.

Arkenan: If we want to handle commands concurrently (a fairly reasonable thing to do) in goroutines, then we might want to add a separate issue for that.

proto/libp2p.proto
```diff
@@ -87,11 +87,17 @@ message NewPeer {
     bytes peer_id = 1;
 }
 
+message ResultMessage {
+    repeated bytes message = 3;
```
Arkenan, line 91: Is this a raw binary or is it just a string?

lib/lambda_ethereum_consensus/application.ex
MegaRedHand: If I understand correctly, bootnodes aren't required for peer discovery. They are just permanent peers.

lib/lambda_ethereum_consensus/p2p/block_downloader.ex
MegaRedHand: I'll undo the refactor then

lib/lambda_ethereum_consensus/application.ex
MegaRedHand: I removed the default values

lib/lambda_ethereum_consensus/p2p/block_downloader.ex
```diff
@@ -79,49 +69,44 @@ defmodule LambdaEthereumConsensus.P2P.BlockDownloader do
...
-         {:ok, chunk} <- read_response(stream),
-         {:ok, block} <- decode_response(chunk) do
+    protocol = @blocks_by_root_protocol_id
+    message = size_header <> compressed_payload
+    error_msg = "Retrying request for block with root #{Base.encode16(root)}"
+
+    retrying_request(peer_id, protocol, message, error_msg, retries)
+  end
+
+  defp retrying_request(peer_id, protocol, message, error_msg, retries) do
```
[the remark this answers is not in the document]
MegaRedHand: The info is above the `@default_retries` attribute, although we can move it somewhere else.
https://github.com/lambdaclass/lambda_ethereum_consensus/blob/41d1f3b75b2c9621919fc9cd215c197a57d0676b/lib/lambda_ethereum_consensus/p2p/block_downloader.ex#L15-L17

lib/lambda_ethereum_consensus/p2p/incoming_request_handler.ex
```diff
@@ -22,32 +23,28 @@ defmodule LambdaEthereumConsensus.P2P.IncomingRequestHandler do
       "metadata/2"
     ]
     |> Stream.map(&Enum.join([@prefix, &1, "/ssz_snappy"]))
-    |> Stream.map(&Libp2p.host_set_stream_handler(host, &1))
+    |> Stream.map(&Libp2pPort.set_handler/1)
```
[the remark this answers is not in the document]
MegaRedHand: I opened the issue https://github.com/lambdaclass/lambda_ethereum_consensus/issues/343 to address this

lib/lambda_ethereum_consensus/p2p/peerbook.ex
```diff
@@ -26,6 +20,7 @@ defmodule LambdaEthereumConsensus.P2P.Peerbook do
 
   @impl true
   def init(_opts) do
+    Libp2pPort.set_new_peer_handler(self())
```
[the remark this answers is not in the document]
MegaRedHand: I didn't do it that way because I didn't know how to access the child's PID from the supervisor.

proto/libp2p.proto
MegaRedHand: A string when errors happen, and a raw binary on most commands that return a result. Do you think we should make that distinction?

native/libp2p_port/internal/subscriptions/subscriptions.go
MegaRedHand: You're right. Before it was needed, but I think I fixed it when refactoring. I'll remove it.

lib/lambda_ethereum_consensus/p2p/block_downloader.ex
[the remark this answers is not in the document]
Arkenan: Oh nice. I think it's better to have this comment over `retrying_request`. Specifically the part of "trying again with a different peer" is not as clear from reading the code. Over the constant we can have something simple like `Default amount of attempts for a single libp2p request`.

lib/lambda_ethereum_consensus/p2p/peerbook.ex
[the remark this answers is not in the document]
Arkenan: Right, we can probably register the module in the process registry with its name (atom) and use that directly. Again, for a separate PR probably.

proto/libp2p.proto
Arkenan: Ah, makes sense. No worries.

Issue #298, opened 2023-10-12
Check usage of lists in internal state


MegaRedHand:
Since lists in Python are actually arrays, and the spec is written in Python, random accesses are supposed to be constant-time. As we use Elixir (with linked lists), we should check if migrating to something like [`Arrays`](https://hexdocs.pm/arrays/Arrays.html) or [`:array`](https://www.erlang.org/doc/man/array) is needed.

Example:

```elixir
validators = # list of structs
indices = # list of indices
sum =
  Enum.reduce(indices, 0, fn index, acc ->
    acc + Enum.at(validators, index).effective_balance
  end)
```

MegaRedHand:
Addressed in latest commits

PR #327, opened 2023-10-20
feat: implement process registry updates

### LINKED ISSUE #257: [epoch processing] Implement `process_registry_updates`
https://github.com/lambdaclass/lambda_ethereum_consensus/blob/66059de3653c17ae620ce2aca2db61c558057f6a/docs/specs/beacon-chain.md?plain=1#L2439-L2462

Acceptance criteria:
- Run corresponding spec tests by removing https://github.com/lambdaclass/lambda_ethereum_consensus/blob/66059de3653c17ae620ce2aca2db61c558057f6a/test/spec/runners/epoch_processing.ex#L14

f3r10:
**Motivation**

Closes #257 

**Description**

Added the function `process_registry_updates`


config/configs/minimal.yaml
```diff
@@ -83,7 +83,7 @@ INACTIVITY_SCORE_RECOVERY_RATE: 16
 # 2**4 * 10**9 (= 16,000,000,000) Gwei
 EJECTION_BALANCE: 16000000000
 # [customized] more easily demonstrate the difference between this value and the activation churn limit
-MIN_PER_EPOCH_CHURN_LIMIT: 2
+MIN_PER_EPOCH_CHURN_LIMIT: 4
```
f3r10, line 86: this is the constant that was breaking the minimal tests. I looked for other client consensus clients, and I think the value should be 4 now.

MegaRedHand: Should be fixed by #328. We were using the wrong config files 👀

lib/lambda_ethereum_consensus/state_transition/epoch_processing.ex
```diff
@@ -83,6 +84,108 @@ defmodule LambdaEthereumConsensus.StateTransition.EpochProcessing do
...
 
+  @spec process_registry_updates(BeaconState.t()) :: {:ok, BeaconState.t()} | {:error, binary()}
+  def process_registry_updates(%BeaconState{validators: validators} = state) do
+    ejection_balance = ChainSpec.get("EJECTION_BALANCE")
+    churn_limit = Accessors.get_validator_churn_limit(state)
+    churn_limit_range = 0..(churn_limit - 1)
+
+    validators_keyword =
+      0..length(validators)
+      |> Stream.zip(validators)
```
MegaRedHand, line 95: You can use `Stream.with_index()`

lib/lambda_ethereum_consensus/state_transition/epoch_processing.ex
```diff
@@ -83,6 +84,108 @@ defmodule LambdaEthereumConsensus.StateTransition.EpochProcessing do
...
+      |> Keyword.new(fn {index, v} -> {:"#{index}", v} end)
```
MegaRedHand, line 96: We **mustn't** create atoms dynamically. The Elixir VM has a maximum amount of atoms that can be created and terminates if this limit is reached.

test/spec/runners/epoch_processing.ex
```diff
@@ -68,18 +68,31 @@ defmodule EpochProcessingTestRunner do
...
     assert result == {:ok, post}
   end
 
+  defp handle_case("registry_updates", pre, post) do
+    result = EpochProcessing.process_registry_updates(pre)
+
+    case post do
+      nil ->
+        assert {output, _error_msg} = result
+        assert output == :error
```
MegaRedHand, line 82: Matching this way we get better error messages:
```suggestion
        assert {:error, _error_msg} = result
```

lib/lambda_ethereum_consensus/state_transition/epoch_processing.ex
f3r10: Oh ok, I am going to use just a simple list and the  `replace_at` function :sweat_smile:

f3r10: I changed to use a list but maybe performance-wise is not the best option because the list of validators could be quite large. Perhaps a  better option would be to use a native `:array` from erlang.? https://www.theerlangelist.com/article/sequences

lib/lambda_ethereum_consensus/state_transition/mutators.ex
```diff
@@ -0,0 +1,65 @@
...
+        if exit_queue_churn >= Accessors.get_validator_churn_limit(state) do
+          exit_queue_epoch + 1
+        else
+          exit_queue_epoch
+        end
+
+      next_withdrawable_epoch = exit_queue_epoch + min_validator_withdrawability_delay
+
+      if next_withdrawable_epoch > 2 ** 64 - 1 do
+        {:error, :withdrawable_epoch_too_large}
```
MegaRedHand, line 54: We can use a binary here, for consistency.

config/configs/minimal.yaml
MegaRedHand: #328 was merged 🚀. Please fix conflicts

lib/lambda_ethereum_consensus/state_transition/epoch_processing.ex
MegaRedHand: Good catch! We are keeping it simple for now, but we have already opened an issue for it (#298).

MegaRedHand (approved):

PR #331, opened 2023-10-23
feat: add gossip validator

### LINKED ISSUE #164: Libp2p: validate messages received from peers
This can be done with [`(*PubSub).RegisterTopicValidator`](https://pkg.go.dev/github.com/libp2p/go-libp2p-pubsub#PubSub.RegisterTopicValidator), which should receive a [`ValidatorEx`](https://pkg.go.dev/github.com/libp2p/go-libp2p-pubsub#ValidatorEx) so we can differentiate between _REJECT_ and _IGNORE_ actions.

MegaRedHand:
Depends on #324
Closes #164

### DIFF
--- a/lib/lambda_ethereum_consensus/p2p/subscriber.ex
+++ b/lib/lambda_ethereum_consensus/p2p/subscriber.ex
@@ -40,9 +40,11 @@ defmodule LambdaEthereumConsensus.P2P.Subscriber do
 
   @impl true
   def handle_info(
-        {:gossipsub, {topic, message}},
+        {:gossipsub, {topic, msg_id, message}},
         %{topic: topic, queue: queue, demand: demand} = state
       ) do
+    # TODO: validate
+    Libp2pPort.validate_message(msg_id, :accept)
     queue = :queue.in(message, queue)
     pop_events(demand, [], %{state | queue: queue, demand: 0})
   end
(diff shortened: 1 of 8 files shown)


lib/lambda_ethereum_consensus/p2p/subscriber.ex
```diff
@@ -40,9 +40,11 @@ defmodule LambdaEthereumConsensus.P2P.Subscriber do
 
   @impl true
   def handle_info(
-        {:gossipsub, {topic, message}},
+        {:gossipsub, {topic, msg_id, message}},
         %{topic: topic, queue: queue, demand: demand} = state
       ) do
+    # TODO: validate
+    Libp2pPort.validate_message(msg_id, :accept)
```
mpaulucci, line 47: is it specified what a valid message means? For example, just that we can decode it or the contents as well?

What are the consequences of sending a reject or ignore? I guess that it is not going to be gossiped to other peers? Will the peer that sent this bad message be penalized somehow?

mpaulucci (approved):

lib/lambda_ethereum_consensus/p2p/subscriber.ex
MegaRedHand: Oops. I forgot to document the function 😋. Will do that

MegaRedHand: Done: 449049e9a854ea13ca64ba2b22ce060377bc6063

mpaulucci: thanks!

PR #332, opened 2023-10-24
feat: add `process_attestation`

### LINKED ISSUE #300: [operations] Implement `process_attestation`
https://github.com/lambdaclass/lambda_ethereum_consensus/blob/dd0a0523e6be2a21fa3f6a96d4466926307d236c/docs/specs/beacon-chain.md?plain=1#L2761-L2795

Acceptance criteria:

sm-stack:
**Description**

Closes #300


sm-stack:
Currently having issue with:
- Comparing aggregation bits length with beacon committee length
- `get_beacon_proposer_index` and `get_attesting_indices` seems to output wrong index

Keep trying to solve issues by debugging into several steps, and comparing with Lighthouse / Prysm implementation.

lib/lambda_ethereum_consensus/state_transition/math.ex
```diff
@@ -0,0 +1,42 @@
+defmodule LambdaEthereumConsensus.StateTransition.Math do
+  @moduledoc """
+  Math functions
+  """
+
+  @spec uint_to_bytes(SszTypes.uint64()) :: binary
+  def uint_to_bytes(value) when is_integer(value) do
+    byte_size = calculate_byte_size(value)
```
MegaRedHand, line 8: The problem with `get_beacon_proposer_index` is probably here. Elixir integers are big-nums, so their size is not fixed. Because of that, we can't use a single `uint_to_bytes` function, and need multiple `uintN_to_bytes` functions. See:
https://github.com/lambdaclass/lambda_ethereum_consensus/blob/cbb9dc206e8602dfd27636ed35910c33e4933b52/lib/lambda_ethereum_consensus/state_transition/misc.ex#L121-L132

MegaRedHand (commented):
Left some comments. Hope it helps!

lib/lambda_ethereum_consensus/state_transition/math.ex
sm-stack: Great, thanks. I'll try fix that part. 
Then will it be good to implement that in `accessor.ex` as a private function, or make a separate .ex file for that like `math.ex`?

MegaRedHand: You can add them in `misc.ex` or make the existing ones public. We'll probably reorganize all of these modules anyway.

sm-stack: Got it!

sm-stack:
Fixed some issues, now facing 2 more errors
- 'invalid', 'incorrect' tests does not return `{:error, _}`. Maybe there's a problem with error handling logic.
- `epoch_participation` and `balance` is updated as a value less than the correct value. Maybe `get_attestation_participation_flag_indices` return shorter lists than expected, so that the values are underrated.

sm-stack:
Completed initial implementation of `process_attestation`!

Now have one question and one thought about this:
- Should tests prefixed by `incorrect_` (e.g. `incorrect_head_included_at_epoch_delay`) return error? Seems there's no `assert` logic for them, so they just pass without generating errors. Am I missing something?
- To avoid errors from `make lint`, I refactored `process_attestation` function a lot and I think it might not be good for code readability, or consistency. I appreciate any feedbacks not only on this, but anything else that can improve the commits!

mpaulucci:
@sm-stack regarding the first point, if the tests is supposed to fail then post should be `nil`.

Regarding the linting errors, I see these: `Function is too complex`, `Function body is nested too deep`. Are these the ones you are refering to?

sm-stack:
> @sm-stack regarding the first point, if the tests is supposed to fail then post should be `nil`.
> 
> Regarding the linting errors, I see these: `Function is too complex`, `Function body is nested too deep`. Are these the ones you are refering to?

1. I'll re-check that for the tests I referred to.
2. Yep they were I was facing, so I had to divide functions into smaller ones

mpaulucci:
> > @sm-stack regarding the first point, if the tests is supposed to fail then post should be `nil`.
> > Regarding the linting errors, I see these: `Function is too complex`, `Function body is nested too deep`. Are these the ones you are refering to?
> 
> 1. I'll re-check that for the tests I referred to.
> 2. Yep they were I was facing, so I had to divide functions into smaller ones

2. Splitting a function into smaller ones should generally improve readability. Could you point me out to the commit where you did this?

sm-stack:
> > > @sm-stack regarding the first point, if the tests is supposed to fail then post should be `nil`.
> > > Regarding the linting errors, I see these: `Function is too complex`, `Function body is nested too deep`. Are these the ones you are refering to?
> > 
> > 
> > 
> > 1. I'll re-check that for the tests I referred to.
> > 2. Yep they were I was facing, so I had to divide functions into smaller ones
> 
> 2. Splitting a function into smaller ones should generally improve readability. Could you point me out to the commit where you did this?

Did that on [this commit](https://github.com/lambdaclass/lambda_ethereum_consensus/pull/332/commits/518714768168a0648f44e1b46b198282f84eedef). Felt that there are too many private / helper functions.

lib/lambda_ethereum_consensus/state_transition/accessors.ex
```diff
@@ -205,45 +205,60 @@ defmodule LambdaEthereumConsensus.StateTransition.Accessors do
...
-              [Constants.timely_head_flag_index()]
-            else
-              []
-            end
-
-          {:ok, source_indices ++ target_indices ++ head_indices}
-        else
-          {:error, "Attestation source does not match justified checkpoint"}
-        end
+        determine_flag_indices(
```
mpaulucci, line 209: maybe, instead of this function that doesn't really add anything, you could have 3 helper functions

compute_source_indices
compute_target_indices
compute_head_indices

lib/lambda_ethereum_consensus/state_transition/misc.ex
```diff
@@ -152,27 +149,29 @@ defmodule LambdaEthereumConsensus.StateTransition.Misc do
...
-        {:ok, acc_list} ->
-          case compute_shuffled_index(i, length(indices), seed) do
-            {:ok, shuffled_index} ->
-              {:ok, [Enum.at(indices, shuffled_index) | acc_list]}
-
-            {:error, _} = error ->
-              error
-          end
+    result =
+      Enum.reduce_while(start_..end_, {:ok, []}, fn i, acc ->
```
mpaulucci, line 154: I would move the `Enum.reduce_while` into `compute_committee_indices` since the function is coupled to the concept of a reducer (uses :halt, :cont)

lib/lambda_ethereum_consensus/state_transition/operations.ex
```diff
@@ -42,53 +41,18 @@ defmodule LambdaEthereumConsensus.StateTransition.Operations do
...
-        end)
-
-      # Reward proposer
-      proposer_reward_denominator =
-        ((Constants.weight_denominator() - Constants.proposer_weight()) *
-           Constants.weight_denominator())
-        |> div(Constants.proposer_weight())
-
-      proposer_reward = div(proposer_reward_numerator, proposer_reward_denominator)
+        update_epoch_participation(
```
mpaulucci, line 48: this function extraction improves code readability IMO

lib/lambda_ethereum_consensus/state_transition/accessors.ex
sm-stack: Cool, just refactored it

lib/lambda_ethereum_consensus/state_transition/accessors.ex
```diff
@@ -121,4 +125,294 @@ defmodule LambdaEthereumConsensus.StateTransition.Accessors do
...
+    end
+  end
+
+  @doc """
+  Return the block root at the start of a recent ``epoch``.
+  """
+  @spec get_block_root(BeaconState.t(), SszTypes.epoch()) ::
+          {:ok, SszTypes.root()} | {:error, binary()}
+  def get_block_root(state, epoch) do
+    case get_block_root_at_slot(state, Misc.compute_start_slot_at_epoch(epoch)) do
```
mpaulucci, line 278: same

lib/lambda_ethereum_consensus/state_transition/misc.ex
```diff
@@ -130,4 +133,130 @@ defmodule LambdaEthereumConsensus.StateTransition.Misc do
...
+
+  @spec compute_committee(
+          list(SszTypes.validator_index()),
+          SszTypes.bytes32(),
+          SszTypes.uint64(),
+          SszTypes.uint64()
+        ) ::
+          {:ok, list(SszTypes.validator_index())} | {:error, binary()}
+  def compute_committee(indices, seed, index, count) do
+    start_ = div(length(indices) * index, count)
```
mpaulucci, line 150: why `start_` and not `start`?

lib/lambda_ethereum_consensus/state_transition/predicates.ex
```diff
@@ -60,4 +63,45 @@ defmodule LambdaEthereumConsensus.StateTransition.Predicates do
...
+            state,
+            Constants.domain_beacon_attester(),
+            indexed_attestation.data.target.epoch
+          )
+
+        signing_root = Misc.compute_signing_root(indexed_attestation.data, domain)
+        Bls.fast_aggregate_verify(pubkeys, signing_root, indexed_attestation.signature)
+
+      false ->
+        {:error, "Invalid"}
```
mpaulucci, line 104: nit: better error message

test/spec/runners/operations.ex
```diff
@@ -89,4 +91,16 @@ defmodule OperationsTestRunner do
 
     assert true
   end
+
+  defp handle_case("attestation", pre, operation, post, case_dir) do
+    dir_split = String.split(case_dir, "/")
+    test_name = List.last(dir_split)
+
+    if String.starts_with?(test_name, "invalid") do
```
mpaulucci, line 99: can you do `if post == nil` instead?

mpaulucci (commented):
A bunch of nits but looking really good! 🎉

test/spec/runners/operations.ex
sm-stack: Got it!

lib/lambda_ethereum_consensus/state_transition/misc.ex
sm-stack: That was for naming convention, since if I use `end` instead of `end_` at the next line it generates error.

lib/lambda_ethereum_consensus/state_transition/accessors.ex
```diff
@@ -121,4 +125,294 @@ defmodule LambdaEthereumConsensus.StateTransition.Accessors do
...
+  @doc """
+  Return the beacon committee at ``slot`` for ``index``.
+  """
+  @spec get_beacon_committee(BeaconState.t(), SszTypes.slot(), SszTypes.commitee_index()) ::
+          {:ok, list(SszTypes.validator_index())} | {:error, binary()}
+  def get_beacon_committee(state, slot, index) do
+    epoch = Misc.compute_epoch_at_slot(slot)
+    committees_per_slot = get_committee_count_per_slot(state, epoch)
+
+    case Misc.compute_committee(
```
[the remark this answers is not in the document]
sm-stack: But what if `compute_committee_indices` generates error due to error defined at `compute_shuffled_index`? I remember that some of mainnet tests were designed to generate error at that part, so I had to apply `case` conditional.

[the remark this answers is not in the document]
sm-stack: That's definitely a kind of dup, you're right. Thanks!

lib/lambda_ethereum_consensus/state_transition/accessors.ex
sm-stack: Thanks :)

lib/spec/runners/operations.ex
```diff
@@ -2,27 +2,11 @@ defmodule OperationsTestRunner do
...
-    "block_header",
-    "deposit",
-    "proposer_slashing",
-    "voluntary_exit",
-    "sync_aggregate",
-    # "execution_payload",
-    "withdrawals",
-    "bls_to_execution_change"
-  ]
-
```
sm-stack, line 25: Seems that it's a kind of dups, removed it

mpaulucci (approved):
Congrats, great work!

mpaulucci:
@all-contributors add @sm-stack for code

PR #334, opened 2023-10-24
perf: move spec test generation to runners

### LINKED ISSUE #333: [Spec test perf] Make spec test generators per-runner
A first step towards having more perfomant spec tests is having them generate tests if a runner changed instead of always generating all cases. If someone is working on a specific runner, this will make their modifications only regenerate the runners test instead of all.

Part of #288

Arkenan:
**Changes**

- Moves spec test generation to the runner instead of the whole config.
- The chainspec application env is put in each test so that we don't need per-config modules.

Closes #333


Makefile
```diff
@@ -60,33 +60,43 @@ SPECTEST_VERSION := $(shell cat .spectest_version)
...
 $(SPECTEST_ROOTDIR)/%_${SPECTEST_VERSION}.tar.gz:
 	curl -L -o "$@" \
 		"https://github.com/ethereum/consensus-spec-tests/releases/download/${SPECTEST_VERSION}/$*.tar.gz"
 
-$(SPECTEST_ROOTDIR)/tests/%: $(SPECTEST_ROOTDIR)/%_${SPECTEST_VERSION}.tar.gz
+$(VECTORS_DIR)/%: $(SPECTEST_ROOTDIR)/%_${SPECTEST_VERSION}.tar.gz
 	-rm -rf $@
 	tar -xzmf "$<" -C $(SPECTEST_ROOTDIR)
 
+$(SPECTEST_GENERATED_ROOTDIR): $(VECTORS_DIR)
```
MegaRedHand, line 80: This results in "there's no target to generate `test/spec/vectors/tests`". We could make this depend on all subdirs:
```suggestion
$(SPECTEST_GENERATED_ROOTDIR)/%: $(VECTORS_DIR)/%
```
And then use `$(SPECTEST_GENERATED)` instead of `$(SPECTEST_GENERATED_ROOTDIR)` in the remaining targets.

Makefile
```diff
@@ -60,33 +60,43 @@ SPECTEST_VERSION := $(shell cat .spectest_version)
...
 
 clean-vectors:
 	-rm -rf $(SPECTEST_ROOTDIR)/tests
 	-rm $(SPECTEST_ROOTDIR)/*.tar.gz
 
+clean-tests:
+	-rm -r test/generated
+
+gen-spec:
+	mix generate_spec_tests
```
MegaRedHand, line 93: Shouldn't this be `gen-spec: $(SPECTEST_GENERATED)`?

Makefile
```diff
@@ -119,11 +129,29 @@ deps:
 test: compile-native compile-port $(PROTOBUF_EX_FILES)
 	mix test --no-start --exclude spectest
 
-spec-test: compile-port $(PROTOBUF_EX_FILES) $(SPECTEST_DIRS)
-	mix test --no-start --only implemented_spectest
+# Run all spec tests
+spec-test: compile-port $(PROTOBUF_EX_FILES) $(SPECTEST_GENERATED_ROOTDIR)
```
MegaRedHand, line 133: We should add these to `.PHONY`

MegaRedHand (commented):
Since this is a performance PR, we should add some benchmark results.

Makefile
Arkenan: Changed this so it depends on each vector dir and on the script.

Makefile
Arkenan: Fixed. Depending on the generated dir.

Makefile
Arkenan: Done.

MegaRedHand (approved):

PR #340, opened 2023-10-24
feat: add checkpoint-sync make target

MegaRedHand:


Makefile
```diff
@@ -79,12 +79,16 @@ $(OUTPUT_DIR)/libp2p_port: $(PORT_SOURCES)
```
mpaulucci, line 79: should we add something in the README? In the installing and running part.

MegaRedHand: Yes. Good catch!

mpaulucci (approved):

PR #341, opened 2023-10-24
feat: allow to ask peers for many blocks at once.

mpaulucci:
**Motivation**
We were just supporting to ask a single block. With this change, we will be able to ask for several, both per root and per slot


lib/lambda_ethereum_consensus/p2p/block_downloader.ex
```diff
@@ -64,13 +64,21 @@ defmodule LambdaEthereumConsensus.P2P.BlockDownloader do
...
+          {:ok, [SszTypes.SignedBeaconBlock.t()]} | {:error, binary()}
+  def request_blocks_by_root(roots, host, retries \\ @default_retries) do
+    Logger.debug("requesting block for roots #{Enum.map_join(roots, ", ", &Base.encode16/1)}")
 
     peer_id = get_some_peer()
 
     # TODO ssz encode array of roots
     # {:ok, encoded_payload} = payload |> Ssz.to_ssz()
-    encoded_payload = root
+    encoded_payload = Enum.reduce(roots, <<>>, fn root, acc -> acc <> root end)
```
MegaRedHand, line 81: nit. This is the same as `Enum.join`

lib/lambda_ethereum_consensus/p2p/block_downloader.ex
```diff
@@ -82,46 +90,68 @@ defmodule LambdaEthereumConsensus.P2P.BlockDownloader do
...
+  @spec read_response(Libp2p.stream()) :: {:ok, [binary()]} | {:error, binary()}
+  def read_response(stream) do
     fork_context = @fork_context
 
-    case result do
-      {:ok, ""} ->
-        {:error, "unexpected EOF"}
+    stream
+    |> Libp2p.Stream.from()
+    |> Enum.reduce({:ok, []}, fn
```
MegaRedHand, line 126: `Libp2p.Stream`'s chunks aren't related to `response_chunks`. Are we sure this works?

mpaulucci: Added some comment to the code to clarify

lib/lambda_ethereum_consensus/p2p/block_downloader.ex
```diff
@@ -82,46 +90,71 @@ defmodule LambdaEthereumConsensus.P2P.BlockDownloader do
...
     fork_context = @fork_context
 
-    case result do
-      {:ok, ""} ->
-        {:error, "unexpected EOF"}
+    stream
+    |> Libp2p.Stream.from()
+    # Takes a list of N arbitrary binary chunks and returns a list of M <chunk>s
+    # For example:
+    # [<<0, fork_context, 1>>, <<2, 3>>, <<0, fork_context>>, <<4>>, <<5>>] -> [<<1, 2, 3>>, <<4, 5>>]
```
MegaRedHand, line 128: Wouldn't this fail when headers aren't at chunk start? For example:

```
[<<0, fork_context, 1>>, <<2, 3, 0>>, <<fork_context>>, <<4, 5>>] -> [<<1, 2, 3, 0, fork_context, 4, 5>>]
```

mpaulucci: Yes, they would. But Im getting that the chunks from the stream don't come in total arbitrary order, because if that's the case would the Snappy streaming decompression work?

MegaRedHand (approved):
Talking with @mpaulucci, it seems the headers are arriving at chunk start because of the underlying `Read` returning them in different chunks (probably because of peer delay between each response-chunk sent). Since it works, we should merge it, but we still need to implement this in a less hacky way sometime in the future.

lib/lambda_ethereum_consensus/p2p/block_downloader.ex
```diff
@@ -88,29 +96,43 @@ defmodule LambdaEthereumConsensus.P2P.BlockDownloader do
...
 
     case response_chunk do
-      "" ->
+      <<>> ->
         {:error, "unexpected EOF"}
 
-      <<0, ^fork_context::binary-size(4)>> <> chunk ->
-        {:ok, chunk}
+      <<0, ^fork_context::binary-size(4)>> <> rest ->
+        result = rest |> :binary.split(<<0, fork_context::binary-size(4)>>)
```
MegaRedHand, line 130: You need to specify the `:global` option for it to work on all matches.
```suggestion
        result = rest |> :binary.split(<<0, fork_context::binary-size(4)>>, [:global])
```

lib/lambda_ethereum_consensus/p2p/block_downloader.ex
```diff
@@ -135,18 +157,38 @@ defmodule LambdaEthereumConsensus.P2P.BlockDownloader do
...
+  @spec decode_chunks([binary()]) :: {:ok, [SszTypes.SignedBeaconBlock.t()]} | {:error, binary()}
+  defp decode_chunks(chunks) do
+    results =
+      chunks
+      |> Enum.map(&decode_chunk/1)
+
+    if Enum.all?(results, fn
+         {:ok, _} -> true
+         _ -> false
+       end) do
```
MegaRedHand, line 169: We can leave it for a later PR, but we should check each block independently. The peer could have only _some_ of the blocks we requested and return an error on those missing.

lib/lambda_ethereum_consensus/p2p/block_downloader.ex
```diff
@@ -88,29 +96,43 @@ defmodule LambdaEthereumConsensus.P2P.BlockDownloader do
...
+      else
+        {:ok, blocks}
+      end
     else
       {:error, reason} ->
         if retries > 0 do
-          Logger.debug("Retrying request for block with root #{Base.encode16(root)}")
-          request_block_by_root(root, retries - 1)
+          Logger.debug(
+            "Retrying request for block with roots #{Enum.map_join(roots, ", ", &Base.encode16/1)}"
```
MegaRedHand, line 110:
```suggestion
            "Retrying request for blocks with roots #{Enum.map_join(roots, ", ", &Base.encode16/1)}"
```

MegaRedHand (approved):

lib/lambda_ethereum_consensus/p2p/block_downloader.ex
mpaulucci: Hmm yeah, this is tricky. The problem is that then the caller would have to process the response the data and have potentially complex logic to deal with the case in which some blocks are there and some are not

MegaRedHand: I opened #364 for this

PR #345, opened 2023-10-25
chore: make protobuf files an explicit target in Makefile

MegaRedHand:
This is a continuation of https://github.com/lambdaclass/lambda_ethereum_consensus/pull/342. This PR inlines the `make_protos.sh` script and makes the generated files explicit targets of the Makefile, so the build system regenerates them automatically and only when needed.


Arkenan (approved):

Makefile
```diff
@@ -85,35 +102,35 @@ compile-port: $(OUTPUT_DIR)/libp2p_port
...
 	iex -S mix phx.server
 
 # Run an interactive terminal with the main supervisor setup.
-iex: compile-native compile-port
+iex: $(PROTOBUF_EX_FILES) compile-native compile-port
 	iex -S mix
 
 # Run an interactive terminal using checkpoint sync.
-checkpoint-sync: compile-native compile-port
+checkpoint-sync: $(PROTOBUF_EX_FILES) compile-native compile-port
```
mpaulucci, line 113: I see `$(PROTOBUF_EX_FILES) compile-native compile-port` repeated across several commands. Can't we just create another command, for example "compile" that aggregates these?

MegaRedHand: Yeah. That's a good idea!

PR #349, opened 2023-10-26
feat: reimplement ssz library in elixir

f3r10:
**Motivation**

Reimplement the SSZ library in Elixir

**Description**

An initial approach


lib/ssz.ex
```diff
@@ -139,4 +154,305 @@ defmodule Ssz do
...
   @spec decode_u256(binary) :: non_neg_integer
   def decode_u256(num), do: :binary.decode_unsigned(num, :little)
+
+  ##### elixir native ssz
+  @bytes_per_length_offset 4
+  @bytes_per_chunk 32
+  @bits_per_byte 8
+
+  ### ENCODE ######
+  def serialize(%struct{} = map) when is_struct(map) do
```
MegaRedHand, line 164: `%struct{} = map` already checks if `map` is a struct, so `is_struct` shouldn't be needed.

lib/ssz.ex
```diff
@@ -139,4 +154,305 @@ defmodule Ssz do
...
+      end)
+      |> Enum.to_list()
+
+    serialize_list(elements)
+  end
+
+  def serialize(value) when is_list(value), do: serialize_list(value)
+
+  def serialize(%{type: :uint, size: size, value: value}) when is_integer(value),
+    do: serialize_uint(value, size)
```
MegaRedHand, line 197: This was left from the previous iteration, right? The schema shouldn't have the value embedded in it.

lib/ssz.ex
```diff
@@ -139,4 +154,305 @@ defmodule Ssz do
...
+    struct!(
+      SszTypes.DepositData,
+      %{
+        pubkey: pubkey,
+        withdrawal_credentials: withdrawal_credentials,
+        amount: amount,
+        signature: signature
+      }
+    )
+  end
```
MegaRedHand, line 352: Shouldn't we remove these ones?

lib/ssz_types/beacon_chain/deposit.ex
```diff
@@ -9,6 +9,11 @@ defmodule SszTypes.Deposit do
     :data
   ]
 
+  @schema [
+    %{proof: %{type: :list, schema: %{type: :bytes, size: 32}, max_size: 33, is_variable: false}},
+    %{data: %{type: :struct, schema: SszTypes.DepositData.schema, schema_struct: SszTypes.DepositData}}
```
MegaRedHand, line 14: Isn't it kind of redundant to have both `schema` and `schema_struct`? I think we should drop `schema_struct`.

lib/ssz.ex
```diff
@@ -139,4 +164,349 @@ defmodule Ssz do
...
+      value
+      |> :binary.encode_unsigned(:little)
+      |> String.pad_trailing(div(size, 8), <<0>>)
+
+    {:ok, encoded}
+  end
+
+  ### DECODE ######
+  @spec decode_elixir(binary, module) :: {:ok, struct} | {:error, String.t()}
+  def decode_elixir(bin, schema) do
```
MegaRedHand, line 302: We should rename this to `deserialize`, to be consistent in the naming, and avoid confusion with Elixir structs.

lib/ssz.ex
```diff
@@ -139,4 +164,349 @@ defmodule Ssz do
...
+        key = Enum.at(Map.keys(schema), 0)
+        metadata = Map.get(schema, key)
+        value = Map.get(map, key)
+        {:ok, serialized} = serialize(value, metadata)
+        serialized
+      end)
+      |> :binary.list_to_bin()
+
+    {:ok, serialized}
+  end
```
MegaRedHand, line 189: I think this should be the inverse (general to specific):

```elixir
def serialize(%struct{} = map), do: serialize(map, struct.schema())

def serialize(value, %{type: :struct}) do
  serialized =
    schemas
    |> Enum.map(fn schema ->
      key = Enum.at(Map.keys(schema), 0)
      metadata = Map.get(schema, key)
      value = Map.get(map, key)
      {:ok, serialized} = serialize(value, metadata)
      serialized
    end)
    |> :binary.list_to_bin()
end
```

Also, I think we should change `schema()` to return `%{type: :struct, schema: <current schema()>}`, since we may want to have non-SSZ-container structs like `SszTypes.Transaction`. wdyt? @Arkenan

lib/ssz.ex
```diff
@@ -139,4 +164,349 @@ defmodule Ssz do
...
+
+  def serialize(value) when is_binary(value), do: {:ok, value}
+
+  def serialize(value, []) when is_binary(value), do: {:ok, value}
+
+  def serialize(value, %{type: :struct}) do
```
MegaRedHand, line 207: I think we should use `:container` instead of `:struct`, just to follow the spec names.

lib/ssz.ex
```diff
@@ -139,4 +164,349 @@ defmodule Ssz do
...
+          decode_with_scheme(metadata, key, rest_bytes, items, offsets, position_bin)
+        end)
+
+      {:ok, struct!(schema, items)}
+    end
+  end
+
+  @spec decode_with_scheme(map, atom, binary, map, list, integer()) ::
+          {:cont, {:ok, {binary, %{}}}} | {:halt, {:error, String.t()}}
+  defp decode_with_scheme(schema, key, rest_bytes, items, offsets, position_bin) do
```
MegaRedHand, line 325: Nit
```suggestion
  @spec decode_with_schema(map, atom, binary, map, list, integer()) ::
          {:cont, {:ok, {binary, %{}}}} | {:halt, {:error, String.t()}}
  defp decode_with_schema(schema, key, rest_bytes, items, offsets, position_bin) do
```

lib/ssz.ex
```diff
@@ -139,4 +164,349 @@ defmodule Ssz do
...
+      metadata = Map.get(schema, key)
+      is_variable_size(metadata)
+    end)
+    |> Enum.all?()
+  end
+
+  defp is_variable_size(%{type: :list, is_variable: is_variable}), do: is_variable
+  defp is_variable_size(%{type: :bytes}), do: false
+  defp is_variable_size(%{type: :uint}), do: false
+  defp is_variable_size([]), do: true
```
MegaRedHand, line 491: This clause is just for `SszTypes.Transaction`?

lib/ssz_types/transaction.ex
```diff
@@ -2,4 +2,6 @@ defmodule SszTypes.Transaction do
   @moduledoc """
   Alias for `SszTypes.transaction`. Is used when explicit typing is needed.
   """
+  @schema []
```
MegaRedHand, line 5:
```suggestion
  @schema %{type: :list, schema: %{type: :uint, size: 8}, max_size: 1_073_741_824}  # MAX_BYTES_PER_TRANSACTION
```

lib/ssz.ex
```diff
@@ -139,4 +164,349 @@ defmodule Ssz do
...
+    schema
+    |> Enum.map(fn schema ->
+      key = Enum.at(Map.keys(schema), 0)
+      metadata = Map.get(schema, key)
+      is_variable_size(metadata)
+    end)
+    |> Enum.all?()
+  end
+
+  defp is_variable_size(%{type: :list, is_variable: is_variable}), do: is_variable
```
MegaRedHand, line 488: Lists always have variable size

lib/ssz_types/beacon_chain/deposit.ex
```diff
@@ -9,6 +9,16 @@ defmodule SszTypes.Deposit do
     :data
   ]
 
+  @schema [
+    %{proof: %{type: :list, schema: %{type: :bytes, size: 32}, max_size: 33, is_variable: false}},
```
MegaRedHand, line 13: Vectors are the ones with fixed-size (unless inner values are variable-size).

```suggestion
    %{proof: %{type: :vector, schema: %{type: :bytes, size: 32}, max_size: 33}},
```

MegaRedHand (commented):
Good work! Left some comments about the API and the schemas.

Arkenan (commented):
Hi Fernando,

Thanks for taking this issue, it's a pretty big one, and I see you're doing quite a lot of progress. I'll start being a little more hands-on with this one from tomorrow on. I think some parts of the code are getting a bit confusing. Let's.

1. Start a new PR from a scratch. You can of course reuse whatever logic you need from here, but we'll need smaller and more incremental updates. We'll close this one.
2. Start that PR as a draft until it's ready to be reviewed. 
3.  Make the new SSZ library in a new module. Let's call it SszEx for now. That module shouldn't call any function from rust or from this module.
4. Let's make the first PR only about lists of basic types, no schemas nor containers. Schemas need a bit of rethinking. Let's sync about those in detail tomorrow. The main problem that I want to discuss is big functions with many cases. Ideally we would only have a single simple clause for all containers, and schemas should be recursive, not fully descriptive.

PR #352, opened 2023-10-26
chore: add `compile` Makefile target

MegaRedHand:


Makefile
```diff
@@ -92,17 +92,19 @@ compile-native: $(OUTPUT_DIR)/libp2p_nif.so
 
 compile-port: $(OUTPUT_DIR)/libp2p_port
 
+compile: compile-native compile-port $(PROTOBUF_EX_FILES)
```
mpaulucci, line 95: Ah, I thought compile would include Elixir compilation (compile all), and compile-native would include ports (compile everything native).

WDYT?

MegaRedHand: That's a good idea. I'll try it

mpaulucci (approved):

PR #358, opened 2023-10-27
feat: add peer scoring

MegaRedHand:
WIP. Still needs testing.


MegaRedHand:
It doesn't seem to be helping, according to metrics. I'll drop this one for now.

PR #359, opened 2023-10-30
feat: add process_attester_slashing

### LINKED ISSUE #301: [operations] Implement `process_attester_slashing`
https://github.com/lambdaclass/lambda_ethereum_consensus/blob/dd0a0523e6be2a21fa3f6a96d4466926307d236c/docs/specs/beacon-chain.md?plain=1#L2709-L2724

Acceptance criteria:

karasakalmt:
Closes #301


lib/lambda_ethereum_consensus/beacon/helper_functions.ex
```diff
@@ -29,4 +29,18 @@ defmodule LambdaEthereumConsensus.Beacon.HelperFunctions do
...
   end
+
+  @doc """
+  Return the signing root for the corresponding signing data.
+  """
+  @spec compute_signing_root(any(), SszTypes.domain()) :: SszTypes.root()
+  def compute_signing_root(ssz_object, domain) do
+    {:ok, root} =
+      Ssz.hash_tree_root(%SszTypes.SigningData{
+        object_root: Ssz.hash_tree_root(ssz_object),
```
MegaRedHand, line 40: This returns a result tuple. Also, we should use a `with` and return errors.

lib/lambda_ethereum_consensus/state_transition/accessors.ex
```diff
@@ -121,4 +121,53 @@ defmodule LambdaEthereumConsensus.StateTransition.Accessors do
...
+    proposer_index = Misc.compute_proposer_index(state, indices, seed)
+    proposer_index
+  end
+
+  @doc """
+  Return the signature domain (fork version concatenated with domain type) of a message.
+  """
+  @spec get_domain(BeaconState.t(), SszTypes.domain_type(), SszTypes.epoch() | None) :: SszTypes.domain()
+  def get_domain(state, domain_type, epoch \\ None) do
+    epoch = if epoch == None, do: get_current_epoch(state), else: epoch
```
MegaRedHand, line 164: `None` in Python is equivalent to `nil` in Elixir

```suggestion
    epoch = if epoch == nil, do: get_current_epoch(state), else: epoch
```

native/libp2p_nif/main.h
```diff
@@ -0,0 +1,107 @@
+/* Code generated by cmd/cgo; DO NOT EDIT. */
```
MegaRedHand, line 1: This file can be deleted

lib/spec/runners/operations.ex
```diff
@@ -91,6 +91,15 @@ defmodule OperationsTestRunner do
     assert true
   end
 
+  defp handle_case("attester_slashing", pre, attester_slashing, post, _case_dir) do
```
MegaRedHand, line 94: This shouldn't be needed anymore once it's synced with main

lib/lambda_ethereum_consensus/state_transition/predicates.ex
```diff
@@ -66,44 +64,78 @@ defmodule LambdaEthereumConsensus.StateTransition.Predicates do
...
-
-        signing_root = Misc.compute_signing_root(indexed_attestation.data, domain)
-        Bls.fast_aggregate_verify(pubkeys, signing_root, indexed_attestation.signature)
-
-      false ->
-        {:error, "Attesting Indices are empty or not sorted"}
+      case res do
+        {:ok, r} -> r
+        {:error, _} -> false
+      end
```
MegaRedHand, line 138: Good catch!

lib/lambda_ethereum_consensus/state_transition/operations.ex
```diff
@@ -174,7 +174,56 @@ defmodule LambdaEthereumConsensus.StateTransition.Operations do
...
 
-  @doc """
+  @spec process_attester_slashing(BeaconState.t(), SszTypes.AttesterSlashing.t()) ::
+          {:ok, BeaconState.t()} | {:ok, nil}
+  def process_attester_slashing(state, attester_slashing) do
+    attestation_1 = attester_slashing.attestation_1
+    attestation_2 = attester_slashing.attestation_2
+
+    cond do
+      Predicates.is_slashable_attestation_data(attestation_1.data, attestation_2.data) == false ->
```
MegaRedHand, line 184: nit. Same for other checks
```suggestion
      not Predicates.is_slashable_attestation_data(attestation_1.data, attestation_2.data) ->
```

lib/lambda_ethereum_consensus/state_transition/operations.ex
```diff
@@ -174,7 +174,56 @@ defmodule LambdaEthereumConsensus.StateTransition.Operations do
...
+        {:ok, nil}
```
MegaRedHand, line 185: This should return an error.

lib/lambda_ethereum_consensus/state_transition/operations.ex
```diff
@@ -174,7 +174,56 @@ defmodule LambdaEthereumConsensus.StateTransition.Operations do
...
+
+      Predicates.is_indices_available(length(state.validators), attestation_2.attesting_indices) == false ->
+        {:ok, nil}
+
+      true ->
+        {slashed_any, state} =
+          Enum.uniq(attestation_1.attesting_indices)
+          |> Enum.filter(fn i -> Enum.member?(attestation_2.attesting_indices, i) end)
+          |> Enum.sort()
```
MegaRedHand, line 203: Since they are already sorted, we don't need to sort them again

lib/lambda_ethereum_consensus/state_transition/operations.ex
```diff
@@ -174,7 +174,56 @@ defmodule LambdaEthereumConsensus.StateTransition.Operations do
...
+                Accessors.get_current_epoch(state)
+              ) ->
+                case Mutators.slash_validator(state, i) do
+                  {:ok, state} -> {:cont, {true, state}}
+                  {:error, _msg} -> {:halt, {false, nil}}
+                end
+
+              true ->
+                {:cont, {slashed_any, state}}
+            end
```
MegaRedHand, line 217: We can use a simple if-else here

lib/lambda_ethereum_consensus/state_transition/operations.ex
```diff
@@ -174,7 +174,56 @@ defmodule LambdaEthereumConsensus.StateTransition.Operations do
...
+          end)
+        case slashed_any do
+          false -> {:ok, nil}
+          true -> {:ok, state}
+        end
```
MegaRedHand, line 222: Same here
```suggestion
        if slashed_any do
          {:ok, state}
        else
          {:error, "no slashings"}
        end
```

lib/lambda_ethereum_consensus/state_transition/misc.ex
```diff
@@ -121,15 +121,65 @@ defmodule LambdaEthereumConsensus.StateTransition.Misc do
...
+  @doc """
+  Return from ``indices`` a random index sampled by effective balance.
+  """
+  @spec compute_proposer_index(BeaconState.t(), [SszTypes.validator_index()], SszTypes.bytes32()) ::
+          SszTypes.validator_index() | {:error, binary()}
+  def compute_proposer_index(state, indices, seed) do
+    if not (length(indices) > 0) do
+      {:error, "Empty indices"}
+    else
+      compute_proposer_index(state, indices, seed, 0)
```
MegaRedHand, line 133: This should be wrapped in `{:ok, ...}` for less error-prone matching:

```elixir
# we can do
case compute_proposer_index(...) do
  {:ok, index} -> ...
  {:error, reason} -> ...
end

# instead of
case compute_proposer_index(...) do
  index -> ...     # this is WRONG
  {:error, reason} -> ...
end
```

lib/lambda_ethereum_consensus/state_transition/accessors.ex
```diff
@@ -126,6 +126,27 @@ defmodule LambdaEthereumConsensus.StateTransition.Accessors do
...
+      :crypto.hash(
+        :sha256,
+        get_seed(state, epoch, Constants.domain_beacon_proposer()) <>
+          Misc.uint64_to_bytes(state.slot)
+      )
+
+    indices = get_active_validator_indices(state, epoch)
+    case Misc.compute_proposer_index(state, indices, seed) do
+      {:error, msg} -> {:error, msg}
+      i -> i
```
MegaRedHand, line 146: Same here, it should be wrapped in an `:ok` tuple

MegaRedHand (commented):
Left some more comments

lib/lambda_ethereum_consensus/state_transition/predicates.ex
karasakalmt: thanks!

lib/lambda_ethereum_consensus/state_transition/operations.ex
karasakalmt: In first time I used it with not but it didn't work properly that's why I changed to `== false`. Don't know why but it doesn't work this way

lib/spec/runners/operations.ex
karasakalmt: I will be removing it just before the merge since I keep testing while I am changing the codebase

lib/lambda_ethereum_consensus/state_transition/operations.ex
karasakalmt: Spec tests must be altered for operations for me to return an error since it won't fit with the spec test. I was handling with the errors but expected return value was {:ok, nil} when there is an error. Don't know how can I fix it this way

lib/lambda_ethereum_consensus/state_transition/operations.ex
karasakalmt: This is not actually sorted in case you are mentioning the Enum.uniq() it just removes the duplicates
```elixir
iex(1)> a = [1, 4, 2, 3, 2, 4]
[1, 4, 2, 3, 2, 4]
iex(2)> Enum.uniq(a)
[1, 4, 2, 3]
```

lib/lambda_ethereum_consensus/state_transition/operations.ex
karasakalmt: nvm it was about something else

lib/lambda_ethereum_consensus/state_transition/operations.ex
MegaRedHand: I thought `is_valid_indexed_attestation` checked that the list was sorted and had only unique values.

karasakalmt: you are actually right but 
https://github.com/lambdaclass/lambda_ethereum_consensus/blob/dd0a0523e6be2a21fa3f6a96d4466926307d236c/docs/specs/beacon-chain.md?plain=1#L2719C4-L2719C4
in here sort has been added again I don't know why

lib/spec/runners/operations.ex
MegaRedHand: Disregard the last comment. I thought this was the `epoch_processing` runner 😅

lib/spec/runners/operations.ex
```diff
@@ -91,6 +91,15 @@ defmodule OperationsTestRunner do
...
+    result = Operations.process_attester_slashing(pre, attester_slashing)
+    {:ok, result} = result
+    # if post != nil do
+    #   IO.inspect(LambdaEthereumConsensus.Utils.Diff.diff(result, post))
+    # end
+    assert result == post
```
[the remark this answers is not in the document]
MegaRedHand: You can also use `Diff.diff` for comparing `new_state` and `post`:

```elixir
assert Diff.diff(new_state, post) == :unchanged
```

lib/lambda_ethereum_consensus/state_transition/operations.ex
MegaRedHand: 😮 Let's leave it for now, then.

lib/lambda_ethereum_consensus/state_transition/operations.ex
karasakalmt: Inherited implementation from the previous handlings

karasakalmt: Inherited implementation from the previous handlings

lib/lambda_ethereum_consensus/state_transition/misc.ex
```diff
@@ -121,15 +121,66 @@ defmodule LambdaEthereumConsensus.StateTransition.Misc do
...
+      compute_proposer_index(state, indices, seed, i + 1)
+    end
+  end
+
+  @doc """
+  Return the domain for the ``domain_type`` and ``fork_version``.
+  """
+  @spec compute_domain(SszTypes.domain_type(), Keyword.t()) ::
+          {:ok, SszTypes.domain()}
+  def compute_domain(domain_type, opts \\ []) do
```
MegaRedHand, line 161: This should return the bare value.

lib/lambda_ethereum_consensus/state_transition/operations.ex
```diff
@@ -369,7 +430,7 @@ defmodule LambdaEthereumConsensus.StateTransition.Operations do
   end
 
   defp invalid_signature?(state, indexed_attestation) do
-    Predicates.is_valid_indexed_attestation(state, indexed_attestation) != {:ok, true}
+    Predicates.is_valid_indexed_attestation(state, indexed_attestation) != true
```
MegaRedHand, line 433: Nit
```suggestion
    not Predicates.is_valid_indexed_attestation(state, indexed_attestation)
```

MegaRedHand (commented):
LGTM. Left some minor comments.

MegaRedHand (approved):

PR #363, opened 2023-10-31
feat: allow other peers to query blocks by range

### LINKED ISSUE #275: [P2P] Allow other peers to query blocks by range
We need to implement this handler, in which peers ask for a number of blocks by range. We should fetch these blocks from storage:

https://github.com/lambdaclass/lambda_ethereum_consensus/blob/1aeb3477e9eb940671b1fcc7ecf652d654c9e34f/lib/lambda_ethereum_consensus/p2p/incoming_request_handler.ex#L124

Here is the corresponding spec: https://github.com/lambdaclass/lambda_ethereum_consensus/blob/main/docs/specs/p2p-interface.md#beaconblocksbyrange-v2

Additionally, you should be familiarised with the req/resp specification: https://github.com/lambdaclass/lambda_ethereum_consensus/blob/main/docs/specs/p2p-interface.md#reqresp-interaction

Godspower-Eze:
**Description**

Allow peers request beacon blocks by range

Closes #275


lib/lambda_ethereum_consensus/p2p/incoming_request_handler.ex
```diff
@@ -117,8 +119,46 @@ defmodule LambdaEthereumConsensus.P2P.IncomingRequestHandler do
...
+            :not_found -> current_blocks
+          end
+          if length(blocks) <= ChainSpec.get("MAX_REQUEST_BLOCKS") do
+            {:cont, blocks}
+          else
+            {:halt, blocks}
+          end
+        end)
+
+      blocks_by_range_response = %SszTypes.BeaconBlocksByRangeResponse{body: blocks}
```
MegaRedHand, line 153: This wouldn't work, sadly. The spec shows the response as an SSZ container, but that isn't actually correct. The response is given as a stream of chunks, each corresponding to a block, and prefixed with a header. An example would be:

```
request: 3 blocks: A, B (we don't have B)
response: `0 result (OK)` <> `ForkDigest-context` <> `size varint` <> `A (ssz_snappy encoded)` <>
           `3 result (not found)` <> `error_message`
```
Some relevant links:
- https://github.com/lambdaclass/lambda_ethereum_consensus/blob/d3344795df9d55c52c84da84ef12da8695c57243/docs/specs/p2p-interface.md?plain=1#L770-L771
- https://github.com/lambdaclass/lambda_ethereum_consensus/blob/main/docs/specs/p2p-interface.md?plain=1#L695

MegaRedHand (changes requested):
Left some comments.

MegaRedHand:
Leaving some context here (@Godspower-Eze please correct me if I'm wrong):

Networking configuration wasn't added to the config, but because `MAX_REQUEST_BLOCKS` will be modified (now `MAX_REQUEST_BLOCKS_DENEB`, all of them will be added to the config starting with Deneb. Because of that, we should start adding them manually to the config.

config/configs/mainnet.yaml
```diff
@@ -93,3 +85,8 @@ PROPOSER_SCORE_BOOST: 40
 DEPOSIT_CHAIN_ID: 1
 DEPOSIT_NETWORK_ID: 1
 DEPOSIT_CONTRACT_ADDRESS: 0x00000000219ab540356cBB839Cbe05303d7705Fa
+
+# Networking
+# ---------------------------------------------------------------
+# `2**10` (= 1024)
+MAX_REQUEST_BLOCKS: 1024
```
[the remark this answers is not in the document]
Godspower-Eze: That's the one thing I changed.

config/configs/minimal.yaml
```diff
@@ -78,17 +73,20 @@ MIN_PER_EPOCH_CHURN_LIMIT: 4
...
 # Ethereum Goerli testnet
 DEPOSIT_CHAIN_ID: 5
 DEPOSIT_NETWORK_ID: 5
 # Configured on a per testnet basis
 DEPOSIT_CONTRACT_ADDRESS: 0x1234567890123456789012345678901234567890
+
+# Networking
+# ---------------------------------------------------------------
+# `2**10` (= 1024)
+MAX_REQUEST_BLOCKS: 1024
```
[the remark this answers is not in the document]
Godspower-Eze: That's the only thing I changed here too

lib/lambda_ethereum_consensus/p2p/incoming_request_handler.ex
```diff
@@ -117,8 +119,46 @@ defmodule LambdaEthereumConsensus.P2P.IncomingRequestHandler do
...
+      slot_coverage = start_slot + count
+
+      blocks =
+        start_slot..slot_coverage
+        |> Enum.reduce_while([], fn slot, current_blocks ->
+          blocks = case StateStore.get_state_by_slot(slot) do
+            {:ok, state} -> current_blocks ++ BlockStore.get_blocks(state.block_roots)
+            {:error, _} -> current_blocks
+            :not_found -> current_blocks
+          end
```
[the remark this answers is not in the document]
Godspower-Eze: This returns the latest block received in a slot.

config/configs/mainnet.yaml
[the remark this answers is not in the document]
MegaRedHand: I was talking about the whitespace changes (like [here](https://github.com/lambdaclass/lambda_ethereum_consensus/pull/363/files#diff-1ff307241ed7036b050f1309475f079bf3a87977db78508c5701fd1bd23b8972L54-L56), [here](https://github.com/lambdaclass/lambda_ethereum_consensus/pull/363/files?diff=unified&w=0#diff-1ff307241ed7036b050f1309475f079bf3a87977db78508c5701fd1bd23b8972L43-R40), and [here](https://github.com/lambdaclass/lambda_ethereum_consensus/pull/363/files?diff=unified&w=0#diff-1ff307241ed7036b050f1309475f079bf3a87977db78508c5701fd1bd23b8972L21-L22)). They were probably done by a formatter. If you are using VSCode, you should use [the "save without formatting" command](https://stackoverflow.com/a/63480681) for this

lib/lambda_ethereum_consensus/p2p/incoming_request_handler.ex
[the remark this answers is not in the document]
MegaRedHand: That's how it works now, but in the future, it will return the block corresponding to that slot.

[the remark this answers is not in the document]
Godspower-Eze: Yeah, I was a bit unclear how many blocks should be in a slot. Is this defined?

Godspower-Eze:
> Leaving some context here (@Godspower-Eze please correct me if I'm wrong):
> 
> Networking configuration wasn't added to the config, but because `MAX_REQUEST_BLOCKS` will be modified (now `MAX_REQUEST_BLOCKS_DENEB`, all of them will be added to the config starting with Deneb. Because of that, we should start adding them manually to the config.

Can you please rephrase? I am struggling to understand.

lib/lambda_ethereum_consensus/p2p/incoming_request_handler.ex
Godspower-Eze: result in this case is like a status_code, right?

lib/lambda_ethereum_consensus/p2p/incoming_request_handler.ex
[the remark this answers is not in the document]
MegaRedHand: 0~1 blocks per slot, as per the spec. While fork choice is validating the blocks there might be many, but this is quickly resolved, since:
- either they are blocks proposed by a non-proposer (which are quickly discarded),
- or they are from the proposer and the network should choose only one on the next block (and the proposer will be slashed)

MegaRedHand:
Before, the config files didn't have the networking constants (I guess because they forgot to add them). Since the next fork ("Deneb") modifies some of those, they will be adding the networking constants to the YAMLs. Because of this, and because of those constants being labeled as _configuration_ in the spec, it's safe to have them in the YAMLs.

lib/lambda_ethereum_consensus/p2p/incoming_request_handler.ex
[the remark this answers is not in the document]
Godspower-Eze: Thank you. Please send a link where I can read up.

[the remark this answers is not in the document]
MegaRedHand: The information is in https://eth2book.info/capella/part2/consensus/, but I don't know if there's a single page that explains this.

config/configs/mainnet.yaml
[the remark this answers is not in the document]
mpaulucci: We should maybe configure a yaml formatter so that this doesn't happen?

lib/lambda_ethereum_consensus/p2p/incoming_request_handler.ex
```diff
@@ -117,8 +119,46 @@ defmodule LambdaEthereumConsensus.P2P.IncomingRequestHandler do
     end
   end
 
-  def handle_req("beacon_blocks_by_range/2/ssz_snappy", _message_id, _message),
-    do: :not_implemented
+  def handle_req("beacon_blocks_by_range/2/ssz_snappy", message_id, message) do
+    with <<24, snappy_blocks_by_range_request::binary>> <- message,
```
mpaulucci, line 123: what does the 24 reference to?

Godspower-Eze: That's the length of the `BeaconBlocksByRangeRequest` SSZ Container.

lib/lambda_ethereum_consensus/p2p/incoming_request_handler.ex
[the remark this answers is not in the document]
MegaRedHand: @Godspower-Eze I just found out the `BeaconBlocksByRange` request doesn't send N chunks for N blocks:

https://github.com/lambdaclass/lambda_ethereum_consensus/blob/15617fd64f66daacf52b2e1545ad2a6efb81dd27/docs/specs/p2p-interface.md?plain=1#L898-L899

In this case, we should find a way to represent the "no blocks for this slot" in our DB, and parse that result accordingly here (not returning blocks for empty slots, instead of returning a response code of 3).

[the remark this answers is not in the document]
Godspower-Eze: Gotcha

lib/lambda_ethereum_consensus/p2p/incoming_requests/handler.ex
```diff
@@ -88,8 +99,59 @@ defmodule LambdaEthereumConsensus.P2P.IncomingRequests.Handler do
...
+      ## TODO: there should be check that the `start_slot` is not older than the `oldest_slot_with_block`
+      %SszTypes.BeaconBlocksByRangeRequest{start_slot: start_slot, count: count} =
+        blocks_by_range_request
+
+      count =
+        if count > ChainSpec.get("MAX_REQUEST_BLOCKS") do
+          ChainSpec.get("MAX_REQUEST_BLOCKS")
+        else
+          count
+        end
```
MegaRedHand, line 121: You can use `min` for this

lib/lambda_ethereum_consensus/p2p/incoming_requests/handler.ex
```diff
@@ -88,8 +99,59 @@ defmodule LambdaEthereumConsensus.P2P.IncomingRequests.Handler do
...
+
+      slot_coverage = start_slot + (count - 1)
+
+      blocks =
+        start_slot..slot_coverage
+        |> Enum.map(fn slot -> BlockStore.get_block_by_slot(slot) end)
```
MegaRedHand, line 127: Nit
```suggestion
        |> Enum.map(&BlockStore.get_block_by_slot/1)
```

lib/lambda_ethereum_consensus/p2p/incoming_requests/handler.ex
```diff
@@ -88,8 +99,59 @@ defmodule LambdaEthereumConsensus.P2P.IncomingRequests.Handler do
...
+
+      response_chunk =
+        blocks
+        |> Enum.reduce(<<>>, fn block_response, response_chunk ->
```
MegaRedHand, line 131: You can use `Enum.map` and then `Enum.join` to simplify this. Also, this:

```elixir
fn a ->
  case a do
    {:ok, ...} -> ...
    {:error, ...} -> ...
  end
end
```

is equivalent to this:

```elixir
fn
  {:ok, ...} -> ...
  {:error, ...} -> ...
end
```

lib/lambda_ethereum_consensus/p2p/incoming_requests/handler.ex
```diff
@@ -3,7 +3,18 @@ defmodule LambdaEthereumConsensus.P2P.IncomingRequests.Handler do
...
+  alias LambdaEthereumConsensus.{Libp2pPort, P2P}
+  alias LambdaEthereumConsensus.Store.BlockStore
+
+  # This is the `ForkDigest` for mainnet in the capella fork
+  # TODO: compute this at runtime
+  @fork_context "BBA4DA96" |> Base.decode16!()
+
+  # This is the `Resource Available` error message in binary
+  # TODO: compute this and other messages at runtime
+  @error_message_resourse_avalaible <<?R, ?e, ?s, ?o, ?u, ?r, ?c, ?e, 32, ?U, ?n, ?a, ?v, ?a, ?i,
```
MegaRedHand, line 15: typo

```suggestion
  @error_message_resource_available <<?R, ?e, ?s, ?o, ?u, ?r, ?c, ?e, 32, ?U, ?n, ?a, ?v, ?a, ?i,
```

lib/lambda_ethereum_consensus/store/block_store.ex
```diff
@@ -33,6 +33,21 @@ defmodule LambdaEthereumConsensus.Store.BlockStore do
     end
   end
 
+  @spec get_blocks(list(SszTypes.root())) ::
+          {:ok, list(SszTypes.SignedBeaconBlock.t())} | {:error, String.t()} | :not_found
+  def get_blocks(block_roots) do
```
MegaRedHand, line 38: Is this dead code?

lib/lambda_ethereum_consensus/p2p/incoming_requests/handler.ex
```diff
@@ -88,8 +99,59 @@ defmodule LambdaEthereumConsensus.P2P.IncomingRequests.Handler do
...
+                  |> P2P.Utils.encode_varint()
+
+                response_chunk <> <<0>> <> @fork_context <> size_header <> snappy_ssz_signed_block
+              else
+                {:error, _} ->
+                  response_chunk <> <<2>> <> @error_message_server_error
+              end
+
+            _ ->
+              response_chunk <> <<3>> <> @error_message_resourse_avalaible
```
MegaRedHand, line 148: Error messages are also `ssz_snappy`-encoded

MegaRedHand (commented):
LGTM, but error messages must be SSZ serialized and Snappy compressed according to the spec

lib/lambda_ethereum_consensus/p2p/incoming_requests/handler.ex
```diff
@@ -3,7 +3,17 @@ defmodule LambdaEthereumConsensus.P2P.IncomingRequests.Handler do
...
+  alias LambdaEthereumConsensus.{Libp2pPort, P2P}
+  alias LambdaEthereumConsensus.Store.BlockStore
+
+  # This is the `ForkDigest` for mainnet in the capella fork
+  # TODO: compute this at runtime
+  @fork_context "BBA4DA96" |> Base.decode16!()
+
+  # This is the `Resource Available` error message in binary
+  # TODO: compute this and other messages at runtime
+  @error_message_resource_available "Resource Unavailable"
```
mpaulucci, line 15: nit: `error_message_resource_unavailable`

lib/lambda_ethereum_consensus/p2p/incoming_requests/handler.ex
```diff
@@ -88,12 +98,73 @@ defmodule LambdaEthereumConsensus.P2P.IncomingRequests.Handler do
...
+            size_header =
+              @error_message_server_error
+              |> byte_size()
+              |> P2P.Utils.encode_varint()
+
+            {:ok, snappy_message} = Snappy.compress(@error_message_server_error)
+            <<2>> <> size_header <> snappy_message
+        end
+
+      _ ->
```
mpaulucci, line 159: maybe be more specific here, you need to match `:not_found`

MegaRedHand (commented):
After this comment, I think we're good to merge

lib/lambda_ethereum_consensus/p2p/incoming_requests/handler.ex
MegaRedHand: Also, the comment mentions "Resource Available" when it should be "Resource Unavailable"

MegaRedHand (approved):
The things that are left aren't really blocking (see `resource_available` typo), so I'll just leave my approve here and defer to @mpaulucci

mpaulucci (approved):

PR #366, opened 2023-10-31
fix: store signed blocks instead of regular blocks.

mpaulucci:
**Motivation**
We need to store signed blocks since we will need to pass them to other peers on request.

### DIFF
--- a/lib/lambda_ethereum_consensus/fork_choice/store.ex
+++ b/lib/lambda_ethereum_consensus/fork_choice/store.ex
@@ -8,9 +8,9 @@ defmodule LambdaEthereumConsensus.ForkChoice.Store do
 
   alias LambdaEthereumConsensus.ForkChoice.Helpers
   alias LambdaEthereumConsensus.Store.{BlockStore, StateStore}
-  alias SszTypes.BeaconBlock
   alias SszTypes.BeaconState
   alias SszTypes.Store
+  alias SszTypes.SignedBeaconBlock
 
   ##########################
   ### Public API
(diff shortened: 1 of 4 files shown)


lib/lambda_ethereum_consensus/store/block_store.ex
```diff
@@ -8,13 +8,14 @@ defmodule LambdaEthereumConsensus.Store.BlockStore do
...
     {:ok, block_root} = Ssz.hash_tree_root(block)
-    {:ok, encoded_block} = Ssz.to_ssz(block)
+    {:ok, encoded_signed_block} = Ssz.to_ssz(signed_block)
 
     key = block_key(block_root)
-    Db.put(key, encoded_block)
+    Db.put(key, encoded_signed_block)
 
     # WARN: this overrides any previous mapping for the same slot
     # TODO: this should apply fork-choice if not applied elsewhere
```
Godspower-Eze, line 21: I think `block_root_by_slot_key(block.slot)` should change to `block_root_by_slot_key(block.message.slot)`.

lib/lambda_ethereum_consensus/fork_choice/store.ex
```diff
@@ -50,10 +50,10 @@ defmodule LambdaEthereumConsensus.ForkChoice.Store do
   ##########################
 
   @impl GenServer
-  @spec init({BeaconState.t(), BeaconBlock.t()}) :: {:ok, Store.t()} | {:stop, any}
-  def init({anchor_state = %BeaconState{}, anchor_block = %BeaconBlock{}}) do
+  @spec init({BeaconState.t(), SignedBeaconBlock.t()}) :: {:ok, Store.t()} | {:stop, any}
+  def init({anchor_state = %BeaconState{}, signed_anchor_block = %SignedBeaconBlock{}}) do
     result =
-      case Helpers.get_forkchoice_store(anchor_state, anchor_block) do
+      case Helpers.get_forkchoice_store(anchor_state, signed_anchor_block.message) do
```
MegaRedHand, line 56: I think we should also store the signed version in fork choice.

lib/lambda_ethereum_consensus/store/block_store.ex
mpaulucci: But block is not a signed block in this case, see line 13.

lib/lambda_ethereum_consensus/fork_choice/store.ex
mpaulucci: What would be the benefit? BTW, the Store as defined in the spec, stores blocks, not signed blocks: https://eth2book.info/capella/annotated-spec/#store

MegaRedHand: That's interesting. I was thinking we might want to validate block signatures in the fork choice.

mpaulucci: We do, we get `signed_block` in the `on_block` handler, then we do the validation and we store just the block

PR #367, opened 2023-10-31
feat: functions for struct comparisons

### LINKED ISSUE #365: [Debug tools] Create state_diff function
Currently, when an assertion over a post_state fails in an ExUnit test, the whole state is printed out, which can be very hard to read. At times, this causes the whole terminal to overflow and the actual results to get lost, depending on which emulator the dev is using.

We would highly benefit from having a `state_diff(state_1, state_2)` that returned the differences between them, and a way to print them properly. Initially, a quick implementation would just print the fields that are different and in what manner. Ideally, we would benefit from recursive diffs (e.g. if there's a list, specify which member is the different one, and how that member is different from the one in the other state).

Arkenan:
Closes #365

### DIFF
--- a/lib/utils/diff.ex
+++ b/lib/utils/diff.ex
@@ -1,10 +1,38 @@
 defmodule LambdaEthereumConsensus.Utils.Diff do
   @moduledoc """
-  A module to compare between different structures and get their differences. Useful for debugging.
+  A module to compare between different structures and get their differences.
+  Useful for debugging.
+
+  Most typical cases are maps and lists, which recurse in their elements.
+  Any other type is just compared with ==. Tuples may be added if needed,
+  but it wasn't the case yet.
+
+  Possible results:
+  - When comparing any two elements that are equal, the result is :unchanged.
+  - When comparing different maps or lists, the result will be a map containing the keys
+    :added_right, :added_left and :changed.
+
+  Diff attributes:
+  - :added_right or :added_left are lists of elements. When comparing lists, this will
+    just be the elements that were added. In the case of maps, they are keyword lists
+    showing the key and the value that were added.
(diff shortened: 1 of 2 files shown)


lib/utils/diff.ex
```diff
@@ -0,0 +1,94 @@
...
+  - When comparing any two elements that are equal, the result is :unchanged.
+  - When comparing different maps or lists, the result will be a map containing the keys
+    :added_right, :added_left and :changed.
+
+  Diff attributes:
+  - :added_right or :added_left are lists of elements. When comparing lists, this will
+    just be the elements that were added. In the case of maps, they are keyword lists
+    showing the key and the value that were added.
+  - :changed is always a keyword list. In the case of lists, keys are indices of the
+    changed elements and values are the elements itselfs. In the case of maps, the keys
```
MegaRedHand, line 20: typo
```suggestion
    changed elements and values are the elements themselves. In the case of maps, the keys
```

lib/utils/diff.ex
```diff
@@ -0,0 +1,94 @@
...
+    are the keys of the map.
+  """
+  @type structured_diff :: %{
+          optional(:added_left) => [any()],
+          optional(:added_right) => [any()],
+          optional(:changed) => Keyword.t()
+        }
+  @type base_diff :: %{optional(:left) => any(), optional(:right) => any()}
+  @type t :: :unchanged | base_diff() | base_diff()
```
MegaRedHand, line 29: The second `base_diff` should be `structured_diff`, right?

MegaRedHand (approved):

lib/utils/diff.ex
Arkenan: indeed

PR #368, opened 2023-10-31
feat: prune peers

MegaRedHand:
This PR:
- ~moves calls to `Process.send_after` to the app's `start` method, so the startup process is shorter~ (removed)
- adds peer removal when block requests fail
- adds continuous peer pruning by requesting metadata from peers, and removing any that fail
- increases the gossipsub module concurrency setting
- adds metrics for each call to Libp2pPort functions
- adds grafana vis for previous point, and for peer pruning


lib/lambda_ethereum_consensus/application.ex
```diff
@@ -41,7 +41,16 @@ defmodule LambdaEthereumConsensus.Application do
     # for other strategies and supported options
     opts = [strategy: :one_for_one, name: LambdaEthereumConsensus.Supervisor]
 
-    Supervisor.start_link(children, opts)
+    result = Supervisor.start_link(children, opts)
+
+    # Start scheduled tasks only when app started, to avoid
```
mpaulucci, line 46: does this make any difference?

lib/lambda_ethereum_consensus/beacon/sync_blocks.ex
```diff
@@ -12,7 +12,7 @@ defmodule LambdaEthereumConsensus.Beacon.SyncBlocks do
   alias LambdaEthereumConsensus.P2P.BlockDownloader
   alias LambdaEthereumConsensus.StateTransition.Misc
 
-  @blocks_per_chunk 20
+  @blocks_per_chunk 16
```
mpaulucci, line 15: why this change?

lib/lambda_ethereum_consensus/p2p/peerbook.ex
```diff
@@ -35,10 +45,54 @@ defmodule LambdaEthereumConsensus.P2P.Peerbook do
...
+    |> Map.keys()
+    |> Stream.drop(n)
+    |> Stream.take(prune_size)
+    |> Enum.each(fn peer_id -> Task.start(__MODULE__, :challenge_peer, [peer_id]) end)
+
+    schedule_pruning()
+    {:noreply, peerbook}
+  end
+
+  def challenge_peer(peer_id) do
```
mpaulucci, line 84: Is this logic taken from another client or was this your idea?

lib/lambda_ethereum_consensus/beacon/sync_blocks.ex
MegaRedHand: It just bothers me that it isn't a power of 2 😅

lib/lambda_ethereum_consensus/p2p/peerbook.ex
MegaRedHand: It was my idea, although other clients probably do the same. Requesting metadata and status data from peers is the best way to filter them based on syncing status.

lib/lambda_ethereum_consensus/application.ex
MegaRedHand: Yes. Although, with recent changes it might not make much sense.

mpaulucci (approved):

PR #372, opened 2023-11-02
refactor: simplify "epoch_processing" handlers

MegaRedHand:
This PR removes duplicated code in `epoch_processing` handlers, and changes assertions to use the new `Diff.diff/2` function


Arkenan (approved):
LGTM

PR #380, opened 2023-11-02
build: fix formatter taking forever.

mpaulucci:
**Motivation**
In VsCode, this line is making the formatting task extremely slow (10+ seconds). Not sure what the reason is but removing it fixes the issue. Also, I don't know what the purpose of this line is.

### DIFF
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -2,7 +2,6 @@
 [
   inputs:
     ["{mix,.formatter}.exs", "{config,lib,test}/**/*.{ex,exs}"]
-    |> Enum.flat_map(&Path.wildcard(&1, match_dot: true))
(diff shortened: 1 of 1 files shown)


.formatter.exs
```diff
@@ -2,7 +2,6 @@
 [
   inputs:
     ["{mix,.formatter}.exs", "{config,lib,test}/**/*.{ex,exs}"]
-    |> Enum.flat_map(&Path.wildcard(&1, match_dot: true))
```
MegaRedHand, line 5: What this does is expand the `["{mix,.formatter}.exs", "{config,lib,test}/**/*.{ex,exs}"]` patterns, so we can specify files to be formatted one by one, and reject those under `lib/proto` and `test/generated`.

MegaRedHand (approved):

PR #381, opened 2023-11-02
feat: add `on_tick` handler

### LINKED ISSUE #375: Implement `on_tick` handler
According to the spec:

```python
def on_tick(store: Store, time: uint64) -> None:
    # If the ``store.time`` falls behind, while loop catches up slot by slot
    # to ensure that every previous slot is processed with ``on_tick_per_slot``
    tick_slot = (time - store.genesis_time) // SECONDS_PER_SLOT
    while get_current_slot(store) < tick_slot:
        previous_time = store.genesis_time + (get_current_slot(store) + 1) * SECONDS_PER_SLOT
        on_tick_per_slot(store, previous_time)
    on_tick_per_slot(store, time)
```

MegaRedHand:
Closes #375

### DIFF
--- a/lib/lambda_ethereum_consensus/fork_choice/handlers.ex
+++ b/lib/lambda_ethereum_consensus/fork_choice/handlers.ex
@@ -7,13 +7,19 @@ defmodule LambdaEthereumConsensus.ForkChoice.Handlers do
 
   ### Public API ###
 
+  @doc """
+  Called once every tick (1 second). This function updates the Store's time.
+  Also updates checkpoints and resets proposer boost at the beginning of every slot.
+  """
   def on_tick(%Store{} = store, time) do
+    # If the ``store.time`` falls behind, while loop catches up slot by slot
+    # to ensure that every previous slot is processed with ``on_tick_per_slot``
     seconds_per_slot = ChainSpec.get("SECONDS_PER_SLOT")
     tick_slot = div(time - store.genesis_time, seconds_per_slot)
     current_slot = Store.get_current_slot(store)
 
-    current_slot..(tick_slot - 1)
-    |> Enum.reduce_while(store, fn current_slot, store ->
+    current_slot..(tick_slot - 1)//1
+    |> Enum.reduce(store, fn current_slot, store ->
       # TODO: simplify this by using time instead of slots
       previous_time = store.genesis_time + (current_slot + 1) * seconds_per_slot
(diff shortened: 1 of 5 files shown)


lib/lambda_ethereum_consensus/fork_choice/handlers.ex
```diff
@@ -3,11 +3,80 @@ defmodule LambdaEthereumConsensus.ForkChoice.Handlers do
   Handlers that update the fork choice store.
```
mpaulucci, line 3: nit: would it make sense to move to handlers/on_tick.ex ?

lib/lambda_ethereum_consensus/fork_choice/store.ex
```diff
@@ -65,6 +65,7 @@ defmodule LambdaEthereumConsensus.ForkChoice.Store do
     # TODO: this should be done after validation
     :ok = StateStore.store_state(anchor_state)
     :ok = BlockStore.store_block(signed_anchor_block)
+    schedule_next_tick()
```
mpaulucci, line 68: should we maybe ensure that the time is correctly set here before starting the Genser? Otherwise we will have a full second with a wrong time and there could be some undesired side effects.

MegaRedHand: The time is set according to the retrieved finalized state inside `get_forkchoice_store`

lib/lambda_ethereum_consensus/fork_choice/handlers.ex
MegaRedHand: There are not that many functions, and some are shared with `on_block`, so we would need a `handlers/helpers.ex` too.

lib/lambda_ethereum_consensus/fork_choice/store.ex
mpaulucci: Correct, but it is the time of the slot in the anchor_block. In checkpoint sync, it might not refer to the current time, but several minutes before.

mpaulucci (approved):

mpaulucci:
Merging this since it's blocking some of my work.

lib/lambda_ethereum_consensus/fork_choice/store.ex
MegaRedHand: Added the change in f25c895

PR #384, opened 2023-11-03
feat: add sync progress graph

MegaRedHand:
~Depends on #381~

Image of the new vis:

image showing new visualization in graphana


Arkenan (approved):
This is perfectly fine, although I'd add a card with the current number aside from the line plot. Knowing the exact number is always pretty good vs a rough approximation in the millions.

PR #386, opened 2023-11-05
feat: process proposer slashing

### LINKED ISSUE #304: [operations] Implement `process_proposer_slashing`
https://github.com/lambdaclass/lambda_ethereum_consensus/blob/dd0a0523e6be2a21fa3f6a96d4466926307d236c/docs/specs/beacon-chain.md?plain=1#L2684-L2705

Acceptance criteria:

karasakalmt:
Closes #304


lib/lambda_ethereum_consensus/state_transition/operations.ex
```diff
@@ -174,6 +174,74 @@ defmodule LambdaEthereumConsensus.StateTransition.Operations do
...
+      not (header_1 != header_2) ->
+        {:error, "Headers are same"}
+
+      not Predicates.is_slashable_validator(proposer, Accessors.get_current_epoch(state)) ->
+        {:error, "Proposer is not slashable"}
+
+      true ->
+        is_verified =
+          [proposer_slashing.signed_header_1, proposer_slashing.signed_header_2]
+          |> Enum.reduce_while(false, fn signed_header, _acc ->
```
MegaRedHand, line 200: I suggest using [`all?`](https://hexdocs.pm/elixir/Enum.html#all?/2) for this, like so:

```elixir
|> Enum.all?(fn signed_header -> ... end)
```

lib/lambda_ethereum_consensus/state_transition/operations.ex
```diff
@@ -174,6 +174,74 @@ defmodule LambdaEthereumConsensus.StateTransition.Operations do
...
+          end)
+
+        if not is_verified do
+          {:error, "Signed header 1 or 2 signature is not verified"}
+        end
+
+        case Mutators.slash_validator(state, header_1.proposer_index) do
+          {:ok, state} -> {:ok, state}
+          {:error, msg} -> {:error, msg}
+        end
```
MegaRedHand, line 225: The `case` isn't needed (since we just return the result in all cases), and it should go inside an `else` to make it work.

lib/spec/runners/operations.ex
```diff
@@ -103,6 +103,18 @@ defmodule OperationsTestRunner do
     end
   end
 
+  defp handle_case("proposer_slashing", pre, proposer_slashing, post, _case_dir) do
```
MegaRedHand, line 106: Remember to comment the name of the handler in `@disabled_handlers`

lib/lambda_ethereum_consensus/state_transition/operations.ex
```diff
@@ -174,6 +174,74 @@ defmodule LambdaEthereumConsensus.StateTransition.Operations do
...
+  @spec process_proposer_slashing(BeaconState.t(), SszTypes.ProposerSlashing.t()) ::
+          {:ok, BeaconState.t()} | {:error, String.t()}
+  def process_proposer_slashing(state, proposer_slashing) do
+    header_1 = proposer_slashing.signed_header_1.message
+    header_2 = proposer_slashing.signed_header_2.message
+    proposer = Enum.at(state.validators, header_1.proposer_index)
+
+    cond do
+      not (header_1.slot == header_2.slot) ->
+        {:error, "Slots doesn't match"}
```
MegaRedHand, line 186: Typo

```suggestion
        {:error, "Slots don't match"}
```

lib/lambda_ethereum_consensus/state_transition/operations.ex
```diff
@@ -174,6 +174,74 @@ defmodule LambdaEthereumConsensus.StateTransition.Operations do
...
+
+      not (header_1.proposer_index == header_2.proposer_index) ->
+        {:error, "Proposer indices doesn't match"}
```
MegaRedHand, line 189: Typo
```suggestion
        {:error, "Proposer indices don't match"}
```

lib/lambda_ethereum_consensus/state_transition/operations.ex
karasakalmt: I got the case part but was wondering in line 218 I am checking if the signature is verified and returning an error why doesn't it returns it as default and still executes 222

MegaRedHand: Because [there's no "return" in Elixir](https://stackoverflow.com/questions/41520031/why-there-is-no-return-statement-in-elixir)

karasakalmt: yeap it doesn't but like lets say
```elixir
fn foo(x) do
  if x do
    true
  end
  false
end
```

returns false if x is true as well why this is the case

MegaRedHand: Because, in Elixir, everything is an expression that evaluates to a value:
- `if` clauses evaluate to the last expression inside their `do` block's value if the condition is `true`, or the last expression inside their `else` block's otherwise (`nil` by default)
- function calls evaluate to the last expression in their `do` block's value
- literals evaluate to themselves

In this case, `foo(true)` evaluates to `false`, because

```elixir
if true do
  true
end
false
```
is equivalent to:
```elixir
true
false
```
Since the function returns only the last value, it returns `false`

karasakalmt: thanks ser!

lib/lambda_ethereum_consensus/state_transition/operations.ex
```diff
@@ -174,6 +174,77 @@ defmodule LambdaEthereumConsensus.StateTransition.Operations do
...
+  defp bls_verify_proposer_slashing(pubkey, signing_root, signature) do
+    verification = Bls.verify(pubkey, signing_root, signature)
+
+    case verification do
+      {:ok, bool} ->
+        if bool do
+          true
+        else
+          false
+        end
```
MegaRedHand, line 241: This can be simplified to just `bool`

lib/lambda_ethereum_consensus/state_transition/operations.ex
```diff
@@ -174,6 +174,77 @@ defmodule LambdaEthereumConsensus.StateTransition.Operations do
...
+
+            signing_root =
+              Misc.compute_signing_root(signed_header.message, domain)
+
+            bls_verify_proposer_slashing(
+              proposer.pubkey,
+              signing_root,
+              signed_header.signature
+            )
+          end)
```
MegaRedHand, line 222: This is a bit hard to read. Please move this to a named function, like so:

```elixir
|> Enum.all?(&verify_proposer_slashing(&1, state, proposer))

MegaRedHand (commented):
LGTM. I left some minor comments

MegaRedHand (approved):

MegaRedHand:
Good work! 🚀

PR #394, opened 2023-11-06
feat: add `on_attestation` handler

### LINKED ISSUE #376: Implement `on_attestation` handler
According to the spec:

```python
def on_attestation(store: Store, attestation: Attestation, is_from_block: bool=False) -> None:
    """
    Run ``on_attestation`` upon receiving a new ``attestation`` from either within a block or directly on the wire.

    An ``attestation`` that is asserted as invalid may be valid at a later time,
    consider scheduling it for later processing in such case.
    """
    validate_on_attestation(store, attestation, is_from_block)

    store_target_checkpoint_state(store, attestation.data.target)

    # Get state at the `target` to fully validate attestation
    target_state = store.checkpoint_states[attestation.data.target]
    indexed_attestation = get_indexed_attestation(target_state, attestation)
    assert is_valid_indexed_attestation(target_state, indexed_attestation)

    # Update latest messages for attesting indices
    update_latest_messages(store, indexed_attestation.attesting_indices, attestation)
```

MegaRedHand:
Closes #376

### DIFF
--- a/lib/lambda_ethereum_consensus/fork_choice/handlers.ex
+++ b/lib/lambda_ethereum_consensus/fork_choice/handlers.ex
@@ -86,13 +86,15 @@ defmodule LambdaEthereumConsensus.ForkChoice.Handlers do
 
       # Get state at the `target` to fully validate attestation
       target_state = store.checkpoint_states[attestation.data.target]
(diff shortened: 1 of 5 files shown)


mpaulucci (approved):
💯

PR #397, opened 2023-11-07
chore: use `hash_tree_root!` instead of `{:ok, _} = hash_tree_root`

MegaRedHand:
We are currently matching against the `:ok` result of `Ssz.hash_tree_root` because the error case shouldn't be handled: it can only happen due to values that don't respect the SSZ schema, which means the state is corrupt, and we _should_ crash. 

Changes:
- for ease of use, I'm introducing the `!`-version of the function
- refactoring existing code to use it
- along the way, I also merged the `Beacon.HelperFunctions` module into `StateTransition.Misc`


lib/ssz.ex
```diff
@@ -40,6 +40,12 @@ defmodule Ssz do
     end
   end
 
+  @spec hash_tree_root(struct) :: SszTypes.root()
```
mpaulucci, line 43:
```suggestion
  @spec hash_tree_root!(struct) :: SszTypes.root()
```

mpaulucci (approved):

PR #399, opened 2023-11-07
feat: [ssz]encode/decode basic types

### LINKED ISSUE #396: [SSZ] Encode/decode basic types
Refering to the [SSZ specs](https://github.com/ethereum/consensus-specs/blob/v1.3.0/ssz/simple-serialize.md), we need to implement, in a module called `SszEx`. This module will have two sections:

## Public schema-driven functions:

- `encode(value, {:int, size})`. `decode(binary, {:int, size})`
- `encode(value, :bool})`. `decode(binary, :bool)`

The public functions act as an API and only dispatch the call to the private ones. They don't do any other logic. We may change these with a single dictionary eventually.

## Private, schema-less functions:

- `encode_int(value, size) when is_integer(value)`.
- `encode_bool(false)`.
- `encode_bool(true)`.
- `decode_uint(binary, size)`.
- `decode_bool(binary)`.

These functions contain all ssz logic.

## Guidelines

- The module SszEx can't call any function from the previous Ssz library or the rust library.

Part of #240

f3r10:
**Motivation**

implement basic ssz basic types

Closes #396


Arkenan (changes requested):
Requested some changes. Let's also add ExUnit tests for them.

lib/ssz_ex.ex
```diff
@@ -0,0 +1,36 @@
...
+  #################
+  ### Private functions
+  #################
+  defp encode_int(value, size) when is_integer(value) do
+    <<encoded::binary-size(div(size, 8)), _rest::binary>> =
+      value
+      |> :binary.encode_unsigned(:little)
+      |> String.pad_trailing(div(size, 8), <<0>>)
+
+    encoded
```
Arkenan, line 23: I think this can be simplified to:

```elixir
defp encode_int(value, size) when is_integer(value), do: <<value::size(size)-little>>
```

Or am I missing something?

test/unit/ssz_ex_test.exs
```diff
@@ -0,0 +1,27 @@
...
+  test "serialize and deserialize uint" do
+    assert_roundtrip("05", 5, {:int, 8})
+    assert_roundtrip("0500", 5, {:int, 16})
+    assert_roundtrip("05000000", 5, {:int, 32})
+    assert_roundtrip("0500000000000000", 5, {:int, 64})
+
+    assert_roundtrip("63", 99, {:int, 8})
+    assert_roundtrip("6300", 99, {:int, 16})
+    assert_roundtrip("63000000", 99, {:int, 32})
+    assert_roundtrip("6300000000000000", 99, {:int, 64})
```
Arkenan, line 20: Instead of 99 let's use some value that requires more than one bit, like 276, which should be "1401".

test/unit/ssz_ex_test.exs
```diff
@@ -0,0 +1,27 @@
...
+  use ExUnit.Case
+
+  def assert_roundtrip(hex_serialized, deserialized, schema) do
+    serialized = Base.decode16!(hex_serialized)
+    assert ^serialized = SszEx.encode(deserialized, schema)
+    assert deserialized === SszEx.decode(serialized, schema)
+  end
+
+  test "serialize and deserialize uint" do
+    assert_roundtrip("05", 5, {:int, 8})
```
Arkenan, line 12: Instead of using hex, why don't we use the binaries directly? For example:

```elixir
<<5>>
<<5, 0>>
<<5, 0, 0, 0>>
<<5, 0, 0, 0, 0, 0, 0, 0>>
```

Or for a value like 276:

```elixir
<<14, 1>>
<<14, 1, 0, 0>>
<<14, 1, 0, 0, 0, 0, 0, 0>>
```

Arkenan (approved):
Thanks Fernando! Looks good to me. Added a few comments. Feel free to merge after fixing them.

lib/ssz_ex.ex
f3r10: it is necessary to add  ``<<0>>`` depending of the size. 
For example `5` with size u8 is ``<<5>>``, but if the size is u16 the encoding should be ``<<5, 0>>`` and so on.

f3r10: oh!! you are right, like magic

PR #402, opened 2023-11-07
feat: add optimistic block sync from checkpoint.

mpaulucci:
**Motivation**
When syncing from checkpoint, we are a couple of hundred blocks behind. Instead of relying on Pending Blocks to download parents one by one, let's optimistically download blocks from the checkpoint sync slot.

**Description**
Added `SyncBlocks` module a task that is triggered on application start that downloads all blocks from the last checkpoint sync to the current slot.

### DIFF
--- a/lib/lambda_ethereum_consensus/beacon/sync_blocks.ex
+++ b/lib/lambda_ethereum_consensus/beacon/sync_blocks.ex
@@ -21,9 +21,6 @@ defmodule LambdaEthereumConsensus.Beacon.SyncBlocks do
   end
 
   def run(_opts) do
-    # TODO: this is a hack to make sure the on_tick has been called at least once
-    Process.sleep(2000)
-
     {:ok, checkpoint} = Store.get_finalized_checkpoint()
 
     initial_slot = Misc.compute_start_slot_at_epoch(checkpoint.epoch)
--- a/lib/lambda_ethereum_consensus/beacon/sync_blocks.ex
+++ b/lib/lambda_ethereum_consensus/beacon/sync_blocks.ex
@@ -1,0 +1,121 @@
+defmodule LambdaEthereumConsensus.Beacon.SyncBlocks do
+  @moduledoc false
+
+  use GenServer
+
+  require Logger
+
+  alias LambdaEthereumConsensus.Beacon.PendingBlocks
+  alias LambdaEthereumConsensus.ForkChoice.Store
+  alias LambdaEthereumConsensus.P2P.BlockDownloader
+  alias LambdaEthereumConsensus.StateTransition.Misc
+
+  @blocks_per_chunk 20
+
+  @type state :: %{
+          chunks: [%{from: integer(), count: integer()}]
+        }
+
+  ##########################
+  ### Public API
+  ##########################
+
+  def start_link(opts) do
+    GenServer.start_link(__MODULE__, opts, name: __MODULE__)
+  end
+
+  ##########################
+  ### GenServer Callbacks
+  ##########################
+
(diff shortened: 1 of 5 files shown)


lib/lambda_ethereum_consensus/beacon/sync_blocks.ex
```diff
@@ -0,0 +1,98 @@
...
+
+  @type chunk :: %{from: SszTypes.slot(), count: integer()}
+
+  def start_link(opts) do
+    Task.start_link(__MODULE__, :run, [opts])
+  end
+
+  def run(_opts) do
+    # TODO: this is a hack to make sure the on_tick has been called at least once
+    Process.sleep(2000)
```
MegaRedHand, line 25: This shouldn't be needed since `on_tick` is called on `Store` initialization

lib/lambda_ethereum_consensus/beacon/sync_blocks.ex
```diff
@@ -0,0 +1,98 @@
...
+
+    if Enum.empty?(chunks) do
+      Logger.info("[Optimistic Sync] Sync completed")
+    else
+      Process.sleep(1000)
+      perform_sync(remaining_chunks)
+    end
+  end
+
+  @spec fetch_blocks_by_slot(non_neg_integer(), integer()) ::
```
MegaRedHand, line 81:
```suggestion
  @spec fetch_blocks_by_slot(SszTypes.slot(), non_neg_integer()) ::
```

lib/lambda_ethereum_consensus/beacon/sync_blocks.ex
```diff
@@ -0,0 +1,98 @@
...
+          {:ok, [SszTypes.SignedBeaconBlock]} | {:error, binary()}
```
MegaRedHand, line 82:
```suggestion
          {:ok, [SszTypes.SignedBeaconBlock.t()]} | {:error, String.t()}
```

lib/lambda_ethereum_consensus/p2p/block_downloader.ex
```diff
@@ -126,22 +125,18 @@ defmodule LambdaEthereumConsensus.P2P.BlockDownloader do
...
 
       <<0, ^fork_context::binary-size(4)>> <> rest ->
-        result = rest |> :binary.split(<<0, fork_context::binary-size(4)>>, [:global])
-
-        case result do
-          chunks when length(chunks) == count -> {:ok, chunks}
-          _ -> {:error, "expected #{count} chunks, got #{length(result)}"}
-        end
+        chunks = rest |> :binary.split(<<0, fork_context::binary-size(4)>>, [:global])
+        {:ok, chunks}
```
MegaRedHand, line 139: Do we still check the length somewhere else?

lib/lambda_ethereum_consensus/beacon/sync_blocks.ex
mpaulucci: awesome!

lib/lambda_ethereum_consensus/p2p/block_downloader.ex
mpaulucci: no, it doesn't really make sense in the context of blocks by range since you could have an empty slot. It might make more sense in the context of blocks by root

MegaRedHand: I thought it would just return an error code in the corresponding response chunk... TIL

MegaRedHand (approved):

PR #404, opened 2023-11-08
feat: add `on_attester_slashing` handler

### LINKED ISSUE #378: Implement `on_attester_slashing` handler
According to the spec:

```python
def on_attester_slashing(store: Store, attester_slashing: AttesterSlashing) -> None:
    """
    Run ``on_attester_slashing`` immediately upon receiving a new ``AttesterSlashing``
    from either within a block or directly on the wire.
    """
    attestation_1 = attester_slashing.attestation_1
    attestation_2 = attester_slashing.attestation_2
    assert is_slashable_attestation_data(attestation_1.data, attestation_2.data)
    state = store.block_states[store.justified_checkpoint.root]
    assert is_valid_indexed_attestation(state, attestation_1)
    assert is_valid_indexed_attestation(state, attestation_2)

    indices = set(attestation_1.attesting_indices).intersection(attestation_2.attesting_indices)
    for index in indices:
        store.equivocating_indices.add(index)
```

MegaRedHand:
Closes #378
Depends on #394

### DIFF
--- a/lib/lambda_ethereum_consensus/fork_choice/store.ex
+++ b/lib/lambda_ethereum_consensus/fork_choice/store.ex
@@ -51,6 +51,11 @@ defmodule LambdaEthereumConsensus.ForkChoice.Store do
     GenServer.cast(__MODULE__, {:on_attestation, attestation})
   end
 
+  @spec notify_attester_slashing(SszTypes.AttesterSlashing.t()) :: :ok
+  def notify_attester_slashing(attester_slashing) do
+    GenServer.cast(__MODULE__, {:attester_slashing, attester_slashing})
+  end
+
   ##########################
   ### GenServer Callbacks
   ##########################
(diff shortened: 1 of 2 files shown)


lib/lambda_ethereum_consensus/fork_choice/store.ex
```diff
@@ -91,14 +96,17 @@ defmodule LambdaEthereumConsensus.ForkChoice.Store do
...
-             end) do
+             |> apply_handler(new_state, &Handlers.on_attestation(&1, &2, true)),
+           {:ok, new_state} <-
+             signed_block.message.body.attester_slashings
+             |> apply_handler(new_state, &Handlers.on_attester_slashing/2) do
         new_state
       else
         _ ->
-          Logger.error("Failed to add block #{signed_block.message.slot} to the store.")
+          Logger.error(
```
mpaulucci, line 106: should we add the error reason to the log?

MegaRedHand: Done

mpaulucci (approved):

PR #414, opened 2023-11-10
feat: [ssz] add list encoding/decoding

### LINKED ISSUE #398: [SSZ] list encoding/decoding
## Public function (dispatchers)

- `encode(list, {:list, basic_type, :size})`. 
- `decode(binary, {:list, basic_type, :size})`.

## Private function (actual work)

A helper that takes a schema and returns true if it's a type of variable size.
- `variable_size?({:list, _, _})`. True.
- `variable_size?(:bool)`. False.
- `variable_size?({:uint, _, _})`. False.

And the dispatched SSZ functions:
- `encode_list(list, basic_type, size) when is_list(list)`. 
- `decode_list(binary, basic_type` 

## Guidelines

- Serializing lists should only call the general `encode(value, schema)` function to encode items. The schema is the base_type.
- Acceptance criteria: we need to be able to serialize and deserialize lists of uints and bools.

Part of #240

f3r10:
**Motivation**

Add support for SSZ list encoding/decoding

Closes #398


Arkenan (dismissed):
Hi Fernando! This is an incomplete review, but we can start with this changes first!

test/unit/ssz_ex_test.exs
```diff
@@ -22,4 +22,27 @@ defmodule Unit.SSZExTest do
     assert_roundtrip(<<1>>, true, :bool)
     assert_roundtrip(<<0>>, false, :bool)
   end
+
+  test "serialize and deserialize list" do
+    assert_roundtrip(<<1, 1, 0>>, [true, true, false], {:list, :bool, 3})
```
MegaRedHand, line 27: We should add a comment here clarifying that this schema shouldn't be used in practice. I was thinking this encoding was weirdly inefficient, but that's because this is a _boolean list_ instead of a _bitlist_.

test/unit/ssz_ex_test.exs
```diff
@@ -22,4 +22,27 @@ defmodule Unit.SSZExTest do
...
+
+    assert_roundtrip(
+      <<230, 0, 0, 0, 124, 1, 0, 0, 11, 2, 0, 0>>,
+      [230, 380, 523],
+      {:list, {:int, 32}, 3}
+    )
+
+    variable_list = [[5, 6], [7, 8], [9, 10]]
+    encoded_variable_list = <<12, 0, 0, 0, 14, 0, 0, 0, 16, 0, 0, 0, 5, 6, 7, 8, 9, 10>>
+    assert_roundtrip(encoded_variable_list, variable_list, {:list, {:list, {:int, 8}, 2}, 3})
```
MegaRedHand, line 37: We should add some tests with a list length that differs from the schema's max length.

MegaRedHand (commented):
Really good! Keep up the good work! 🔥 

This tool is quite useful for visualizing SSZ encoding: https://www.ssz.dev/visualizer

test/unit/ssz_ex_test.exs
```diff
@@ -22,4 +27,35 @@ defmodule Unit.SSZExTest do
...
+        0, 0, 0>>
+
+    assert_roundtrip(encoded_variable_list, variable_list, {:list, {:list, {:int, 64}, 2}, 3})
+
+    error_assert_roundtrip(
+      <<230, 0, 0, 0, 124, 1, 0, 0, 11, 2, 0, 0, 127, 91, 0, 0>>,
+      [230, 380, 523, 23_423],
+      {:list, {:int, 32}, 3},
+      "invalid max_size of list"
+    )
```
MegaRedHand, line 59: This is OK, but I was thinking of `:ok` cases:

```elixir
# length < max_size
assert_roundtrip(<<2, 0, 0, 0>>, [2], {:list, {:int, 32}, 53})
# empty list
assert_roundtrip(<<>>, [], {:list, {:int, 32}, 6})
```

f3r10: oh! ok :sweat_smile:

lib/ssz_ex.ex
```diff
@@ -8,22 +8,222 @@ defmodule LambdaEthereumConsensus.SszEx do
...
+      if i == num_items do
+        part = :binary.part(binary, offset, byte_size(binary) - offset)
+
+        with {:ok, decoded} <- decode(part, basic_type) do
+          {:cont, {:ok, {[decoded | acc_decoded], offset, rest_bytes}}}
+        end
+      else
+        get_next_offset(acc_decoded, acc_rest_bytes, basic_type, offset, binary, first_offset)
+      end
+    end)
```
MegaRedHand, line 159: I find this part difficult to understand. Also, the inner `with` has the same problem from `get_next_offset`: it can return an `{:error, error}` tuple that needs handling. I think we should take an approach similar to `decode_chunk` on this one.

lib/ssz_ex.ex
```diff
@@ -8,22 +8,222 @@ defmodule LambdaEthereumConsensus.SszEx do
...
+               |> Enum.reverse()
+               |> Enum.map(&encode(&1, {:int, 32}))
+               |> flatten_results() do
+          (encoded_variable_offsets ++ encoded_variable_parts)
+          |> :binary.list_to_bin()
+          |> then(&{:ok, &1})
+        end
+      else
+        {:error, "invalid lengths"}
+      end
```
MegaRedHand, line 80: Nit. We can extract the `if` into a helper function:

```elixir
defp check_length(fixed_lengths, total_byte_size)
  when fixed_lengths + total_byte_size < 2 ** (@bytes_per_length_offset * @bits_per_byte),
  do: :ok
# handle the other case
```

Then you can merge both `with`s easily:

```elixir
with {:ok, ...} <- encode...,
  :ok <- check_length(...),
  {variable_offsets, _} = ...,
  {:ok, ...} <- variable_offsets... do
    ...
end
```

lib/ssz_ex.ex
```diff
@@ -8,22 +8,222 @@ defmodule LambdaEthereumConsensus.SszEx do
...
+      with {:ok, first_offset} <-
+             sanitize_offset(first_offset, nil, byte_size(binary), first_offset) do
+        {:ok, {decoded_list, _, _}} =
+          decode_variable_list(first_offset, rest_bytes, num_items, binary, basic_type)
+
+        if length(decoded_list) > size do
+          {:error, "invalid length list"}
+        else
+          {:ok, decoded_list |> Enum.reverse()}
+        end
```
MegaRedHand, line 137: We could make this check after we get `num_items`, right?

MegaRedHand (approved):
LGTM. We can leave further changes for later refactors.

PR #415, opened 2023-11-10
fix: number of remaining blocks is rounded up

MegaRedHand:
This PR fixes the "Remaining blocks: ..." message by counting the number of blocks remaining instead of approximating it. It also lowers the log level of attestation notifications to debug, since it's too spammy.

### DIFF
--- a/lib/lambda_ethereum_consensus/beacon/sync_blocks.ex
+++ b/lib/lambda_ethereum_consensus/beacon/sync_blocks.ex
@@ -40,7 +40,8 @@ defmodule LambdaEthereumConsensus.Beacon.SyncBlocks do
 
   @spec perform_sync([chunk()]) :: :ok
   def perform_sync(chunks) do
-    Logger.info("[Optimistic Sync] Blocks remaining: #{Enum.count(chunks) * @blocks_per_chunk}")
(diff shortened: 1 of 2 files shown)


lib/lambda_ethereum_consensus/beacon/sync_blocks.ex
```diff
@@ -40,7 +40,8 @@ defmodule LambdaEthereumConsensus.Beacon.SyncBlocks do
 
   @spec perform_sync([chunk()]) :: :ok
   def perform_sync(chunks) do
-    Logger.info("[Optimistic Sync] Blocks remaining: #{Enum.count(chunks) * @blocks_per_chunk}")
+    remaining = chunks |> Stream.map(fn %{count: c} -> c end) |> Enum.sum()
```
mpaulucci, line 43: I don't see when this would make any difference?

MegaRedHand: The last chunk has a different number of blocks, so approximating it by `Enum.count(chunks) * @blocks_per_chunk` rounds it up to the next multiple of 20.

mpaulucci (approved):

PR #417, opened 2023-11-13
feat: process_deposit

### LINKED ISSUE #303: [operations] Implement `process_deposit`
https://github.com/lambdaclass/lambda_ethereum_consensus/blob/dd0a0523e6be2a21fa3f6a96d4466926307d236c/docs/specs/beacon-chain.md?plain=1#L2874-L2895

Acceptance criteria:

karasakalmt:
Closes #303


lib/lambda_ethereum_consensus/state_transition/operations.ex
```diff
@@ -375,6 +375,54 @@ defmodule LambdaEthereumConsensus.StateTransition.Operations do
...
   end
 
+  @spec process_deposit(BeaconState.t(), SszTypes.Deposit.t()) ::
+          {:ok, BeaconState.t()} | {:error, String.t()}
+  def process_deposit(state, deposit) do
+    deposit_data_root = Ssz.hash_tree_root(deposit.data)
+
+    case deposit_data_root do
+      {:ok, deposit_data_root} ->
+        is_valid =
```
karasakalmt, line 385: This matches with a wrong value but the other parts work well

lib/lambda_ethereum_consensus/state_transition/predicates.ex
```diff
@@ -102,6 +102,35 @@ defmodule LambdaEthereumConsensus.StateTransition.Predicates do
     is_indices_available(validators, indices, h < validators)
   end
 
+  @doc """
+  Check if merkle branch is valid
+  """
+  @spec is_valid_merkle_branch(
```
karasakalmt, line 108: This is not work as intended but this is same with the implementation

lib/lambda_ethereum_consensus/state_transition/predicates.ex
```diff
@@ -102,6 +102,35 @@ defmodule LambdaEthereumConsensus.StateTransition.Predicates do
...
+    else
+      if rem(div(index, 2 ** i), 2) == 1 do
+        value = :crypto.hash(:sha256, Enum.at(branch, i) <> value)
+        is_valid_merkle_branch(leaf, branch, depth, index, root, value, i + 1)
+      else
+        value = :crypto.hash(:sha256, value <> Enum.at(branch, i))
+        is_valid_merkle_branch(leaf, branch, depth, index, root, value, i + 1)
+      end
+    end
+  end
```
Arkenan, line 132: Let's change this implementation to the following structure:

```elixir
def is_valid_merkle_branch?(leaf, branch, depth, index, root) do
  root == branch
    |> Enum.take(depth)
    |> Enum.with_index()
    |> Enum.reduce(leaf, fn {i, v}, value -> hash_merkle_node(v, value, index, i) end)
end

defp hash_merkle_node(value_1, value_2, index, i) when rem(div(index, 2 ** i), 2) == 1 do
  :crypto.hash(:sha256, value_1 <> value_2)
end

defp hash_merkle_node(value_1, value_2, _index, _i) do
  :crypto.hash(:sha256, value_2 <> value_1)
end
```

Aside from readability, this prevents us from using `Enum.at()` in each iteration, so we go from quadratic to linear complexity.

lib/lambda_ethereum_consensus/state_transition/operations.ex
```diff
@@ -375,6 +375,54 @@ defmodule LambdaEthereumConsensus.StateTransition.Operations do
     end
   end
 
+  @spec process_deposit(BeaconState.t(), SszTypes.Deposit.t()) ::
+          {:ok, BeaconState.t()} | {:error, String.t()}
+  def process_deposit(state, deposit) do
+    deposit_data_root = Ssz.hash_tree_root(deposit.data)
+
+    case deposit_data_root do
```
Arkenan, line 383: Let's simplify this to:

```elixir

with {:ok, deposit_data_root} <- Ssz.hash_tree_root(deposit.data) do
  if Predicates.is_valid_merkle_branch() do
    apply_deposit(state, deposit)
  else
    {:error, "Merkle branch is not valid"}
  end 
end

File: lib/lambda_ethereum_consensus/state_transition/operations.ex [hunk 5]
```diff
@@ -375,6 +375,54 @@ defmodule LambdaEthereumConsensus.StateTransition.Operations do
...
+        deposit.data.pubkey,
+        deposit.data.withdrawal_credentials,
+        deposit.data.amount,
+        deposit.data.signature
+      )
+
+    case res do
+      {:ok, state} -> {:ok, state}
+      {:error, msg} -> {:error, msg}
+    end
```
Arkenan, line 423: This case does nothing. It just returns what it received. We should:

```elixir
state
|> Map.put(:eth1_deposit_index, state.eth1_deposit_index + 1)
|> Mutators.apply_deposit(deposit.data.pubkey, ...)
```

With a bigger refactor in mind we probably wouldn't need this function at all.

lib/lambda_ethereum_consensus/state_transition/mutators.ex
```diff
@@ -160,4 +160,61 @@ defmodule LambdaEthereumConsensus.StateTransition.Mutators do
...
+          BeaconState.t(),
+          SszTypes.bls_pubkey(),
+          SszTypes.bytes32(),
+          SszTypes.uint64(),
+          SszTypes.bls_signature()
+        ) :: {:ok, BeaconState.t()} | {:error, binary()}
+  def apply_deposit(state, pubkey, withdrawal_credentials, amount, signature) do
+    validator_pubkeys = Enum.map(state.validators, fn validator -> validator.pubkey end)
+
+    if Enum.member?(validator_pubkeys, pubkey) do
```
Arkenan, line 174: Let's turn these two and the index finding into one:

```elixir
case Enum.find_index(state.validators, fn validator -> validator.pubkey == pubkey end) do
  index when is_number(index) ->
    # We increase the balance for that index.
  _ -> 
    # We add the validator.
end
```

lib/lambda_ethereum_consensus/state_transition/mutators.ex
```diff
@@ -160,4 +160,61 @@ defmodule LambdaEthereumConsensus.StateTransition.Mutators do
...
+      index = Enum.find_index(validator_pubkeys, fn validator -> validator == pubkey end)
+      state = increase_balance(state, index, amount)
+      {:ok, state}
```
Arkenan, line 177: This can be `{:ok, increase_balance(state, index, amount)}`

lib/lambda_ethereum_consensus/state_transition/mutators.ex
```diff
@@ -160,4 +160,61 @@ defmodule LambdaEthereumConsensus.StateTransition.Mutators do
...
+
+      signing_root = Misc.compute_signing_root(deposit_message, domain)
+
+      case Bls.verify(pubkey, signing_root, signature) do
+        {:ok, verified} ->
+          apply_initial_deposit(verified, state, pubkey, withdrawal_credentials, amount)
+
+        {:error, msg} ->
+          {:error, msg}
+      end
```
Arkenan, line 195: Whenever we have a case and we want to ignore the default case, let's use `with` instead:

```elixir
with {:ok, verified} <- Bls.verify(pubkey, signing_root, signature) do
  apply_initial_deposit(verified, state, pubkey, withdrawal_credentials, amount)
end
```

lib/lambda_ethereum_consensus/state_transition/mutators.ex
```diff
@@ -160,4 +160,61 @@ defmodule LambdaEthereumConsensus.StateTransition.Mutators do
...
+    end
+  end
+
+  defp apply_initial_deposit(verified, state, pubkey, withdrawal_credentials, amount) do
+    if verified do
```
Arkenan, line 200: Why do we need to check if verified is a truthy value? If that's the case we can include it in the `with` as:

```elixir
with {:ok, true} <- Bls.verify... do
  %Beaconstate{...}
else
  {:ok, _} -> {:error, "BLS verification failed"}
  other -> other
end
```

Arkenan (changes requested):
Hi, I wasn't able to find the bug, but I requested some code improvements all throughout.

lib/lambda_ethereum_consensus/state_transition/mutators.ex
karasakalmt: does this also send the `{:error, msg}` in case it doesn't match in the examples i found that was like an else and sending `err -> err`

lib/lambda_ethereum_consensus/state_transition/predicates.ex
karasakalmt: There is an issue about using rem and div in when guard;

error: cannot invoke remote function Kernel.**/2 inside guards

lib/lambda_ethereum_consensus/state_transition/mutators.ex
karasakalmt: I am checking if verified is equal to true actually

lib/lambda_ethereum_consensus/state_transition/mutators.ex
```diff
@@ -160,4 +160,60 @@ defmodule LambdaEthereumConsensus.StateTransition.Mutators do
...
+          amount: amount
+        }
+
+        domain = Misc.compute_domain(Constants.domain_deposit())
+
+        signing_root = Misc.compute_signing_root(deposit_message, domain)
+
+        with {:ok, verified} <- Bls.verify(pubkey, signing_root, signature) do
+          apply_initial_deposit(verified, state, pubkey, withdrawal_credentials, amount)
+        end
```
MegaRedHand, line 192: This should use `Bls.valid?` and mustn't return an error if verification fails.

lib/lambda_ethereum_consensus/state_transition/mutators.ex
```diff
@@ -160,4 +160,58 @@ defmodule LambdaEthereumConsensus.StateTransition.Mutators do
...
+          SszTypes.bytes32(),
+          SszTypes.uint64(),
+          SszTypes.bls_signature()
+        ) :: {:ok, BeaconState.t()} | {:error, binary()}
+  def apply_deposit(state, pubkey, withdrawal_credentials, amount, signature) do
+    validator_pubkeys = Enum.map(state.validators, fn validator -> validator.pubkey end)
+
+    case Enum.find_index(state.validators, fn validator -> validator.pubkey == pubkey end) do
+      index when is_number(index) ->
+        index = Enum.find_index(validator_pubkeys, fn validator -> validator == pubkey end)
```
MegaRedHand, line 176: I think this line isn't needed

lib/constants.ex
```diff
@@ -21,6 +21,9 @@ defmodule Constants do
   @spec domain_beacon_proposer() :: SszTypes.domain_type()
   def domain_beacon_proposer, do: <<0, 0, 0, 0>>
 
+  @spec domain_deposit() :: <<_::32>>
```
MegaRedHand, line 24:
```suggestion
  @spec domain_deposit() :: SszTypes.domain_type()
```

MegaRedHand (commented):
Good work! These are my last comments. Once they're addressed, it should be good to merge

MegaRedHand (approved):

lib/lambda_ethereum_consensus/state_transition/predicates.ex
```diff
@@ -102,6 +102,34 @@ defmodule LambdaEthereumConsensus.StateTransition.Predicates do
...
```
MegaRedHand, line 131: Let's change this to:

```elixir
  defp is_valid_merkle_branch(_, _, depth, _, root, value, depth), do: value == root

defp is_valid_merkle_branch(leaf, branch, depth, index, root, value, i) do
    if rem(div(index, 2 ** i), 2) == 1 do
      value = :crypto.hash(:sha256, Enum.at(branch, i) <> value)
      is_valid_merkle_branch(leaf, branch, depth, index, root, value, i + 1)
    else
      value = :crypto.hash(:sha256, value <> Enum.at(branch, i))
      is_valid_merkle_branch(leaf, branch, depth, index, root, value, i + 1)
    end
  end
```

Then we can refactor the `rem(div(index, 2 ** i), 2) == 1` part to be able to use it in guards. For that, we should divide `index` by 2 at each step, to avoid having to call `**`.

karasakalmt: I tried it https://github.com/lambdaclass/lambda_ethereum_consensus/pull/417#discussion_r1394657081 in here but the thing is that it appears that is not allowed to use rem() or div() in a guard

lib/lambda_ethereum_consensus/state_transition/predicates.ex
```diff
@@ -102,6 +102,32 @@ defmodule LambdaEthereumConsensus.StateTransition.Predicates do
...
+          SszTypes.uint64(),
+          SszTypes.root()
+        ) :: boolean
+  def is_valid_merkle_branch(leaf, branch, depth, index, root) do
+    is_valid_merkle_branch(leaf, branch, depth, index, root, leaf, 0)
+  end
+
+  defp is_valid_merkle_branch(_, _, depth, _, root, value, depth), do: value == root
+
+  defp is_valid_merkle_branch(leaf, branch, depth, index, root, value, i) do
```
MegaRedHand, line 121: The first argument (`leaf`) isn't used.

lib/lambda_ethereum_consensus/state_transition/predicates.ex
MegaRedHand: The one that's not allowed on guards is the `**` operator. `div` and `rem` are.

lib/lambda_ethereum_consensus/state_transition/predicates.ex
Arkenan: I still think it's better to have this in a reduce, without the manual recursion. If we can't use the guards, we can just use an `if` for the `hash_merkle_node` function:

```elixir
defp hash_merkle_node(value_1, value_2, index, i) do
  if rem(div(index, 2 ** i), 2) == 1 do
    :crypto.hash(:sha256, value_1 <> value_2)
  else
    :crypto.hash(:sha256, value_2 <> value_1)
  end
end
```

lib/lambda_ethereum_consensus/state_transition/mutators.ex
Arkenan: Yes, it returns the error when it doesn't match. Doing `err -> err` is redundant UNLESS there there's a special error case being handled.

Arkenan (approved):
LGTM

PR #426, opened 2023-11-14
feat: store invalid blocks to make sure we don't download it indefinitely.

mpaulucci:
**Motivation**
We want to store a reference to invalid blocks so that we don't try to download them or descendants indefinitely.

### DIFF
--- a/lib/lambda_ethereum_consensus/beacon/pending_blocks.ex
+++ b/lib/lambda_ethereum_consensus/beacon/pending_blocks.ex
@@ -69,8 +69,6 @@ defmodule LambdaEthereumConsensus.Beacon.PendingBlocks do
   @impl true
   @spec handle_info(atom(), state()) :: {:noreply, state()}
   def handle_info(:process_blocks, state) do
-    schedule_blocks_processing()
-
     state.pending_blocks
     |> Enum.sort_by(fn {_, signed_block} -> signed_block.message.slot end)
     |> Enum.reduce(state, fn {block_root, signed_block}, state ->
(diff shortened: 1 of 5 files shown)


lib/lambda_ethereum_consensus/fork_choice/store.ex
```diff
@@ -88,30 +88,28 @@ defmodule LambdaEthereumConsensus.ForkChoice.Store do
...
-          Logger.error(
-            "[Fork choice] Failed to add block #{signed_block.message.slot} to the store: #{reason}"
-          )
-
-          state
-      end
-
-    {:noreply, state}
+    with {:ok, new_store} <- Handlers.on_block(state, signed_block) do
+      #  # process block attestations
```
MegaRedHand, line 95: We should add a TODO

lib/lambda_ethereum_consensus/beacon/pending_blocks.ex
```diff
@@ -12,7 +12,11 @@ defmodule LambdaEthereumConsensus.Beacon.PendingBlocks do
   alias LambdaEthereumConsensus.P2P.BlockDownloader
   alias SszTypes.SignedBeaconBlock
 
-  @type state :: %{pending_blocks: %{}}
+  @type state :: %{
+          pending_blocks: %{SszTypes.root() => SignedBeaconBlock.t()},
+          invalid_blocks: %{SszTypes.root() => SignedBeaconBlock.t()},
```
MegaRedHand, line 17:
```suggestion
          invalid_blocks: %{SszTypes.root() => map()},
```

MegaRedHand (approved):

PR #405, opened 2023-11-08
feat: implement process_bls_to_execution_change operation

h3lio5:
**Motivation**
Implements the process_bls_to_execution_change() operation.

Closes #308

PR #430, opened 2023-11-14
feat: implement process_bls_to_execution_change

h3lio5:
**Motivation**

Implements the process_bls_to_execution_change operation.

Closes #405


lib/constants.ex
```diff
@@ -6,6 +6,12 @@ defmodule Constants do
   @spec genesis_epoch() :: integer
   def genesis_epoch, do: 0
 
+  @spec bls_withdrawal_prefix() :: <<_::1>>
```
MegaRedHand, line 9: The size after the `::` is in bits. In this case, it should be `<<_::8>>`.

lib/constants.ex
```diff
@@ -6,6 +6,12 @@ defmodule Constants do
...
+  def bls_withdrawal_prefix, do: <<0>>
+
+  @spec eth1_address_withdrawal_prefix() :: <<_::1>>
```
MegaRedHand, line 12: Same

lib/lambda_ethereum_consensus/state_transition/operations.ex
```diff
@@ -713,4 +713,87 @@ defmodule LambdaEthereumConsensus.StateTransition.Operations do
...
+    if address_change.validator_index < length(state.validators) do
+      {:ok, address_change}
+    else
+      {:error, "Invalid address change"}
+    end
+  end
+
+  defp validate_withdrawal_credentials(validator, address_change) do
+    if binary_part(validator.withdrawal_credentials, 0, 1) == Constants.bls_withdrawal_prefix() and
+         binary_part(validator.withdrawal_credentials, 1, 31) ==
```
MegaRedHand, line 774: Note you can do the same with matches:

```elixir
<<prefix::binary-size(1), address::binary-size(31)>> = validator.withdrawal_credentials
<<_, hash::binary-size(31)>> = :crypto.hash(:sha256, address_change.from_bls_pubkey)
if prefix == Constants.bls_withdrawal_prefix() and address = hash
```

lib/lambda_ethereum_consensus/state_transition/operations.ex
```diff
@@ -713,4 +713,87 @@ defmodule LambdaEthereumConsensus.StateTransition.Operations do
...
+           binary_part(:crypto.hash(:sha256, address_change.from_bls_pubkey), 1, 31) do
+      {:ok, true}
```
MegaRedHand, line 776: We should just return `:ok` here

h3lio5: but isn't it a convention to use `{:ok, value}` , `{:error, err_msg}`? If it omit `value` in the `:ok` variant, should I drop the `error_msg` in the `:error` variant to keep consistent?

lib/lambda_ethereum_consensus/state_transition/operations.ex
h3lio5: sure, I changed it!

lib/lambda_ethereum_consensus/state_transition/operations.ex
MegaRedHand: `{:ok, value}` is a convention for returning a value along with the `:ok`. We don't need to return a value if it isn't needed.
Also, we can drop the message for the error variant, but doing so would mean all callers of the function have to handle the specific `:error` case, and that might be cumbersome.

h3lio5: ok I changed it.

lib/constants.ex
```diff
@@ -18,6 +24,9 @@ defmodule Constants do
   @spec domain_voluntary_exit() :: <<_::32>>
   def domain_voluntary_exit, do: <<4, 0, 0, 0>>
 
+  @spec domain_bls_to_execution_change() :: <<_::32>>
```
MegaRedHand, line 27: We should use `SszTypes.domain_type()` as the return type. Feel free to update the other `domain_*` functions

lib/spec/runners/operations.ex
```diff
@@ -98,7 +98,7 @@ defmodule OperationsTestRunner do
 
     case post do
       nil ->
-        assert {:error, _error_msg} = result
+        assert {:error, error_msg} = result
```
MegaRedHand, line 101: This should still be `_error_msg`

MegaRedHand (approved):
Good work! 🔥

PR #439, opened 2023-11-23
perf: ease the load to the fork choice GenServer.

mpaulucci:
**Motivation**
We're seeing that when there is a big load of messages to the `ForkChoice.Store` GenServer (typically a combination of ~500 attestations and ~100 blocks that are being synced) then some other call to the same server (for example, to get the current slot) ends up in 5000ms timeout.

**Description**
We need to think about how to we deal with this holistically. But in the meantime, I added a couple of hacks to ease the load.

Some questions:
- Do we really need to process 500 aggregated attestations per slot? (In one example for a single slot it meant 141416 attestations. That numbers times 32 is 4M, much more than the 0.8M of active validators on mainnet, if my math is right). 
- Should we create different GenServers to process attestations for example?

### DIFF
--- a/lib/lambda_ethereum_consensus/fork_choice/store.ex
+++ b/lib/lambda_ethereum_consensus/fork_choice/store.ex
@@ -13,6 +13,8 @@ defmodule LambdaEthereumConsensus.ForkChoice.Store do
   alias SszTypes.SignedBeaconBlock
   alias SszTypes.Store
 
+  @default_timeout 10_000
+
   ##########################
   ### Public API
   ##########################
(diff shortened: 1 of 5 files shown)


lib/lambda_ethereum_consensus/fork_choice/store.ex
```diff
@@ -36,14 +38,14 @@ defmodule LambdaEthereumConsensus.ForkChoice.Store do
 
   @spec has_block?(SszTypes.root()) :: boolean()
   def has_block?(block_root) do
-    [blocks] = get_store_attrs([:blocks])
-    Map.has_key?(blocks, block_root)
+    block = get_block(block_root)
+    block != nil
```
MegaRedHand, line 42: This could be optimized even further by making `get_block` return a boolean instead of the whole block.

lib/lambda_ethereum_consensus/fork_choice/store.ex
```diff
@@ -130,11 +137,11 @@ defmodule LambdaEthereumConsensus.ForkChoice.Store do
     id = attestation.signature |> Base.encode16() |> String.slice(0, 8)
     Logger.debug("[Fork choice] Adding attestation #{id} to the store.")
 
-    state =
-      case Handlers.on_attestation(state, attestation, false) do
-        {:ok, new_state} -> new_state
-        _ -> state
-      end
+    # state =
```
MegaRedHand, line 140: Why's this commented?

MegaRedHand (approved):

lib/lambda_ethereum_consensus/fork_choice/store.ex
mpaulucci: Good catch, I was just testing stuff

lib/lambda_ethereum_consensus/fork_choice/store.ex
mpaulucci: We could further optimize it, but im keeping it simple for now (we also want to be able to get a full block)

PR #442, opened 2023-11-23
fix: eternal `make lint` runs

MegaRedHand:
This PR fixes the issue with `make lint` taking forever to run. This was because the formatter included generated test files. The solution was doing a partial expansion of the test folder, and excluding the `test/generated` folder from the matched paths.

Also, I moved the protobuf generated files to the `proto` directory, so the formatter and linter don't include them.

### DIFF
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -1,4 +1,7 @@
 # Used by "mix format"
 [
-  inputs: ["{mix,.formatter}.exs", "{config,lib,test}/**/*.{ex,exs}"]
+  inputs:
+    (["{mix,.formatter}.exs", "{config,lib}/**/*.{ex,exs}"] ++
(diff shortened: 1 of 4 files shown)


.formatter.exs
```diff
@@ -1,4 +1,6 @@
 # Used by "mix format"
 [
-  inputs: ["{mix,.formatter}.exs", "{config,lib,test}/**/*.{ex,exs}"]
+  inputs:
+    ["{mix,.formatter}.exs", "{config,lib}/**/*.{ex,exs}"] ++
+      ((Path.wildcard("test/*") -- ["test/generated"]) |> Enum.map(&(&1 <> "/**/*.{ex,exs}")))
```
mpaulucci, line 5: this does not bring back the regression about local formatting taking forever, right?

mpaulucci (approved):

PR #452, opened 2023-11-23
fix: allow some block downloads to fail

mpaulucci:
**Motivation**
Before, if you were requesting several blocks and there was one failure, all the request would be considered failed. Instead of that, just return the blocks that have been succesfull


MegaRedHand (approved):
LGTM

PR #456, opened 2023-11-26
feat: add Open Api Spex

### LINKED ISSUE #165: Add OpenApiSpex to the phoenix endpoints
The [OpenApiSpex](https://github.com/open-api-spex/open_api_spex) projects allows for input validation using open api schemas that describe the fields and types we accept as part of a query. This issue will be considered finished once all implemented endpoints use Open API spex. After this issue is finished, all new endpoints should validate inputs with OpenApiSpex.

okekefrancis112:
**Motivation**

Add OpenApiSpex

**Description**

Add OpenApiSpex to restrict users to pass in the correct parameters while calling the function.

Here is the link to the issue:  https://github.com/lambdaclass/lambda_ethereum_consensus/issues/165

Closes #165


MegaRedHand (commented):
There're some unrelated changes in the readme, and some compilation artifacts (`libp2p_port/internal/proto/libp2p.pb.go` and `libp2p_port/libp2p_port`). You can fix those with the following commands:

```shell
# restores readme from main
git show main:README.md > README.md
# removes compilation artifacts
git rm libp2p_port/libp2p_port libp2p_port/internal/proto/libp2p.pb.go
```

Arkenan (changes requested):
Hi Francis, let's fix the CI issues before reviewing this. You have two separate problems:

1. Linter fails. Fix this with `mix format`
2. The project doesn't compile. You added the plug `OpenApiSpex.Plug.CastAndValidate` but the module is not available, most likely because you need to add `{:open_api_spex, "~> 3.18"}` to the dependencies in the `mix.exs` file.

Let's also remove the files from `libp2p_port` folder, as they are outdated (we don't use them anymore).

okekefrancis112:
Hi Tomas,

Please will it be possible for us to have a session together, so we can
take a look at this together, if you don't mind?
Let me know the date and time you are free.

PR #463, opened 2023-11-27
feat: implement ssz encode/decode container

### LINKED ISSUE #416: [SSZ] Encode/decode a simple container
## Goals

The goal of this issue is to encode Validator, which is an example of a container that has different types in it, but none of them is another container. The type is described in the specs as:

```python
class Validator(Container):
    pubkey: BLSPubkey
    withdrawal_credentials: Bytes32  # Commitment to pubkey for withdrawals
    effective_balance: Gwei  # Balance at stake
    slashed: boolean
    # Status epochs
    activation_eligibility_epoch: Epoch  # When criteria for activation were met
    activation_epoch: Epoch
    exit_epoch: Epoch
    withdrawable_epoch: Epoch  # When validator can withdraw funds
```

We need to:

- Add a `Container` behavior module which defines a `schema()` callback.
- Make the `Validator` module implement `Container` behavior, by implementing a `schema()` function.
- in the ssz library, we need an `encode(container, module) when is_map(container)` function that calls `encode(container, module.schema())`. In this way we won't need to add an encoding function for each module, we just need to define their schemas in the container module, and the module itself will act as a schema. 
- Same with the decoding function.

f3r10:
**Motivation**

Closes #416


lib/ssz_ex.ex
```diff
@@ -184,6 +192,185 @@ defmodule LambdaEthereumConsensus.SszEx do
...
 
+  defp encode_container(container, schemas) do
+    values =
+      schemas
+      |> Enum.map(fn {key, schema} ->
+        value = Map.get(container, key)
+        {value, schema}
+      end)
+
+    with {:ok,
```
Arkenan, line 203: This is a very smart implementation! reduce_while is perfectly applied. Nevertheless, it's a bit convoluted with respect to code readability. I propose a refactor. We should be able to separate this into three parts:

```elixir
{fixed_size_values, fixed_length, variable_values} = analyze_schemas(schemas)
with {:ok, variable_parts} <- encode_schemas(variable_values),
     offsets = calculate_offsets(variable_parts, fixed_length),
     variable_length = Enum.reduce(variable_parts, 0, fn part, acc -> length(part) + acc end),
     :ok <- check_length(fixed_length + variable_length)
     {:ok, fixed_parts} <- replace_offsets(fixed_size_values, offsets) |> encode_schemas() do
       fixed_parts ++ variable_parts
end
```

This helper function might help:

```elixir
defp encode_schemas(schemas) do
  Enum.map(variable_values, fn {value, schema} -> encode(value, schema) end)
  |> flatten_results() 
end
```

The `analyze_schemas` function should return:

- `fixed_size_values`, which is a list of {value, schema} tuples for fixed size elements interleaved with `:offset` values.
- `fixed_length`, which is the sum of the fixed elements and the offset lengths. We can use the function `get_fixed_size` internally.
- `variable_values`m which is a list of {value, schema} only for variable sized elements.

Arkenan (changes requested):
Great PR Fernando! This is a partial review, I only read the encode section, but I wanted to give you quick feedback. I'll review the decoding part tomorrow. Feel free to ask me any questions in the mean time.

lib/ssz_ex.ex
```diff
@@ -184,6 +192,145 @@ defmodule LambdaEthereumConsensus.SszEx do
...
+    {fixed_size_values, _} =
+      Enum.reduce(fixed_size_values, {[], offsets}, fn element,
+                                                       {acc_fixed_list, acc_offsets_list} ->
+        if element == :offset do
+          [offset | rest_offsets] = acc_offsets_list
+          {[offset | acc_fixed_list], rest_offsets}
+        else
+          {[element | acc_fixed_list], acc_offsets_list}
+        end
+      end)
```
Arkenan, line 253: Great reduce! A couple of improvements:

1) Move the lambda to a separate `replace_offset` function, with two clauses:

```elixir
def replace_offset(:offset, {acc_fixed_lists, [offset | rest_offsets]})

def replace_offset(element, {acc_fixed_list, acc_offset_list})
```

Also, aren't we missing a reverse here? If that's the case, why isn't a test failing?

Arkenan (changes requested):
Added some more comments. I'll look at the decode functions now.

lib/ssz_ex.ex
```diff
@@ -184,6 +192,145 @@ defmodule LambdaEthereumConsensus.SszEx do
...
+
+    fixed_size_values
+  end
+
+  defp decode_container(binary, module) do
```
Arkenan, line 258: I like the implementation! Specially that this function is clearly divided in three. The `build_deserialization_strategy` though does a bit too many things and misses some opportunities to perform some other work. I suggest refactoring this a bit into something like this:

```elixir
fixed_length = get_fixed_length(schemas)
<<fixed_binary::bitstring-size(fixed_length), variable_binary::bitstring>> = binary
with {:ok, fixed_parts, offsets} <- decode_fixed_section(fixed_binary, schemas),
     {:ok, variable_parts} <- decode_variable_section(variable_binary, offsets) do
       {:ok, struct!(fixed_parts ++ variable_parts)
end
```

Some clarifications:

```elixir
# Returns {:ok, fixed_parts, offsets, variable_binary} unless there are errors.
# fixed_parts is a list of {key, decoded_value} pairs.
# offsets is a list of {offset, schema} pairs. The first offset is 0.
defp decode_fixed_section(binary, schemas)
```

This function should be a reduce over the schemas, similar to the one you coded, but it also decodes, and the offset list is built differently (the offsets are mapped to schemas and have the fixed_length subtracted).

```elixir
# Returns {:ok, [{key, decoded_value}]}.
defp decode_variable_section
```

This function should reduce over the offsets, and in each step we get a chunk of bytes from the variable section and use the schema to decode them.

Arkenan (changes requested):
Hi! Reviewed the decoding section.

lib/ssz_ex.ex
f3r10: it is not necessary because `analyze_schemas` also does not reverse `fixed_size_values` :sweat_smile:

lib/ssz_ex.ex
f3r10: :thinking: I am not sure to understand the `decode_variable_section` function. I did almost the same approach of creating overlapping subslices over the offsets list. Since the variable elements do not have a fixed length, I think, it is necessary to check the next offset in order to know the length of the variable binary. Let me know if there is another approach that I am not seeing.

lib/ssz_ex.ex
```diff
@@ -184,6 +192,124 @@ defmodule LambdaEthereumConsensus.SszEx do
...
+    end)
+    |> then(fn {<<>>, variable_parts} ->
+      flatten_container_results(variable_parts)
+    end)
+  end
+
+  defp decode_fixed_section(binary, schemas, fixed_length) do
+    schemas
+    |> Enum.reduce({binary, [], []}, fn {key, schema},
+                                        {rest_bytes, acc_fixed_parts, acc_offsets} ->
```
Arkenan, line 286: Nit: Let's change the names of the accumulator members to binary, fixed_parts and offsets. It's understandable that we're in an iterating function and they are accumulators, and also the `rest` in `rest_bytes` makes it a bit confusing with the `rest` variable we create later.

lib/ssz_ex.ex
```diff
@@ -184,6 +192,124 @@ defmodule LambdaEthereumConsensus.SszEx do
...
+      if variable_size?(schema) do
+        <<offset::integer-size(32)-little, rest::bitstring>> = rest_bytes
```
Arkenan, line 288: Nit: Let's move the "32" to an `@offset_bits 32` module attribute.

Arkenan (approved):
LGTM! I added some small nit comments.

lib/ssz_ex.ex
Arkenan: I think your approach is great and follows the idea I was getting at!

PR #464, opened 2023-11-27
chore: cleanup & enhance spec test generator.

mpaulucci:
**Description**
- Made the runners to be added automatic instead of manual.
- Added target to make sure the spec tests are regenerated if the runners change
- Removed unused code


Makefile
```diff
@@ -79,7 +79,7 @@ $(VECTORS_DIR)/%: $(SPECTEST_ROOTDIR)/%_${SPECTEST_VERSION}.tar.gz
 	-rm -rf $@
 	tar -xzmf "$<" -C $(SPECTEST_ROOTDIR)
 
-$(SPECTEST_GENERATED_ROOTDIR): $(VECTORS_DIR)/mainnet $(VECTORS_DIR)/minimal $(VECTORS_DIR)/general lib/mix/tasks/generate_spec_tests.ex
+$(SPECTEST_GENERATED_ROOTDIR): $(VECTORS_DIR)/mainnet $(VECTORS_DIR)/minimal $(VECTORS_DIR)/general lib/spec/runners/*.ex lib/mix/tasks/generate_spec_tests.ex
```
mpaulucci, line 82: BTW, why are runners in `lib` and not in `test`folder?

MegaRedHand (approved):
🚀 🚀 🚀

Makefile
MegaRedHand: Because they are used by the mix task. We should find a way to move them to `test` in the future

PR #470, opened 2023-11-29
feat: optimize process registry updates

### LINKED ISSUE #466: Optimize `process_registry_updates`
We are currently using `List.replace_at` here:

https://github.com/lambdaclass/lambda_ethereum_consensus/blob/1431d25a00483dd04a65127ff6175b141f1b2775/lib/lambda_ethereum_consensus/state_transition/epoch_processing.ex#L162-L166

and here:

https://github.com/lambdaclass/lambda_ethereum_consensus/blob/1431d25a00483dd04a65127ff6175b141f1b2775/lib/lambda_ethereum_consensus/state_transition/epoch_processing.ex#L205

We should be transforming the list via `Stream` + `Enum` instead.

h3lio5:
**Motivation**

Optimise process registry updates by replacing List.replace_at() with a combination of Stream + Enum.

Closes #466

### DIFF
--- a/lib/lambda_ethereum_consensus/state_transition/epoch_processing.ex
+++ b/lib/lambda_ethereum_consensus/state_transition/epoch_processing.ex
@@ -167,7 +167,12 @@ defmodule LambdaEthereumConsensus.StateTransition.EpochProcessing do
             }
 
             updated_validators_list =
-              List.replace_at(updated_validators_list_acc, index, updated_validator)
+              updated_validators_list_acc
+              |> Stream.with_index()
+              |> Stream.map(fn {elem, idx} ->
+                if idx == index, do: updated_validator, else: elem
+              end)
+              |> Enum.to_list()
 
             {%{state_acc | validators: updated_validators_list}, updated_validators_list}
           end)


MegaRedHand (changes requested):
This doesn't solve the problem.

Currently, we are getting the validator indices, and then updating some other collection individually for each index. What we should do, is `zip` the collections we use (in this case only validators, so no `zip`), and then use `Stream.map` to transform them. That way we go through the list only once, instead of `length(validators)` times.

h3lio5:
> This doesn't solve the problem. See [this change](https://github.com/lambdaclass/lambda_ethereum_consensus/pull/468/files#diff-e3fa50eff866fb15ea225b270bd0bbff74af2f6d4203f7256a1ac9968c428d8dL192-R206) for an example.
> 
> Currently, we are getting the validator indices, and then updating some other collection individually for each index. What we should do, is `zip` the collections we use (in this case only validators, so no `zip`), and then use `Stream.map` to transform them. That way we go through the list only once, instead of `length(validators)` times.



> This doesn't solve the problem. See [this change](https://github.com/lambdaclass/lambda_ethereum_consensus/pull/468/files#diff-e3fa50eff866fb15ea225b270bd0bbff74af2f6d4203f7256a1ac9968c428d8dL192-R206) for an example.
> 
> Currently, we are getting the validator indices, and then updating some other collection individually for each index. What we should do, is `zip` the collections we use (in this case only validators, so no `zip`), and then use `Stream.map` to transform them. That way we go through the list only once, instead of `length(validators)` times.

Even now, we are traversing the validator_list only once, isn't it? The Stream operations (`Stream.with_index()` and `Stream.map()`) are lazy and only get applied when you actually enumerate over the list with `Enum.to_list()`? please correct me if I'm wrong.

MegaRedHand:
That stream is later turned into a list. The `Stream` vs `Enum` difference isn't too important in this case. Here's [another example](https://github.com/lambdaclass/lambda_ethereum_consensus/pull/486/files#diff-f0233547583f8e97e07a699a695ac6420f3c4a75c16c60b9adabdfd69075d4f6R135-L167).
The thing is, in `process_registry_updates` we have an `Enum.reduce` with a `List.replace` inside. Both need to iterate the validator set (or a subset), so the complexity ends up being $O(n^2)$. What we want, is to transform the `Enum.reduce(List.replace())` into a single `Enum.map` (or something similar). With that, we are reducing the complexity to $O(n)$.
I suggest you check where's the bottleneck (maybe it's not directly in `process_registry_updates` but in `activation_eligibility_and_ejections`), and if your optimizations work, by adding prints like `IO.inspect("[#{DateTime.utc_now()}] process_registry_updates")` and running the test on #475.

mpaulucci:
hey @h3lio5 , what's the status of this task?

PR #486, opened 2023-11-29
perf: optimize `process_withdrawals` and `process_sync_committee_updates`

MegaRedHand:


lib/lambda_ethereum_consensus/state_transition/accessors.ex
```diff
@@ -184,8 +184,13 @@ defmodule LambdaEthereumConsensus.StateTransition.Accessors do
...
   @spec get_total_active_balance(BeaconState.t()) :: SszTypes.gwei()
   def get_total_active_balance(state) do
-    active_validator_indices = get_active_validator_indices(state, get_current_epoch(state))
-    get_total_balance(state, active_validator_indices)
+    epoch = get_current_epoch(state)
+
+    state.validators
+    |> Stream.filter(&Predicates.is_active_validator(&1, epoch))
+    |> Stream.map(fn %Validator{effective_balance: effective_balance} -> effective_balance end)
+    |> Enum.sum()
```
mpaulucci, line 192: is there any benefit in using stream here considering that you are going to do a Enum.sum afterwards?

lib/lambda_ethereum_consensus/state_transition/accessors.ex
```diff
@@ -227,23 +232,16 @@ defmodule LambdaEthereumConsensus.StateTransition.Accessors do
...
-
     indices = get_active_validator_indices(state, epoch)
 
-    case Misc.compute_proposer_index(state, indices, seed) do
-      {:error, msg} -> {:error, msg}
-      {:ok, i} -> {:ok, i}
-    end
+    state
+    |> get_seed(epoch, Constants.domain_beacon_proposer())
+    |> then(&:crypto.hash(:sha256, &1 <> Misc.uint64_to_bytes(state.slot)))
```
mpaulucci, line 243: nit: we should wrap `crypto.hash(:sha256, x)` into a function.

lib/lambda_ethereum_consensus/state_transition/accessors.ex
MegaRedHand: We would need benchmarks to know whether `Stream` or `Enum` is more efficient for this case, but at least it should consume less memory, as it can sum the result in batches, instead of first transforming the whole list and then reducing it.

lib/lambda_ethereum_consensus/state_transition/accessors.ex
MegaRedHand: Done

Issue #480, opened 2023-11-29
[SSZ] Support bitvectors


Arkenan:
This includes:
- Encoding
- Decoding
- Merkleization

PR #494, opened 2023-12-04
feat: add bitlist and bitvector ssz support

### LINKED ISSUE #479: [SSZ] Support bitlists
This includes:
- Encoding
- Decoding
- Merkleization

f3r10:
**Motivation**

Closes #479 and #480 only encoding/decoding support


lib/ssz_ex.ex
```diff
@@ -64,6 +77,95 @@ defmodule LambdaEthereumConsensus.SszEx do
...
+
+  defp decode_bitvector(bit_vector, size) when bit_size(bit_vector) == size,
+    do: {:ok, BitVector.new(bit_vector, size)}
+
+  defp decode_bitvector(_bit_vector, _size), do: {:error, "invalid bit_vector length"}
+
+  defp encode_bitlist(bit_list, max_size) do
+    len = bit_size(bit_list)
+    bytes_len = byte_size(bit_list)
+    bytes_for_bit_len = max(1, div(len + 1 + 7, 8))
```
Arkenan, line 148: This could be `div(len, 8) + 1` instead, without the max.

lib/ssz_ex.ex
```diff
@@ -64,6 +77,95 @@ defmodule LambdaEthereumConsensus.SszEx do
...
+
+    if len > max_size do
```
Arkenan, line 150: Instead of nested `ifs`, we can use `cond`:

```elixir
cond do
  len > max_size -> {:error, "Bitlist of size #{len}, but has a max of #{max_size}"}
  bytes_for_bit_len > bytes_len -> ...
  true -> ...
end
```

lib/ssz_ex.ex
```diff
@@ -64,6 +77,95 @@ defmodule LambdaEthereumConsensus.SszEx do
...
+      {:error, "ExcessBits"}
+    else
+      if bytes_for_bit_len > bytes_len do
+        append_size = (bytes_for_bit_len - bytes_len) * @bits_per_byte
```
Arkenan, line 154: This will always be 8, no need to calculate it.

lib/ssz_ex.ex
```diff
@@ -64,6 +77,95 @@ defmodule LambdaEthereumConsensus.SszEx do
...
+        resized = <<bit_list::bitstring, 0::size(append_size)>>
+        skip = bit_size(resized) - 1
+        <<pre::bitstring-size(skip), _::size(1), rest::bitstring>> = resized
+        {:ok, <<pre::bitstring, 1::1, rest::bitstring>>}
+      else
+        <<pre::binary-size(bytes_len - 1), last::bitstring>> = bit_list
+        {:ok, <<pre::bitstring, 0::size(8 - bit_size(last) - 1), last::bitstring, 1::1>>}
+      end
```
Arkenan, line 162: Can't we replace this whole logic with something simpler? If I understand correctly, we only need to append a 1 and many 0s. Can't we just do something like this?

```elixir
padding = rem(bit_size(bit_list), 8)
<<bit_list-bitstring, 1::1, 0::size(padding)>>
```

test/unit/ssz_ex_test.exs
```diff
@@ -133,4 +133,52 @@ defmodule Unit.SSZExTest do
 
     assert_roundtrip(serialized, sync, SszTypes.SyncCommittee)
   end
+
+  test "serialize and deserialize bitlist" do
```
Arkenan, line 137: In this tests you're only adding bytestrings (number of bits multiple of 8). We should add some bitstrings too, for instance <<10::5>>

Arkenan (changes requested):

lib/ssz_ex.ex
f3r10: It depends. For example, for `<<15::size(4)>>` it would be only necessary to add  `1` but for `<<255>>` it would be necessary to add `0::size(7), 1::1` :thinking:

test/unit/ssz_ex_test.exs
f3r10: :sweat_smile: this test breaks everything

f3r10: I think these kinds of cases should return an error because getting the bitlist's length is impossible. 
The operation  to get the length is `byte_index (in this case 0) * 8 + 7 - leading_zeros (in this case 1) -> 6`

f3r10: I think that valid encoded `bitlists` are always multiple of 8 :thinking:

lib/ssz_ex.ex
```diff
@@ -72,6 +85,103 @@ defmodule LambdaEthereumConsensus.SszEx do
...
+            <<truncated::binary-size(bytes_for_bit_len), _rest::bitstring>> = formatted
+            {:ok, truncated}
+          else
+            {:ok, formatted}
+          end
+      end
+    end
+  end
+
+  defp get_bitlist_len(bit_list) do
```
Arkenan, line 126: There's a function that already does that, implemented by Mega, in the old ssz nif wrapper: https://github.com/lambdaclass/lambda_ethereum_consensus/blob/146e36d6b17be5b57f5fd2772b65dcd562e94fdf/lib/lambda_ethereum_consensus/state_transition/operations.ex#L856

lib/ssz_ex.ex
```diff
@@ -72,6 +85,103 @@ defmodule LambdaEthereumConsensus.SszEx do
     |> flatten_results_by(&Enum.join/1)
   end
 
+  defp decode_bitlist(bit_list, max_size) do
+    num_bytes = byte_size(bit_list)
+    num_bits = num_bytes * 8
+    padding = num_bits - bit_size(bit_list)
+    formatted_bit_list = <<0::size(padding), bit_list::bitstring>>
```
Arkenan, line 92: This padding is not correctly calculated. bit_size(bit_list) will be exactly equal to num_bits, as we always decode full lists.

The first thing the decode function should do is get the trailing bit and use that for the bit length.

lib/ssz_ex.ex
```diff
@@ -72,6 +85,103 @@ defmodule LambdaEthereumConsensus.SszEx do
...
+
+          remain_bits_size = len - bit_size(pre)
+          <<_::size(8 - remain_bits_size), rest::bitstring>> = rest
+          formatted = <<pre::bitstring, rest::bitstring>>
+
+          if byte_size(formatted) > bytes_for_bit_len do
+            <<truncated::binary-size(bytes_for_bit_len), _rest::bitstring>> = formatted
+            {:ok, truncated}
+          else
+            {:ok, formatted}
```
Arkenan, line 120: Instead of doing all of this I would suggest:

1. Separate the last byte from the rest of the binary, as the last byte is the only one that can have the trailing bit.
2. Clear the highest bit that's set to 1 (trailing bit).
3. If the last byte is now 0, ignore it. If not, append it to the "pre", and we're done.

Arkenan (changes requested):
Hi! I reviewed the PR and requested some more changes. Also, this is a nit, but can we have the encoding functions first and the decoding functions afterwards?

lib/ssz_ex.ex
Arkenan: Yes, my original implementation misinterpreted how the little-endian-bit-order worked. I added a new suggestion after doing some research.

lib/ssz_ex.ex
```diff
@@ -72,6 +85,103 @@ defmodule LambdaEthereumConsensus.SszEx do
...
+        <<pre::binary-size(bytes_len - 1), last::bitstring>> = bit_list
+        {:ok, <<pre::bitstring, 0::size(8 - bit_size(last) - 1), last::bitstring, 1::1>>}
+
+      true ->
+        resized_bitlist = <<0::1, bit_list::bitstring>>
+        skip = bit_size(resized_bitlist) - len - 1
+        <<pre::bitstring-size(skip), _::size(1), rest::bitstring>> = resized_bitlist
+        padding = 8 - bit_size(pre) - 1 - bit_size(rest)
+        {:ok, <<0::size(padding), pre::bitstring, 1::1, rest::bitstring>>}
+    end
```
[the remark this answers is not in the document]
Arkenan: Can we apply this change? It simplifies the code to:

```elixir
    len = bit_size(bit_list)

    if len > max_size do
      {:error, "excess bits"}
    else
      r = rem(len, 8)
      <<pre::bitstring-size(len - r), post::bitstring-size(r)>> = bit_list
      {:ok, <<pre::bitstring, 1::size(8 - r), post::bitstring>>}
    end
```

This handles the border cases, no need to have separate.

lib/ssz_ex.ex
```diff
@@ -114,6 +145,46 @@ defmodule LambdaEthereumConsensus.SszEx do
...
+
+        cond do
+          last == 0 ->
+            {:ok, pre}
+
+          pre == 0 ->
+            {:ok, <<last::size(len)>>}
+
+          true ->
+            {:ok, <<pre::size(pre_trailing_bits_size), last::size(len - pre_trailing_bits_size)>>}
```
Arkenan, line 178: This is still overly complicated and doesn't address the padding problem. We can simplify this to something shorter:

Again, let's address this. It can be done in a simpler way:

```elixir
size = byte_size(bitlist)
<<pre::size(byte_size-1), last_byte::1>> = bitlist
decoded = <<pre, remove_trailing_bit(last_byte)>>
cond do
  decoded == <<>> -> {:error, "missing length information"}
  # Other edge cases
  true -> decoded
end
```

And that's it. The only missing thing is to code the removal of the trailing bit. Instead of using the length_of_bitlist you can just code a new function that's similar to leading_zeros (matching each possible byte) that ignores the first `1` bit and returns the rest. In the case of a `<<0>>` byte you can just return `<<>>` and the rest of the functionality will stil work.

lib/ssz_ex.ex
```diff
@@ -72,6 +88,21 @@ defmodule LambdaEthereumConsensus.SszEx do
...
+    if len > max_size do
+      {:error, "excess bits"}
+    else
+      r = rem(len, @bits_per_byte)
+      <<pre::bitstring-size(len - r), post::bitstring-size(r)>> = bit_list
+      {:ok, <<pre::bitstring, 1::size(@bits_per_byte - r), post::bitstring>>}
+    end
+  end
+
+  defp encode_bitvector(bit_vector, size) when bit_size(bit_vector) == size, do: {:ok, bit_vector}
```
Arkenan, line 103: Instead of `{:ok, bit_vector}` here let's return `{:ok, BitVector.to_bytes(bit_vector)}`, which is a fix that has been added a couple of weeks ago (`BitVector.new()` was modified too).

Arkenan (changes requested):
Hi Fernando, last changes of the PR probably. Thanks for the work on this.

lib/ssz_ex.ex
f3r10: there is a warning about the opaque type:
`Call does not have expected opaque term of type LambdaEthereumConsensus.Utils.BitVector.t() in the 1st position` :thinking:

Arkenan: Can we add a `BitVector.bit_size` guard in the `BitVector` and use it here to see if the issue gets fixed? On guards: https://hexdocs.pm/elixir/main/Kernel.html#defguard/1

Arkenan (approved):
LGTM!

PR #495, opened 2023-12-04
feat: added functions to handle empty slots

### LINKED ISSUE #453: Incoming requests for blocks by range should consider empty slots
Right now, if there is an empty slot, our server returns `error_message_resource_unavailable`. But it is not considering the case in which a slot is empty and the next slot has a parent block two (or more) slots behind.

| slot n | slot n + 1  | slot n + 2 |
|--------|--------|--------|
| block m | empty | block m + 1 |

This is a perfectly valid scenario and should not return error. It should just return blocks m and m+1.

See https://eth2book.info/capella/part2/consensus/overview/

`A slot can be empty: a block proposer might be offline, or propose an invalid block, or have its block subsequently reorged out of the chain. These things should not happen often in a well-running beacon chain, but the protocol is intended to be robust when empty slots occur.`

Godspower-Eze:
**Motivation**

On rare occasions, we could have a slot with no block so we want to be able to handle this when peers try to get the block in that slot using `BeaconBlockByRange` requests.

**Description**

WIP

Closes #453


Godspower-Eze:
> Instead of comparing with the previous block, we should modify the store/DB to mark some slots as empty. That way, here we don't need to hash or compare with previous blocks. Keep in mind that there could be many slots in a row without blocks. In that case, we would need to grab a lot of unnecessary blocks if we follow this method.

There are two ways we can go about this:

1. We check for this at the point of downloading the block; that is, in the `request_blocks_by_slot` function we request 1 block earlier than the `slot` so we could use that for the checking.
2. We check this at the point of responding to `BeaconBlocksByRange` requests by using the same function `request_blocks_by_slot` and the previous slots needed.

I think the 2 option is better because this is a rare situation and it would be a bit inefficient to get extra every time.

MegaRedHand (commented):
Can you add the logic for storing empty slots? Basically, we should store an empty binary in the slot to block root "table" (and nothing on the blocks one), and handle that case in the `get_*_by_slot` functions.

Edit: just to clarify, only the `BlockStore` side of the logic. We can leave the whole "detecting when a slot is empty" for another PR.

Godspower-Eze:
I have added the logic in the `get_block_root_by_slot` and it affects `get_block_by_slot` but I am not clear on how to go about storing it in `store_block` since I don't know what the `request_blocks_by_slot` returns for empty slots and I am not writing the logic for checking.

lib/lambda_ethereum_consensus/p2p/incoming_requests/handler.ex
```diff
@@ -169,4 +170,8 @@ defmodule LambdaEthereumConsensus.P2P.IncomingRequests.Handler do
     {:ok, snappy_message} = Snappy.compress(@error_message_resource_unavailable)
     <<3>> <> size_header <> snappy_message
   end
+
+  defp create_block_response_chunk(:empty_slot) do
+    <<>>
+  end
```
MegaRedHand, line 176: nit
```suggestion
  defp create_block_response_chunk(:empty_slot), do: <<>>
```

MegaRedHand (approved):
Good work! I opened #526 to tackle the rest of the issue.

PR #496, opened 2023-12-05
Feat/merkle proof spec test

### LINKED ISSUE #462: Implement `merkle_proof` spec test runner
See: https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/merkle_proof

karasakalmt:
Closes #462

### DIFF
--- a/lib/spec/runners/light_client.ex
+++ b/lib/spec/runners/light_client.ex
@@ -1,0 +1,35 @@
+defmodule LightClientTestRunner do
+  alias LambdaEthereumConsensus.StateTransition.Predicates
+  use ExUnit.CaseTemplate
+  use TestRunner
+
+  @moduledoc """
+  Runner for LightClient test cases. See: https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/light_client
+  """
+
+  # Remove handler from here once you implement the corresponding functions
+  @disabled_handlers [
+    "single_merkle_proof",
+    "sync",
+    "update_ranking",
+  ]
+
+  @impl TestRunner
+  def skip?(%SpecTestCase{} = testcase) do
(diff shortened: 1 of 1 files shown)


lib/spec/runners/light_client.ex
```diff
@@ -0,0 +1,53 @@
...
+        case_dir <> "/object.ssz_snappy",
+        String.to_existing_atom("Elixir.SszTypes." <> testcase.suite)
+      )
+
+    %{leaf: leaf, leaf_index: leaf_index, branch: branch} =
+      YamlElixir.read_from_file!(case_dir <> "/proof.yaml")
+      |> SpecTestUtils.sanitize_yaml()
+
+    handle(testcase.handler, leaf, leaf_index, branch, object)
+  end
```
MegaRedHand, line 37: These files are specific for the `single_merkle_proof` handler. They should be moved inside the `handle` function

lib/spec/runners/light_client.ex
```diff
@@ -0,0 +1,53 @@
...
+
+    res =
+      Predicates.is_valid_merkle_branch?(
+        leaf,
+        branch,
+        Constants.deposit_contract_tree_depth() + 1,
+        leaf_index,
+        object_root
+      )
+
```
MegaRedHand, line 50: According to the test format, here we'd assert the generated proof matches the expected one, if we were capable of generating them. We should add a TODO for when/if we add merkle proof generation.

MegaRedHand (commented):
Good work!

lib/spec/runners/light_client.ex
MegaRedHand: That was fast! I thought it'd be more difficult 😅. This will help us when working on RPC

MegaRedHand (approved):
LGTM

PR #497, opened 2023-12-05
fix: don't pre-calculate hashes in `weigh_justification_and_finalization`

MegaRedHand:
We were pre-calculating block hashes with `Accessors.get_block_root` while also propagating any errors from those calls. This was a problem since those calls are done inline in the spec, and inside an `if`. If that branch wasn't taken, then our implementation would fail, differing from the spec's behavior.


mpaulucci (approved):
🎉

PR #503, opened 2023-12-06
fix: errors in `compute_proposer_index`

MegaRedHand:
This fixes two different bugs:
- we were overwriting `i` with the value of `compute_shuffled_index` (be1f9ca13418cb0756080dc3303217e10534b265)
- we were using `uint_to_bytes4` instead of `uint_to_bytes8` (709dc942f6932d13cf85e965fdf6d3e7f1fe0853)

Also some refactors: be1f9ca13418cb0756080dc3303217e10534b265...f20b58080795a2c1c9bf088f0fa8953494a1c6d6

This makes all tests in the `random` runner pass 🚀

### DIFF
--- a/lib/spec/runners/random.ex
+++ b/lib/spec/runners/random.ex
@@ -7,22 +7,22 @@ defmodule RandomTestRunner do
   use TestRunner
 
   @disabled_cases [
-    "randomized_0",
+    # "randomized_0",
     # "randomized_1",
-    "randomized_2",
-    "randomized_3",
+    # "randomized_2",
+    # "randomized_3",
     # "randomized_4",
-    "randomized_5",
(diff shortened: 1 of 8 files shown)


lib/spec/runners/random.ex
```diff
@@ -7,22 +7,22 @@ defmodule RandomTestRunner do
...
+    # "randomized_7",
     # "randomized_8",
     # "randomized_9",
-    "randomized_10",
+    # "randomized_10",
     # "randomized_11",
-    "randomized_12",
-    "randomized_13",
-    "randomized_14",
-    "randomized_15"
```
mpaulucci, line 25: 🎉

mpaulucci (approved):

PR #514, opened 2023-12-11
feat: merklelization of basic types

### LINKED ISSUE #481: [SSZ] Support merkleization for basic types (ints, bools)
We need to add support for `hash_tree_root(n)`, with `n` a `uintN` or `boolean`.

Spec is [here](https://github.com/ethereum/consensus-specs/blob/b2f2102dad0cd8b28a657244e645e0df1c0d246a/ssz/simple-serialize.md#merkleization)

We are still in the process of adding spec-tests for SSZ+hashing.

Godspower-Eze:
Closes #481


MegaRedHand (commented):
The new API should be in `SszEx`

lib/ssz_ex.ex
```diff
@@ -398,4 +404,16 @@ defmodule LambdaEthereumConsensus.SszEx do
     |> Enum.map(fn {_, schema} -> variable_size?(schema) end)
     |> Enum.any?()
   end
+
+  defp pack(value, size) when is_integer(value) and value >= 0 do
+    pad = 256 - size
```
MegaRedHand, line 409: We should name the `256`. Maybe `@leaf_size`. Wdyt?

Godspower-Eze: Yeah. Maybe `@max_bit_size`?

MegaRedHand: The value per byte is called `BYTES_PER_CHUNK` in the spec, so how about `@bits_per_chunk`?

Godspower-Eze: Changed to this. It's better

lib/ssz_ex.ex
```diff
@@ -398,4 +406,16 @@ defmodule LambdaEthereumConsensus.SszEx do
...
+
+  defp pack(value, size) when is_integer(value) and value >= 0 do
+    pad = @bits_per_chunk - size
+    <<value::size(size)-little, 0::size(pad)>>
+  end
+
+  defp pack(value) when is_boolean(value) do
+    case value do
+      true -> <<1::256-little>>
+      false -> <<0::256>>
```
MegaRedHand, line 418: These two should also be `@bits_per_chunk`

MegaRedHand (approved):

PR #515, opened 2023-12-11
perf: optimize `Misc.compute_committee` a bit

MegaRedHand:
This reduces the time of `process_attestation` by ~0.5s (~30%)

### DIFF
--- a/lib/lambda_ethereum_consensus/state_transition/misc.ex
+++ b/lib/lambda_ethereum_consensus/state_transition/misc.ex
@@ -174,6 +174,10 @@ defmodule LambdaEthereumConsensus.StateTransition.Misc do
     <<value::unsigned-integer-little-size(64)>>
   end
 
+  @doc """
+  Computes the validator indices of the ``committee_index``-th committee at some epoch
+  with ``committee_count`` committees, and for some given ``indices`` and ``seed``.
+  """
   @spec compute_committee(
           list(SszTypes.validator_index()),
           SszTypes.bytes32(),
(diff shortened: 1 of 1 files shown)


lib/lambda_ethereum_consensus/state_transition/misc.ex
```diff
@@ -182,25 +182,29 @@ defmodule LambdaEthereumConsensus.StateTransition.Misc do
...
-    Enum.reduce_while(start_..end_, {:ok, []}, fn i, {:ok, acc_list} ->
-      case compute_shuffled_index(i, length(indices), seed) do
-        {:ok, shuffled_index} ->
-          {:cont, {:ok, [Enum.at(indices, shuffled_index) | acc_list]}}
-
-        {:error, _} = error ->
-          {:halt, error}
-      end
-    end)
+    index_count = length(indices)
```
Arkenan, line 185: Nit: can we add a comment in the function as to what each parameter is: e.g. indices is the list of validator indices, index is the index of the committee being computed, count is the number of committees in the epoch.

lib/lambda_ethereum_consensus/state_transition/misc.ex
```diff
@@ -182,25 +182,29 @@ defmodule LambdaEthereumConsensus.StateTransition.Misc do
...
+    a = div(index_count * index, count)
+    b = div(index_count * (index + 1), count) - 1
```
Arkenan, line 187: Nit: can we change a and b for `committee_start` and `committee_end`?

mpaulucci (approved):

lib/lambda_ethereum_consensus/state_transition/misc.ex
```diff
@@ -182,25 +182,29 @@ defmodule LambdaEthereumConsensus.StateTransition.Misc do
...
+
+    to_swap_indices =
+      a..b//1
+      |> Stream.map(&compute_shuffled_index(&1, index_count, seed))
+      |> Stream.map(fn {:ok, i} -> i end)
```
Arkenan, line 192: What if this is not `:ok`? We should fix that edge case. E.g. with a reduce similar to what we use in SSZ when we have to encode several types and they fail.

Arkenan (approved):
This PR looks great! We only need to take the case where compute_shuffled_index fails into account.

lib/lambda_ethereum_consensus/state_transition/misc.ex
MegaRedHand: Done

PR #516, opened 2023-12-12
feat: ping metadata

### LINKED ISSUE #445: Implement `ping/1` and `metadata/2` incoming requests.
They are implemented here currently, but with hardcoded metadata: https://github.com/lambdaclass/lambda_ethereum_consensus/blob/edf4c11dea7a2251ceea4d44ee1040c7b5518ede/lib/lambda_ethereum_consensus/p2p/incoming_requests/handler.ex#L78-L101

Here are the corresponding specs:
- https://github.com/lambdaclass/lambda_ethereum_consensus/blob/main/docs/specs/p2p-interface.md#ping
- https://github.com/lambdaclass/lambda_ethereum_consensus/blob/main/docs/specs/p2p-interface.md#getmetadata-v2

karasakalmt:
Closes #445


lib/lambda_ethereum_consensus/p2p/incoming_requests/handler.ex
```diff
@@ -85,9 +85,9 @@ defmodule LambdaEthereumConsensus.P2P.IncomingRequests.Handler do
 
   defp handle_req("metadata/2/ssz_snappy", message_id, _message) do
     # Values are hardcoded
-    with {:ok, payload} <-
-           <<0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0>>
-           |> Snappy.compress() do
+    with metadata <- P2P.Metadata.get_metadata(),
```
Arkenan, line 88: We can leave the metadata assignment outside of the with, as we don't need to match its contents. We can do either:

```elixir
metadata = P2P.Metadata.get_metadata()
with {:ok, payload} <-  Ssz.to_ssz(metadata),
#...
```

or 

```elixir
with {:ok, payload} <- P2P.Metadata.get_metadata() |> Ssz.to_ssz(),
#...

File: lib/lambda_ethereum_consensus/p2p/incoming_requests/handler.ex [hunk 2]
```diff
@@ -85,9 +85,9 @@ defmodule LambdaEthereumConsensus.P2P.IncomingRequests.Handler do
...
+         {:ok, payload} <- Ssz.to_ssz(metadata),
+         {:ok, payload} <- Snappy.compress(payload) do
```
Arkenan, line 90: Let's not repeat variable names, it's very error prone. We can have "serialized" for the first one and "payload" for the second one.

lib/lambda_ethereum_consensus/p2p/metadata.ex
```diff
@@ -0,0 +1,55 @@
...
+
+  @spec get_seq_number() :: SszTypes.uint64()
+  def get_seq_number do
+    [seq_number] = get_metadata_attrs([:seq_number])
+    seq_number
+  end
+
+  @spec get_metadata() :: Metadata.t()
+  def get_metadata do
+    GenServer.call(__MODULE__, :get_metadata, @default_timeout)
```
Arkenan, line 24: Let's not add a timeout. This should always return.

lib/lambda_ethereum_consensus/p2p/metadata.ex
```diff
@@ -0,0 +1,55 @@
...
+  @impl GenServer
+  def init(metadata) do
+    {:ok, metadata}
+  end
+
+  @impl GenServer
+  def handle_call({:get_metadata_attrs, attrs}, _from, metadata) do
+    values = Enum.map(attrs, &Map.fetch!(metadata, &1))
+    {:reply, values, metadata}
+  end
```
Arkenan, line 40: Are we planning on using this call for something specific? If not, we may want to delete it.

lib/lambda_ethereum_consensus/p2p/metadata.ex
```diff
@@ -0,0 +1,55 @@
+defmodule LambdaEthereumConsensus.P2P.Metadata do
```
Arkenan, line 1: Where is this genserver being started? I think that we should add it to `application.ex`. We also need to add either a `start_link` or `child_spec` public api function for that.

Arkenan (changes requested):

lib/lambda_ethereum_consensus/p2p/metadata.ex
karasakalmt: It is for the ping in first glance but we may need to use it later not sure I can just make a `get_seq_number` instead as well

lib/lambda_ethereum_consensus/application.ex
```diff
@@ -34,6 +34,8 @@ defmodule LambdaEthereumConsensus.Application do
       {LambdaEthereumConsensus.Beacon.PendingBlocks, []},
       {LambdaEthereumConsensus.Beacon.SyncBlocks, []},
       {LambdaEthereumConsensus.P2P.GossipSub, []},
+      {LambdaEthereumConsensus.P2P.Metadata, []},
+      # Start the Endpoint (http/https)
```
mpaulucci, line 38: no need to add this comment

lib/lambda_ethereum_consensus/p2p/metadata.ex
```diff
@@ -0,0 +1,58 @@
...
+  @moduledoc """
```
mpaulucci, line 2: who sets the data that is stored in this genserver?

karasakalmt: I don't really know how can i handle that part actually forgot about it. what should i do about here

mpaulucci: I don't know. Maybe you should take a look at other clients to see where they take this data from.

lib/constants.ex
```diff
@@ -117,4 +117,10 @@ defmodule Constants do
 
   @spec proposer_score_boost() :: non_neg_integer()
   def proposer_score_boost, do: 3
+
+  @spec attestation_subnet_count() :: non_neg_integer()
+  def attestation_subnet_count, do: 2 ** 6
```
mpaulucci, line 122: is this the same constant for every network?

karasakalmt: took it from
https://github.com/lambdaclass/lambda_ethereum_consensus/blob/main/docs/specs/p2p-interface.md#configuration

MegaRedHand: Configuration -> changes depending on network

lib/lambda_ethereum_consensus/application.ex
```diff
@@ -29,6 +29,12 @@ defmodule LambdaEthereumConsensus.Application do
       {LambdaEthereumConsensus.Libp2pPort, libp2p_opts},
       {LambdaEthereumConsensus.Store.Db, []},
       {LambdaEthereumConsensus.P2P.Peerbook, []},
+      {LambdaEthereumConsensus.ForkChoice, [checkpoint_sync]},
+      {LambdaEthereumConsensus.P2P.IncomingRequests, []},
+      {LambdaEthereumConsensus.Beacon.PendingBlocks, []},
+      {LambdaEthereumConsensus.Beacon.SyncBlocks, []},
+      {LambdaEthereumConsensus.P2P.GossipSub, []},
+      {LambdaEthereumConsensus.P2P.Metadata, []},
```
MegaRedHand, line 37: I think there are some merge conflicts here 🤔. Recently we moved some of these to `LambdaEthereumConsensus.Beacon.BeaconNode`

karasakalmt: hey I don't know how that happened actually just fixed it

MegaRedHand: It's just something `git` does sometimes 😞

config/networks/mainnet/config.yaml
```diff
@@ -96,5 +88,29 @@ DEPOSIT_CONTRACT_ADDRESS: 0x00000000219ab540356cBB839Cbe05303d7705Fa
 
 # Networking
 # ---------------------------------------------------------------
+# `10 * 2**20` (= 10485760, 10 MiB)
+GOSSIP_MAX_SIZE: 10485760
```
mpaulucci, line 92: where did you get this stuff from?

karasakalmt: https://github.com/ethereum/consensus-specs/blob/dev/configs/mainnet.yaml
https://github.com/ethereum/consensus-specs/blob/dev/configs/minimal.yaml

lib/lambda_ethereum_consensus/p2p/metadata.ex
```diff
@@ -0,0 +1,100 @@
...
+  end
+
+  ##########################
+  ### Private Functions
+  ##########################
+
+  @spec get_metadata_attrs([atom()]) :: [any()]
+  defp get_metadata_attrs(attrs) do
+    GenServer.call(__MODULE__, {:get_metadata_attrs, attrs})
+  end
```
Arkenan, line 99: Noo need to have this as a separate function, let's have the call inlined in the api call.

lib/lambda_ethereum_consensus/p2p/metadata.ex
```diff
@@ -0,0 +1,100 @@
...
+
+  @impl true
+  def handle_call(:get_metadata, _from, metadata) do
+    {:reply, metadata}
+  end
+
+  @impl true
+  def handle_cast({:set_attestation_subnet, i, set}, metadata) do
+    attnets =
+      if set, do: BitVector.set(metadata.attnets, i), else: BitVector.clear(metadata.attnets, i)
```
Arkenan, line 69: Let's have this private function 

```elixir
def set_or_clear(bitvector, i, set)
```

and call it both in hear and in the other cast like follows.

```elixir
attnets = set_or_clear(metadata.attnets, i, set)
```

Arkenan (dismissed):
LGTM, I added a couple of style comments.

mpaulucci (dismissed):

karasakalmt:
@mpaulucci I tried some options to correctly run the CI / Start and stop the node (pull_request) but it doesn't work seems like you have approved with its failure is existing is this something which doesn't need to be fixed?

mpaulucci:
@karasakalmt that should work. Have you tried running the node using `make checkpoint-sync`?

Arkenan (approved):
LGTM

PR #521, opened 2023-12-13
feat: add integration to engine api.

mpaulucci:
**Description**
Added integration to engine api, with two methods:
- `engine_exchangeCapabilities`
- `engine_forkchoiceUpdatedV2`


lib/lambda_ethereum_consensus/execution/execution_client.ex
```diff
@@ -0,0 +1,109 @@
...
+  end
+
+  ##########################
+  ### GenServer Callbacks
+  ##########################
+
+  @impl GenServer
+  @spec init(any) :: {:ok, state()} | {:error, any}
+  def init([engine_endpoint, jwt_path]) do
+    if engine_endpoint do
```
mpaulucci, line 45: should we do these checks in a different module for all cli options instead of the module that uses them?

lib/lambda_ethereum_consensus/execution/execution_client.ex
```diff
@@ -0,0 +1,109 @@
+defmodule LambdaEthereumConsensus.Execution.ExecutionClient do
```
mpaulucci, line 1: maybe a Genserver is not needed here, if we decide to store all state in the config.

mpaulucci: I suspect we will have to store some state in the future that is not config, but we can refactor when the time comes

lib/lambda_ethereum_consensus/execution/execution_client.ex
MegaRedHand: I think we should. That way, we can validate the options all in one place. We also want to enable or disable some of the supervision tree depending on the CLI flags.

lib/lambda_ethereum_consensus/execution/execution_client.ex
MegaRedHand: This GenServer will become a bottleneck eventually. We are blocking it on RPC calls in each method, so each call will take quite a while.

MegaRedHand (approved):

lib/lambda_ethereum_consensus/execution/execution_client.ex
mpaulucci: done

PR #525, opened 2023-12-13
perf: add `StateTransition.Cache` module

MegaRedHand:
This PR adds a `StateTransition.Cache` module that caches various results.


lib/lambda_ethereum_consensus/state_transition/cache.ex
```diff
@@ -0,0 +1,91 @@
...
+  end
+
+  @spec cache_total_active_balance(SszTypes.BeaconState.t(), (-> SszTypes.gwei())) ::
+          SszTypes.gwei()
+  def cache_total_active_balance(state, compute_fun) do
+    epoch = Accessors.get_current_epoch(state)
+    root = get_epoch_root(state, epoch)
+
+    lookup(:total_active_balance, {epoch, root}, compute_fun)
+  rescue
```
mpaulucci, line 35: can you achieve the same using :ok, :error? With this, you might be swallowing other errors

lib/lambda_ethereum_consensus/state_transition/cache.ex
```diff
@@ -0,0 +1,91 @@
...
+    _ -> compute_fun.()
+  end
+
+  @spec cache_beacon_proposer_index(SszTypes.BeaconState.t(), (-> SszTypes.validator_index())) ::
+          SszTypes.validator_index()
+  def cache_beacon_proposer_index(%SszTypes.BeaconState{slot: slot} = state, compute_fun) do
```
mpaulucci, line 41: couldn't you just pass the lookup keys to a generic function? 

Like:

```
Cache.cache_fun(:beacon_proposer_index, {slot, root}, fn ->
      epoch = get_current_epoch(state)

      indices = get_active_validator_indices(state, epoch)

      state
      |> get_seed(epoch, Constants.domain_beacon_proposer())
      |> then(&SszEx.hash(&1 <> Misc.uint64_to_bytes(state.slot)))
      |> then(&Misc.compute_proposer_index(state, indices, &1))
    end)
)
```

lib/lambda_ethereum_consensus/state_transition/cache.ex
```diff
@@ -0,0 +1,91 @@
...
+  Caches expensive function calls.
+  """
+
+  alias LambdaEthereumConsensus.StateTransition.{Accessors, Misc}
+
+  @spec init_cache_tables() :: :ok
+  def init_cache_tables do
+    [:total_active_balance, :beacon_proposer_index, :beacon_committee, :active_validator_count]
+    |> Enum.each(fn table ->
+      :ets.new(table, [:set, :public, :named_table])
```
mpaulucci, line 12: how are values cleaned up? Does this ets grow indefinitely?

MegaRedHand: Yeah. I'm leaving the clean up for a future PR. Basically, we should leave only current + previous epoch data.

lib/lambda_ethereum_consensus/state_transition/cache.ex
MegaRedHand: Done

lib/lambda_ethereum_consensus/state_transition/cache.ex
MegaRedHand: Yeah, it's cleaner like that.

lib/lambda_ethereum_consensus/state_transition/cache.ex
```diff
@@ -0,0 +1,28 @@
+defmodule LambdaEthereumConsensus.StateTransition.Cache do
```
mpaulucci, line 1: 🎉

mpaulucci (approved):

lib/lambda_ethereum_consensus/state_transition/accessors.ex
```diff
@@ -187,12 +187,15 @@ defmodule LambdaEthereumConsensus.StateTransition.Accessors do
...
   def get_total_active_balance(state) do
     epoch = get_current_epoch(state)
+    {:ok, root} = get_epoch_root(state, epoch)
 
-    state.validators
-    |> Stream.filter(&Predicates.is_active_validator(&1, epoch))
-    |> Stream.map(fn %Validator{effective_balance: effective_balance} -> effective_balance end)
-    |> Enum.sum()
-    |> max(ChainSpec.get("EFFECTIVE_BALANCE_INCREMENT"))
+    Cache.cache_fun(:total_active_balance, {epoch, root}, fn ->
```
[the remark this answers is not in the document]
MegaRedHand: Maybe `cache_function_call`?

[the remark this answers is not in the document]
MegaRedHand: I'll change it to `Cache.lazily_compute`. We can change it later if it's confusing

PR #527, opened 2023-12-14
feat: add `Ssz` vs `SszEx` benchmark (w `Checkpoint`)

MegaRedHand:
This PR adds [`benchee`](https://github.com/bencheeorg/benchee?tab=readme-ov-file) as a `:dev` dependency and adds a simple benchmark comparing both SSZ libraries we have.

As expected, on small structs like `Checkpoint` the Elixir-native implementation is faster than the NIF one.


mpaulucci (approved):
nice!

PR #528, opened 2023-12-14
feat: merklelization of lists

### LINKED ISSUE #482: [SSZ] Support merkleization for lists
We need to add support for `hash_tree_root([...])`

Spec is [here](https://github.com/ethereum/consensus-specs/blob/b2f2102dad0cd8b28a657244e645e0df1c0d246a/ssz/simple-serialize.md#merkleization)

We are still in the process of adding spec-tests for SSZ+hashing.

Godspower-Eze:
Closes #482


lib/ssz_ex.ex
```diff
@@ -8,23 +8,29 @@ defmodule LambdaEthereumConsensus.SszEx do
   #################
   ### Public API
   #################
+  import Bitwise
 
   @bits_per_chunk 256
+  @bytes_per_chunk 32
+  @bits_per_byte 8
```
MegaRedHand, line 15: We should compute `@bits_per_chunk` by multiplying `@bits_per_byte` and `@bytes_per_chunk`

lib/ssz_ex.ex
```diff
@@ -56,16 +62,108 @@ defmodule LambdaEthereumConsensus.SszEx do
...
+    chunks_len = chunks |> byte_size() |> div(@bytes_per_chunk)
+
+    if chunks_len == 1 and leaf_count == nil do
+      chunks
+    else
+      node_count = 2 * leaf_count - 1
+      interior_count = node_count - leaf_count
+      leaf_start = interior_count * @bytes_per_chunk
+      padded_chunks = chunks |> convert_to_next_pow_of_two(leaf_count)
+      buffer = <<0::size(leaf_start * @bits_per_byte), padded_chunks::bitstring>>
```
MegaRedHand, line 113: We can leave this for the next PR, but using a binary as a buffer in Elixir [kills performance](https://stackoverflow.com/questions/49017003/elixir-replace-at-position-in-binary), in the same way using a tuple does. Each time you modify a binary, the VM has to copy it whole, so this is bound to be really slow for big enough inputs.

As a simple solution, using a list or array and then joining its values once at the end should work.

lib/ssz_ex.ex
```diff
@@ -56,16 +62,108 @@ defmodule LambdaEthereumConsensus.SszEx do
...
+          parent_index = (index - 1) |> div(2)
+          start = parent_index * @bytes_per_chunk
+          stop = (index + 1) * @bytes_per_chunk
+          focus = acc_buffer |> :binary.part(start, stop - start)
+          focus_len = focus |> byte_size()
+          children_index = focus_len - 2 * @bytes_per_chunk
+          children = focus |> :binary.part(children_index, focus_len - children_index)
+
+          <<left::bitstring-size(@bytes_per_chunk * @bits_per_byte),
+            right::bitstring-size(@bytes_per_chunk * @bits_per_byte)>> = children
```
MegaRedHand, line 129: Same here. We can change `bitstring` for `binary` and then drop the `@bits_per_byte` part

lib/ssz_ex.ex
```diff
@@ -495,15 +593,67 @@ defmodule LambdaEthereumConsensus.SszEx do
...
-      true -> <<1::@bits_per_chunk-little>>
-      false -> <<0::@bits_per_chunk>>
+  defp chunk_count({:list, :bool, max_size}) do
+    size = size_of(:bool)
+    (max_size * size + 31) |> div(32)
+  end
+
+  defp pack_basic_type_list(list, schema) do
+    list
+    |> Enum.reduce(<<>>, fn x, acc ->
```
MegaRedHand, line 612: We should replace this with a list + `Enum.join` (in a following PR)

lib/ssz_ex.ex
```diff
@@ -56,16 +62,108 @@ defmodule LambdaEthereumConsensus.SszEx do
...
+    <<value::size(size)-little>> |> pack_bytes()
+  end
+
+  @spec pack(boolean, :bool) :: binary()
+  def pack(value, :bool) do
+    case value do
+      true -> <<1::@bits_per_chunk-little>>
+      false -> @zero_chunk
+    end
+  end
```
MegaRedHand, line 151: You can do the matching in the function clause:

```elixir
def pack(true, :bool), do: <<1::@bits_per_chunk-little>>
def pack(false, :bool), do: @ zero_chunk
```

lib/ssz_ex.ex
```diff
@@ -56,16 +62,108 @@ defmodule LambdaEthereumConsensus.SszEx do
...
+
+  @spec pack(list(), {:list, any, non_neg_integer}) :: binary()
+  def pack(list, {:list, schema, _size}) do
+    if variable_size?(schema) do
+      # pack_complex_type_list(list)
```
MegaRedHand, line 156: We should raise here too (and add a TODO)

test/unit/ssz_ex_test.exs
```diff
@@ -12,6 +12,493 @@ defmodule Unit.SSZExTest do
     assert {:error, ^error_message} = SszEx.decode(serialized, schema)
   end
 
+  test "packing a list of uints" do
+    list_1 = [1, 2, 3, 4, 5]
+
+    expected_1 =
```
MegaRedHand, line 18: How did you compute these expected values?

test/unit/ssz_ex_test.exs
```diff
@@ -12,6 +12,493 @@ defmodule Unit.SSZExTest do
...
+  end
+
+  test "packing a list of bytes" do
+    list_1 = [<<1>>, <<2>>, <<3>>, <<4>>, <<5>>]
+
+    expected_1 =
+      <<1, 2, 3, 4, 5, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0>>
+
+    actual_1 = SszEx.pack(list_1, {:list, {:bytes, 8}, 5})
```
MegaRedHand, line 81: The bytes type isn't actually a basic type, as it'd be defined as `{:vector, {:int, 8}, N}` (see #593). Let's just change this for something else.

test/unit/ssz_ex_test.exs
```diff
@@ -12,6 +12,493 @@ defmodule Unit.SSZExTest do
...
+
+    list_2 = [<<255, 12>>, <<2>>, <<64, 78, 65, 90>>, <<4>>, <<5>>]
+
+    expected_2 =
+      <<255, 12, 2, 64, 78, 65, 90, 4, 5, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+        0, 0, 0, 0>>
+
+    ## the size doesn't matter because the bytes are self describing
+    size = 0
+    actual_2 = SszEx.pack(list_2, {:list, {:bytes, size}, 5})
```
MegaRedHand, line 92: Same here

test/unit/ssz_ex_test.exs
```diff
@@ -12,6 +12,493 @@ defmodule Unit.SSZExTest do
...
+    # leaf_count = 9_223_372_036_854_775_808 # 2 ** 63
+    # root = chunks |> SszEx.merklelize_chunks(leaf_count)
+    # expected_value = "9317695d95b5a3b46e976b5a9cbfcfccb600accaddeda9ac867cc9669b862979"
+    # assert root |> Base.encode16(case: :lower) == expected_value
+  end
+
+  test "hash tree root of list" do
+    ## reference: https://github.com/ralexstokes/ssz-rs/blob/1f94d5dfc70c86dab672e91ac46af04a5f96c342/ssz-rs/src/merkleization/mod.rs#L459
+
+    list = [
```
MegaRedHand, line 173: We should change this for something like `Stream.cycle([65_535]) |> Enum.take(316)`

test/unit/ssz_ex_test.exs
```diff
@@ -12,6 +12,493 @@ defmodule Unit.SSZExTest do
...
+    root = chunks |> SszEx.merklelize_chunks(2 ** 10)
+    expected_value = "2647cb9e26bd83eeb0982814b2ac4d6cc4a65d0d98637f1a73a4c06d3db0e6ce"
+    assert root |> Base.encode16(case: :lower) == expected_value
+
+    ## TOO HEAVY COMPUTATION!
+    # chunks = 1..70 |> Enum.reduce(<<>>, fn _, acc -> acc <> ones end)
+    # leaf_count = 9_223_372_036_854_775_808 # 2 ** 63
+    # root = chunks |> SszEx.merklelize_chunks(leaf_count)
+    # expected_value = "9317695d95b5a3b46e976b5a9cbfcfccb600accaddeda9ac867cc9669b862979"
+    # assert root |> Base.encode16(case: :lower) == expected_value
```
MegaRedHand, line 167: Did this take too much time to run?

MegaRedHand (commented):
Left some comments (only minor ones about performance and style). Really good work!

lib/ssz_ex.ex
```diff
@@ -495,15 +593,67 @@ defmodule LambdaEthereumConsensus.SszEx do
...
+    if len == 0 do
+      0
+    else
+      n = ((len <<< 1) - 1) |> :math.log2() |> trunc()
+      2 ** n
+    end
+  end
+
+  defp replace_chunk(chunks, start, new_chunk) do
+    <<left::bitstring-size(start * @bits_per_byte), _::size(@bytes_per_chunk * @bits_per_byte),
```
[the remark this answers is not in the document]
Godspower-Eze: removing `* @bits_per_byte` would fail. I don't know why but changing from `bitstring` to `binary` fails.

test/unit/ssz_ex_test.exs
Godspower-Eze: Packing is an intermediate step so I just tested with the values I would expect.

For correctness, the `hash_tree_root` of a list works which means this is right and it's also quite straightforward to compute.

test/unit/ssz_ex_test.exs
Godspower-Eze: Should I just add a todo on this since it depends on other issues?

test/unit/ssz_ex_test.exs
Godspower-Eze: Yeah, it took forever and didn't run

test/unit/ssz_ex_test.exs
MegaRedHand: I think we can safely remove this test.

Godspower-Eze:
The failing test is not from PR. @MegaRedHand can you please take a look?

lib/ssz_ex.ex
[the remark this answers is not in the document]
MegaRedHand: I just checked and it works on my machine™️. What's the error? Is it something related to the module? There seems to be a test with a race condition (see #607)

MegaRedHand:
Please pull the newest commits from main. It should have been fixed by #607

lib/ssz_ex.ex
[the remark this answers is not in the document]
Godspower-Eze: This has to do with the algorithm. `start` is the index of byte and `start * 8` in bits specifies the size.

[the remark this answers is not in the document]
Godspower-Eze: Done now. It's kinda tricky. The difference between `bitstring` and `binary`. 

I was thinking that the value you put in `size` should be in bits not bytes. Can you please share a resource that explains the difference between the two?

lib/ssz_ex.ex
Godspower-Eze: Does `Enum.join()` take functions?

MegaRedHand: No. I mean replacing the binary with a list of binaries. That way you can avoid expensive string concatenations while iterating, and then at the end just use `Enum.join`.

But this case shouldn't be too bad in terms of performance, since we append the data in the accumulator (the VM seems to have optimizations for this kind of case).

Godspower-Eze: Gotcha

lib/ssz_ex.ex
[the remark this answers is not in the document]
MegaRedHand: Bitstrings are strings of bits. Binaries are bitstrings with a multiple-of-8 size (i.e. strings of bytes). See https://hexdocs.pm/elixir/binaries-strings-and-charlists.html#bitstrings and https://hexdocs.pm/elixir/binaries-strings-and-charlists.html#binaries

MegaRedHand (approved):
🚀

lib/ssz_ex.ex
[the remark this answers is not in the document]
Godspower-Eze: Thanks

PR #542, opened 2023-12-19
fix: allow not passing `--execution-jwt` flag

MegaRedHand:
Currently, the Engine API isn't integrated into the rest of the node. This PR removes the `System.exit` call when `--execution-jwt` isn't specified. This is intended as a quick patch, and should ideally be solved in another way, like disabling the whole Engine API module when either the endpoint or JWT flags are unspecified (or when some special flag IS specified).

### DIFF
--- a/lib/lambda_ethereum_consensus/cli.ex
+++ b/lib/lambda_ethereum_consensus/cli.ex
@@ -24,7 +24,7 @@ defmodule LambdaEthereumConsensus.Cli do
   end
 
   defp init_engine_api_config(_endpoint, nil) do
-    Logger.error(
+    Logger.warning(
       "No jwt file provided. Please specify the path to fetch it from via the --execution-jwt flag."
     )
   end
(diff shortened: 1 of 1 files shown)


lib/lambda_ethereum_consensus/cli.ex
```diff
@@ -24,11 +24,9 @@ defmodule LambdaEthereumConsensus.Cli do
   end
 
   defp init_engine_api_config(_endpoint, nil) do
-    Logger.error(
+    Logger.warning(
       "No jwt file provided. Please specify the path to fetch it from via the --execution-jwt flag."
     )
-
-    System.stop(1)
```
mpaulucci, line 31: we still need to store the endpoint in config here, it needs a bigger refactor.

Besides that, Im fine with the change

MegaRedHand: Done

mpaulucci (approved):

PR #543, opened 2023-12-20
feat: snappy framing format

### LINKED ISSUE #224: [Snappy] Implement Snappy framing format in Elixir
_Needs some thinking_

Since the framing format is fairly simple, it's worth considering implementing [the framing format](https://github.com/google/snappy/blob/main/framing_format.txt) ourselves in Elixir (using `:snappyer` for the underlying block compression) to remove the complexity of interfacing with Rust code. This should simplify the implementation of a true streaming de/compression (as of now, it needs the whole stream in memory to work).

h3lio5:
**Motivation**

Implements the snappy decompression for the framing format.

Closes #224


lib/snappy_ex.ex
```diff
@@ -0,0 +1,38 @@
...
+      end
+
+    acc = <<acc::binary, local_acc::binary>>
+    {acc, chunks}
+  end
+
+  def process_stream_chunks(chunks, size) do
+    <<stream_data::binary-size(size), remaining_chunks::binary>> = chunks
+    {stream_data, remaining_chunks}
+  end
```
MegaRedHand, line 37: For the stream identifier:
- we have to check the stream starts with it
- we have to check it's _**exactly**_ `<<0xff, 0x06, 0x00, 0x00, 0x73, 0x4e, 0x61, 0x50, 0x70, 0x59>>`

MegaRedHand (commented):
Good work!

lib/snappy_ex.ex
```diff
@@ -0,0 +1,63 @@
...
+    <<stream_data::binary-size(size), remaining_chunks::binary>> = chunks
+    {acc, remaining_chunks}
+  end
+
+  defp process_chunks(chunks, acc, size, 0x00) do
+    # process compressed data
+    <<checksum::binary-size(4), compressed_data::binary-size(size - 4), remaining_chunks::binary>> =
+      chunks
+
+    {:ok, decompressed_data} = :snappyer.decompress(compressed_data)
```
MegaRedHand, line 50: This error must be propagated

lib/snappy_ex.ex
```diff
@@ -0,0 +1,63 @@
...
+      decompress_frames(chunks, <<>>)
+    end
+  end
+
+  defp decompress_frames(chunks, acc) do
+    if byte_size(chunks) != 0 do
+      {chunk_id, chunks} = process_chunk_id(chunks)
+      {chunks_size, chunks} = process_chunk_size(chunks)
+      {acc, chunks} = process_chunks(chunks, acc, chunks_size, chunk_id)
+      decompress_frames(chunks, acc)
```
MegaRedHand, line 17: We should make all these return `{:ok, result}` and match on them using `with`, to handle errors.

lib/snappy_ex.ex
```diff
@@ -0,0 +1,63 @@
+defmodule LambdaEthereumConsensus.SnappyEx do
```
MegaRedHand, line 1: We should add some tests for this. You can check `test/snappy_test.exs` for the rust NIF's tests.

lib/snappy_ex.ex
```diff
@@ -0,0 +1,63 @@
...
+  """
+
+  def decompress_frames(chunks) do
+    with {:ok, chunks} <- validate_stream(chunks) do
+      decompress_frames(chunks, <<>>)
+    end
+  end
+
+  defp decompress_frames(chunks, acc) do
+    if byte_size(chunks) != 0 do
```
MegaRedHand, line 13: You can split this into two function definitions:

```elixir
def decompress_frames(chunks, acc) when byte_size(chunks) == 0, do: acc

def decompress_frames(chunks, acc) do
   {chunk_id, chunks} = process_chunk_id(chunks)
   {chunks_size, chunks} = process_chunk_size(chunks)
   {acc, chunks} = process_chunks(chunks, acc, chunks_size, chunk_id)
   decompress_frames(chunks, acc)
end
```

MegaRedHand (changes requested):
Left some more comments

MegaRedHand:
Also, since [CRC-32 checksums seem to be supported by the erlang stdlib](https://www.erlang.org/doc/man/erlang#crc32-1), we could add checksum checking in this PR (should be just a call to CRC-32 and a comparison).

lib/snappy_ex.ex
```diff
@@ -9,55 +9,87 @@ defmodule LambdaEthereumConsensus.SnappyEx do
...
 
-  defp process_chunk_id(<<chunk_id::size(8), chunks::binary>>), do: {chunk_id, chunks}
+  # chunk layout: 1-byte chunk_id, 3-bytes chunk_size, remaining chunks if any data present.
+  defp process_chunk_metadata(chunks) when byte_size(chunks) < 4,
+    do: {:error, "invalid chunks metadata size"}
 
-  defp process_chunk_size(<<chunk_size::little-integer-size(24), chunks::binary>>),
-    do: {chunk_size, chunks}
+  defp process_chunk_metadata(chunks) do
+    <<chunk_id::size(8), chunks_size::size(24), chunks::binary>> = chunks
```
MegaRedHand, line 26: You have to specify little-endian for the `chunks_size` (the default is big-endian, and the spec uses little):
```suggestion
    <<chunk_id, chunks_size::little-size(24), chunks::binary>> = chunks
```

lib/snappy_ex.ex
```diff
@@ -9,55 +9,87 @@ defmodule LambdaEthereumConsensus.SnappyEx do
...
-    {acc, remaining_chunks}
+    <<_stream_data::binary-size(size), remaining_chunks::binary>> = chunks
+    {:ok, {acc, remaining_chunks}}
   end
 
+  defp process_chunks(chunks, _acc, size, 0x00)
+       when byte_size(chunks) < size,
+       do: {:error, "invalid size: compressed data chunks"}
+
+  defp process_chunks(chunks, _acc, _size, 0x00) when byte_size(chunks) > 65_540,
```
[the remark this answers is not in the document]
h3lio5: 1. sure, I'll do that.
2. sorry, this was a super dumb mistake from my side. I'll fix that.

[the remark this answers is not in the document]
MegaRedHand: NP. That's what reviews are for!

lib/snappy_ex.ex
```diff
@@ -0,0 +1,123 @@
...
+      decompress_frames(chunks, <<>>)
+    end
+  end
+
+  defp decompress_frames(chunks, acc) when byte_size(chunks) == 0, do: acc
+
+  defp decompress_frames(chunks, acc) do
+    with {:ok, {chunk_id, chunk_size, chunks}} <- process_chunk_metadata(chunks),
+         {:ok, {acc, chunks}} <- process_chunks(chunks, acc, chunk_size, chunk_id) do
+      decompress_frames(chunks, acc)
```
Arkenan, line 21: This looks great, but I suggest a slightly different division:

```elixir
with {:ok, {chunk_id, data, remaining_chunks}} <- process_chunk_metadata(chunks),
     {:ok, new_acc} <- apply_chunk(acc, id, data) do
  decompress_frames(remaining_chunks, new_acc)
end 
```

The main difference is that the size check and the chunk data extraction is common to all types, so it should be part of process_chunk_metadata. Also, apply_chunk doesn't need the size or the remaining chunks.

lib/snappy_ex.ex
```diff
@@ -0,0 +1,123 @@
+defmodule SnappyEx do
+  import Bitwise
+
+  @moduledoc """
+    SSZ library in Elixir
+  """
+  @bit_mask_32 2 ** 31 - 1
```
Arkenan, line 7: This should be `2**32-1`

lib/snappy_ex.ex
```diff
@@ -0,0 +1,123 @@
...
+  # chunk layout: 1-byte chunk_id, 3-bytes chunk_size, remaining chunks if any data present.
+  defp process_chunk_metadata(chunks) when byte_size(chunks) < 4,
+    do: {:error, "invalid chunks metadata size"}
+
+  defp process_chunk_metadata(chunks) do
+    <<chunk_id::size(8), chunks_size::little-size(24), chunks::binary>> = chunks
+    {:ok, {chunk_id, chunks_size, chunks}}
+  end
+
+  defp validate_stream(chunks) when byte_size(chunks) < 10,
```
Arkenan, line 34: Let's call this `validate_stream_id` and add a description. For example:

```elixir
# Validates that the stream identifier is present at the beginning of the stream. If successful, 
# returns the stream with the identifier removed. If not, returns an error tuple.
```

lib/snappy_ex.ex
```diff
@@ -0,0 +1,123 @@
...
+      {:error, "invalid stream identifier"}
+    end
+  end
+
+  defp process_chunks(chunks, _acc, size, 0xFF) when byte_size(chunks) < size,
+    do: {:error, "invalid size: stream identifier chunks"}
+
+  defp process_chunks(chunks, acc, size, 0xFF) do
+    # process stream identifier
+    # according to the specs, you just ignore it given the size and contents are correct
```
Arkenan, line 52: "and contents are correct". We should check if the contents are correct then! We need to check that "stream_data" corresponds to the stream id.

lib/snappy_ex.ex
```diff
@@ -0,0 +1,123 @@
...
+    <<_stream_data::binary-size(size), remaining_chunks::binary>> = chunks
+    {:ok, {acc, remaining_chunks}}
+  end
+
+  defp process_chunks(chunks, _acc, size, 0x00)
+       when byte_size(chunks) < size,
+       do: {:error, "invalid size: compressed data chunks"}
+
+  defp process_chunks(chunks, _acc, size, 0x00) when size > @chunk_size_limit,
```
Arkenan, line 61: After the change I proposed before, this guards won't be here anymore, but instead of having two separate clauses, we can always have one condition with both checks (`when byte_size(chunks) < size or size > @chunk_size_limit`).

lib/snappy_ex.ex
```diff
@@ -0,0 +1,123 @@
...
+      chunks
+
+    with {:ok, decompressed_data} <- :snappyer.decompress(compressed_data) do
+      computed_checksum = :erlang.crc32(decompressed_data)
+
+      # the crc32 checksum of the uncompressed data is masked before inserted into the frame using masked_checksum = ((checksum >> 15) | (checksum << 17)) + 0xa282ead8
+      masked_computed_checksum =
+        computed_checksum >>> 15 |||
+          ((computed_checksum <<< 17 &&& @bit_mask_32) +
+             0xA282EAD8 &&& @bit_mask_32)
```
Arkenan, line 77: The parentheses are wrong. It should be something like this:

```elixir
        (computed_checksum >>> 15 |||
           (computed_checksum <<< 17 &&& @bit_mask_32)) +
          0xA282EAD8 &&& @bit_mask_32
```

lib/snappy_ex.ex
```diff
@@ -0,0 +1,123 @@
...
+    <<masked_checksum::little-size(32), uncompressed_data::binary-size(size - 4),
+      remaining_chunks::binary>> = chunks
+
+    computed_checksum = :erlang.crc32(uncompressed_data)
+
+    # the crc32 checksum of the uncompressed data is masked before inserted into the frame using masked_checksum = ((checksum >> 15) | (checksum << 17)) + 0xa282ead8
+    masked_computed_checksum =
+      computed_checksum >>> (15 &&& @bit_mask_32) |||
+        ((computed_checksum <<< (17 &&& @bit_mask_32)) +
+           0xA282EAD8 &&& @bit_mask_32)
```
Arkenan, line 111: Instead of duplicating the code let's have a function called (mask_checksum(checksum)) that performs this. It's also less error prone.

test/unit/snappyex_test.exs
```diff
@@ -0,0 +1,81 @@
...
+  def stream_identifier, do: <<0xFF, 0x06, 0x00, 0x00, 0x73, 0x4E, 0x61, 0x50, 0x70, 0x59>>
+
+  def assert_snappy_decompress_frames(compressed, uncompressed) do
+    {:ok, ^uncompressed} =
+      compressed
+      |> Base.decode16!()
+      |> SnappyEx.decompress_frames()
+  end
+
+  test "test valid stream identifier" do
```
Arkenan, line 16: We should have three separate tests, one per assert here, so we can run them separately and they can fail separately.

Arkenan (changes requested):
Hello Akash, here are some code quality and bug related comments. They still don't fully fix the problem though. You need to keep debugging. I suggest you do some of the following:

1. Check if the masking works. Create a simple 32-bit element, like <<1, 0, 2, 0>>. Think how that would work when masked. That is, follow the rotation and then the addition, by hand (e.g. in a piece of paper). Then, check that your mask function does the same.
2. Check with a different implementation, like lighthouse or lodestar, if their implementation of checksums computes the same checksum given the same payload. If they are different we might need to configure something in our crc32 call.

Don't hesitate on printing intermediate results both here and on the other implementation.

mix.exs
```diff
@@ -53,7 +53,8 @@ defmodule LambdaEthereumConsensus.MixProject do
       {:benchee, "~> 1.2", only: [:dev]},
       {:aja, "~> 0.6.2"},
       {:dialyxir, "~> 1.1", only: [:dev, :test], runtime: false},
-      {:credo, "~> 1.7", only: [:dev, :test], runtime: false}
+      {:credo, "~> 1.7", only: [:dev, :test], runtime: false},
+      {:crc32c, "~> 1.2"}
```
Arkenan, line 57: Please change this to 

```elixir
      {:crc32c, git: "https://github.com/lambdaclass/crc32c", branch: "bump-rustler-to-29"}
```

Arkenan (changes requested):

mix.exs
h3lio5: thanks a lot! I really appreciate your help 🙏🫡

Arkenan (changes requested):

test/unit/snappyex_test.exs
```diff
@@ -0,0 +1,80 @@
...
+  #   msg =
+  #     "011CFF060000734E6150705900220000EF99F84B1C6C4661696C656420746F20756E636F6D7072657373206D657373616765"
+
+  #   "01" <> "1C" <> compressed_payload = msg
+
+  #   assert_snappy_decompress_frames(
+  #     compressed_payload,
+  #     "Failed to uncompress message"
+  #   )
+  # end
```
MegaRedHand, line 79: Do these work?

test/unit/snappyex_test.exs
```diff
@@ -0,0 +1,80 @@
...
+  def stream_identifier, do: <<0xFF, 0x06, 0x00, 0x00, 0x73, 0x4E, 0x61, 0x50, 0x70, 0x59>>
+
+  def assert_snappy_decompress_frames(compressed, uncompressed) do
+    # {:ok, value} =
+    {:ok, ^uncompressed} =
+      compressed
+      |> Base.decode16!()
+      |> SnappyEx.decompress_frames()
+
+    # IO.inspect(value, label: "snappy value")
```
MegaRedHand, line 16:
```suggestion
```

lib/snappy_ex.ex
```diff
@@ -0,0 +1,123 @@
...
+    checksum = Crc32c.calc!(data)
+
+    # the crc32 checksum of the uncompressed data is masked before inserted into the
+    # frame using masked_checksum = ((checksum >> 15) | (checksum << 17)) + 0xa282ead8
+    checksum_mask =
+      (checksum >>> 15 |||
+         (checksum <<< 17 &&& @bit_mask_32)) +
+        0xA282EAD8 &&& @bit_mask_32
+
+    checksum_mask
```
MegaRedHand, line 121:
```suggestion
    (checksum >>> 15 |||
       (checksum <<< 17 &&& @bit_mask_32)) +
      0xA282EAD8 &&& @bit_mask_32
```

test/unit/snappyex_test.exs
```diff
@@ -0,0 +1,80 @@
+defmodule Unit.SnappyExTest do
+  use ExUnit.Case
+  use ExUnitProperties
+  alias SnappyEx
+  doctest SnappyEx
+
+  def stream_identifier, do: <<0xFF, 0x06, 0x00, 0x00, 0x73, 0x4E, 0x61, 0x50, 0x70, 0x59>>
+
+  def assert_snappy_decompress_frames(compressed, uncompressed) do
+    # {:ok, value} =
```
MegaRedHand, line 10:
```suggestion
```

test/unit/snappyex_test.exs
```diff
@@ -0,0 +1,80 @@
+defmodule Unit.SnappyExTest do
+  use ExUnit.Case
+  use ExUnitProperties
+  alias SnappyEx
+  doctest SnappyEx
+
+  def stream_identifier, do: <<0xFF, 0x06, 0x00, 0x00, 0x73, 0x4E, 0x61, 0x50, 0x70, 0x59>>
```
MegaRedHand, line 7: This is unused

mpaulucci:
Closing this since Akash is no longer in the bootcamp. Will link this PR to the issue though since there is some good work that has been done.

PR #548, opened 2023-12-22
refactor: move config to runtime.exs.

mpaulucci:
**Motivation**
Config should ideally be set in the config files. Before this change, we were changing config values in application.

**Description**
- The only bad part is that we're no longer logging when there are config values that are not valid.

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -1,0 +1,43 @@
+import Config
+
+{args, _remaining_args, _errors} =
+  OptionParser.parse(System.argv(),
+    switches: [
+      network: :string,
+      checkpoint_sync: :string,
+      execution_endpoint: :string,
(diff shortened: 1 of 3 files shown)


config/runtime.exs
```diff
@@ -0,0 +1,43 @@
+import Config
+
+{args, _remaining_args, _errors} =
+  OptionParser.parse(System.argv(),
```
MegaRedHand, line 4: We should use [`parse!`](https://hexdocs.pm/elixir/1.12.3/OptionParser.html#parse!/2)

MegaRedHand (approved):

config/runtime.exs
mpaulucci: ok, will tackle in a follow up

PR #550, opened 2023-12-22
perf: store `X_epoch_participation` and `balances` in vectors

MegaRedHand:
This reduces whole block processing time to ~10.5s


mpaulucci (approved):
🎉

PR #553, opened 2024-01-02
feat: implement `notify_forkchoice_updated` function

### LINKED ISSUE #534: Implement `notify_forkchoice_updated` function that calls the Engine API
See: https://github.com/ethereum/execution-apis/blob/main/src/engine/shanghai.md#engine_forkchoiceupdatedv2

Preliminary read: 
- https://hackmd.io/@danielrachi/engine_api#Block-validation
- https://eth2book.info/capella/part3/forkchoice/bellatrix/#notify_forkchoice_updated

To calculate the safe block:
- https://eth2book.info/capella/part3/safe-block/
- https://github.com/ethereum/consensus-specs/blob/dev/fork_choice/safe-block.md

For `payloadAttributes` field, let's send `null` for now. `payloadAttributes` is set when building a block.

f3r10:
**Motivation**

Closes #534


lib/types/beacon_chain/execution_payload.ex
```diff
@@ -50,4 +51,25 @@ defmodule Types.ExecutionPayload do
   def decode(%__MODULE__{} = map) do
     Map.update!(map, :base_fee_per_gas, &Ssz.decode_u256/1)
   end
+
+  @spec prepare_execution_payload(
```
mpaulucci, line 55: who calls this function?

f3r10: oh right!!! :sweat_smile: `prepare_execution_payload` is totally wrong 
So, I think that  `notify_forkchoice_updated` should be called inside the `on_block` handler :thinking: 

I going to check the prism consensus client: https://docs.prylabs.network/docs/how-prysm-works/optimistic-sync#22-the-engine-api. I hope that the go implementation is simpler than rust one :sweat_smile:

f3r10: I already removed the `prepare_execution_payload`.
But I am not sure where `notify_forkchoice_updated` should be called :disappointed:

mpaulucci: Screenshot 2024-01-05 at 13 10 06
this is in lighthouse. So everytime you compute a new head

f3r10: oh!! ok, I think I get it now. In our case, this call should be done on `ForkChoice.Helpers.get_head`, right?.

mpaulucci: we should ideally recompute a new head for after every processed block and then `get_head` should return the cached value. But yeah, for now we could add it there.

lib/lambda_ethereum_consensus/execution/engine_api.ex
```diff
@@ -23,7 +23,7 @@ defmodule LambdaEthereumConsensus.Execution.EngineApi do
     call("engine_newPayloadV2", [execution_payload])
   end
 
-  @spec forkchoice_updated(map, map) :: {:ok, any} | {:error, any}
+  @spec forkchoice_updated(map, any) :: {:ok, any} | {:error, any}
```
mpaulucci, line 26: map | nil?

mpaulucci (commented):
Created this as a follow up https://github.com/lambdaclass/lambda_ethereum_consensus/issues/578

mpaulucci (dismissed):

mpaulucci (approved):

PR #566, opened 2024-01-04
refactor: add beacon node supervisor.

mpaulucci:
**Motivation**
We need a bigger abstraction that has the current fork and controls which topics to subscribe to. Also it should store the information not directly related to the fork choice.

### DIFF
--- a/lib/lambda_ethereum_consensus/beacon/beacon_chain.ex
+++ b/lib/lambda_ethereum_consensus/beacon/beacon_chain.ex
@@ -63,7 +63,7 @@ defmodule LambdaEthereumConsensus.Beacon.BeaconChain do
       {LambdaEthereumConsensus.Beacon.PendingBlocks, []},
       {LambdaEthereumConsensus.Beacon.SyncBlocks, []},
       {LambdaEthereumConsensus.P2P.IncomingRequests, []},
-      {LambdaEthereumConsensus.P2P.GossipSub, []},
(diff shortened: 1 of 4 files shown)


lib/lambda_ethereum_consensus/beacon/beacon_chain.ex
```diff
@@ -60,7 +60,10 @@ defmodule LambdaEthereumConsensus.ForkChoice do
 
     children = [
       {LambdaEthereumConsensus.ForkChoice.Store, {anchor_state, anchor_block}},
-      {LambdaEthereumConsensus.ForkChoice.Tree, []}
+      {LambdaEthereumConsensus.Beacon.PendingBlocks, []},
+      {LambdaEthereumConsensus.Beacon.SyncBlocks, []},
+      {LambdaEthereumConsensus.P2P.IncomingRequests, []},
+      {LambdaEthereumConsensus.P2P.GossipSub, []},
```
mpaulucci, line 66: in a follow up, we can calculate the fork digest from the anchor state and pass it down to the gossipsub and incoming requests to subscribe to the topics with the proper digest.

Eventually we will have to create a more complex logic, because when the fork changes we have to subscribe to the new fork topics.

MegaRedHand (dismissed):
LGTM

MegaRedHand (approved):

PR #569, opened 2024-01-04
feat: add beacon chain with some data that was in fork choice.

mpaulucci:
**Motivation**
We want to move away from having store as a God object that stores all the data of the chain. This is a first step in that direction.

**Description**
Moved the time calculation to beacon chain. This is a helper function used all over the place and should not be slowed down by block processing.

### DIFF
--- a/lib/lambda_ethereum_consensus/beacon/beacon_chain.ex
+++ b/lib/lambda_ethereum_consensus/beacon/beacon_chain.ex
@@ -15,9 +15,9 @@ defmodule LambdaEthereumConsensus.Beacon.BeaconChain do
     ]
 
     @type t :: %__MODULE__{
-      genesis_time: Types.uint64(),
-      time: Types.uint64()
-    }
+            genesis_time: Types.uint64(),
+            time: Types.uint64()
+          }
   end
 
   @spec start_link({BeaconState.t(), Types.uint64()}) :: :ignore | {:error, any} | {:ok, pid}
@@ -27,7 +27,7 @@ defmodule LambdaEthereumConsensus.Beacon.BeaconChain do
 
   @spec get_current_slot() :: integer()
   def get_current_slot do
-    GenServer.call(__MODULE__, {:get_current_slot})
+    GenServer.call(__MODULE__, :get_current_slot)
   end
 
   ##########################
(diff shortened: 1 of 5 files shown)


lib/lambda_ethereum_consensus/beacon/beacon_chain.ex
```diff
@@ -0,0 +1,66 @@
...
+  def init({anchor_state = %BeaconState{}, time}) do
+    schedule_next_tick()
+    {:ok, %BeaconChainState{
+      genesis_time: anchor_state.genesis_time,
+      time: time
+    }}
+  end
+
+  @impl true
+  def handle_call({:get_current_slot}, _from, state) do
```
MegaRedHand, line 48: Nit: 
```suggestion
  def handle_call(:get_current_slot, _from, state) do
```

lib/lambda_ethereum_consensus/fork_choice/store.ex
```diff
@@ -195,20 +191,7 @@ defmodule LambdaEthereumConsensus.ForkChoice.Store do
     GenServer.call(__MODULE__, {:get_store_attrs, attrs}, @default_timeout)
   end
 
-  @spec on_tick_now(Store.t()) :: Store.t()
-  defp on_tick_now(store) do
-    new_store = Handlers.on_tick(store, :os.system_time(:second))
-    current_slot = Store.get_current_slot(new_store)
-    :telemetry.execute([:sync, :store], %{slot: current_slot})
```
MegaRedHand, line 202: We should move this too

lib/lambda_ethereum_consensus/beacon/beacon_node.ex
```diff
@@ -58,8 +58,11 @@ defmodule LambdaEthereumConsensus.Beacon.BeaconNode do
   defp init_children(anchor_state, anchor_block) do
     Cache.initialize_cache()
 
+    time = :os.system_time(:second) |> floor()
```
MegaRedHand, line 61: `:os.system_time` returns an integer, so no need to use floor
```suggestion
    time = :os.system_time(:second)
```

lib/lambda_ethereum_consensus/beacon/beacon_chain.ex
```diff
@@ -0,0 +1,66 @@
...
+    {:reply, div(state.time - state.genesis_time, ChainSpec.get("SECONDS_PER_SLOT")), state}
+  end
+
+  @impl true
+  def handle_info(:on_tick, state) do
+    schedule_next_tick()
+    time = :os.system_time(:second) |> floor()
```
MegaRedHand, line 55: Here too

MegaRedHand (commented):
Left some minor comments. Otherwise, LGTM 🚀

MegaRedHand (approved):
LGTM

MegaRedHand (approved):

PR #574, opened 2024-01-04
feat: clean up old cache entries

### LINKED ISSUE #531: Clean up old caches when advancing an epoch

MegaRedHand:
Closes #531

### DIFF
--- a/lib/lambda_ethereum_consensus/state_transition/cache.ex
+++ b/lib/lambda_ethereum_consensus/state_transition/cache.ex
@@ -2,14 +2,46 @@ defmodule LambdaEthereumConsensus.StateTransition.Cache do
   @moduledoc """
   Caches expensive function calls.
   """
+  require Ex2ms
 
   @tables [
+    # k = {epoch, root} ; v = int
     :total_active_balance,
+    # k = {slot, root} ; v = int
     :beacon_proposer_index,
+    # k = {epoch, root} ; v = int
     :active_validator_count,
+    # k = {slot, {index, root}} ; v = [index]
     :beacon_committee,
+    # k = {epoch, root} ; v = Aja.vec(index)
     :active_validator_indices
   ]
+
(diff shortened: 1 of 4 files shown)


mix.exs
```diff
@@ -47,6 +47,7 @@ defmodule LambdaEthereumConsensus.MixProject do
       {:telemetry_poller, "~> 1.0"},
       {:telemetry_metrics, "~> 0.6"},
       {:telemetry_metrics_prometheus, "~> 1.1.0"},
+      {:ex2ms, "~> 1.6", runtime: false},
```
mpaulucci, line 50: what is this for?

lib/lambda_ethereum_consensus/state_transition/cache.ex
```diff
@@ -30,9 +62,15 @@ defmodule LambdaEthereumConsensus.StateTransition.Cache do
...
       [] ->
+        clean_up_old_entries(table, key)
         value = compute_fun.()
         :ets.insert_new(table, {key, value})
         value
     end
   end
+
+  defp clean_up_old_entries(table, key) do
+    match_spec = generate_cleanup_spec(table, key)
```
mpaulucci, line 73: is this potentially an expensive operation?

MegaRedHand: It shouldn't be. It is necessary, however, and I already have some thoughts on how to optimize it in case it starts impacting performance.

mix.exs
MegaRedHand: Erlang had a kind-of-macro system that was used by the `ets` module for the [`fun2ms/1` "function"](https://www.erlang.org/doc/man/ets#fun2ms-1). This is the Elixir equivalent.

mpaulucci (dismissed):

mpaulucci (approved):

PR #575, opened 2024-01-04
fix: run `mix clean` on `clean` target

MegaRedHand:
When running `make clean`, following runs could fail to load rustler-compiled artifacts. I guess this is because we are removing these on `make clean`, but didn't remove the NIF modules so they weren't recompiled.


Arkenan (approved):
Nice, LGTM

PR #576, opened 2024-01-05
fix: fetch block when checkpoint sync's race condition happens

### LINKED ISSUE #565: Remove race condition on checkpoint-sync's state and block fetching
We currently are fetching the block and state concurrently. We should later check that the two correspond to the same slot, and if not, fetch the correct block/state by specifying the slot in the requested url.

MegaRedHand:
Fixes #565

New workflow:
1. fetch both state and block optimistically
2. if slots mismatch, we just ask for the block on the state's slot

We might want to ask for the state if the block has a higher slot, but I leave that for future PRs (but it's not a priority).


mpaulucci (approved):
nice!

PR #577, opened 2024-01-05
feat(beacon_api): add id_to_state_root function

### LINKED ISSUE #557: [Beacon API] id_to_state_root function
We have different ways to address states:

- [ ] `"genesis"` literal. This represents the genesis state according to the node.
- [ ] `"justified"`. This represents the last justified state according to the node.
- [ ] `"finalized"`. This represents the last finalized state according to the node.
- [ ] `"head"` string. The current head state according to the node.
- [ ] by a root. The most direct addressing is the root of a state.

We can have a parser in `beacon_api/utils.ex`:

```elixir
parse_id("genesis") -> :genesis
parse_id("justified") -> :justified
parse_id("finalized") -> :finalized
parse_id("head") -> :head
parse_id("0x" <> hex_root) -> <<root>>
parse_id(other) -> :invalid_id
```

We can then have a `root_by_id(id)` function that takes any of the previous results and returns a root that can be used to get a state using the db.

```elixir
root = Utils.parse_id(id) |> ForkChoice.Utils.root_by_id()
```

The checkboxes in the issue mark the completion of each specific parsing and root_by_id clause. This task will be considered finished when the `get_state_root` endpoint of the beacon api uses this function.

f3r10:
**Motivation**

Closes #557


lib/beacon_api/utils.ex
```diff
@@ -10,4 +10,25 @@ defmodule BeaconApi.Utils do
...
   end
+
+  @spec parse_id(binary) :: atom | binary
+  def parse_id("genesis"), do: :genesis
+  def parse_id("justified"), do: :justified
+  def parse_id("finalized"), do: :finalized
+  def parse_id("head"), do: :head
+
+  def parse_id("0x" <> hex_root) do
+    if byte_size(hex_root) == 66 do
```
MegaRedHand, line 21: Shouldn't this be 64, since we already removed the leading "0x" ?

Also, I'm nitpicking, but the if could be moved to the clause guard:

```elixir
def parse_id("0x" <> hex) when byte_size(hex) == 64 do ...
def parse_id("0x" <> _hex), do: :invalid_id
```

lib/beacon_api/utils.ex
```diff
@@ -10,4 +10,25 @@ defmodule BeaconApi.Utils do
...
+      <<hex_root::binary>>
```
MegaRedHand, line 22: Shouldn't we use `Base.decode16` ?

lib/lambda_ethereum_consensus/fork_choice/helpers.ex
```diff
@@ -207,4 +212,58 @@ defmodule LambdaEthereumConsensus.ForkChoice.Helpers do
...
+  def root_by_id(:head) do
+    with {:ok, current_status} <- BeaconChain.get_current_status_message(),
+         {:ok, signed_block} <- BlockStore.get_block(current_status.head_root) do
+      # TODO compute is_optimistic_or_invalid
+      execution_optimistic = true
+      {:ok, {signed_block.message.state_root, execution_optimistic, false}}
+    end
+  end
+
+  def root_by_id(:genesis), do: :invalid_id
```
MegaRedHand, line 227: I think we should differentiate between "invalid" and "unsupported" values. Although, in this case we can just return `:not_found`

lib/lambda_ethereum_consensus/fork_choice/helpers.ex
```diff
@@ -207,4 +212,58 @@ defmodule LambdaEthereumConsensus.ForkChoice.Helpers do
...
+    execution_optimistic = true
+    finalized = false
+    {:ok, {hex_root, execution_optimistic, finalized}}
+  end
+
+  def root_by_id(slot) when is_integer(slot) do
+    with {:ok, state} <- StateStore.get_latest_state(),
+         {:ok, signed_block} <- BlockStore.get_block_by_slot(state.slot),
+         {:ok, store} <- get_forkchoice_store(state, signed_block.message),
+         current_slot = Store.get_current_slot(store),
```
MegaRedHand, line 260: We can get the current slot via the new `BeaconNode`.

lib/lambda_ethereum_consensus/fork_choice/helpers.ex
```diff
@@ -207,4 +212,58 @@ defmodule LambdaEthereumConsensus.ForkChoice.Helpers do
...
+         true <- slot < current_slot,
```
MegaRedHand, line 261: This is a bad way of checking. You should use a `check_x(...)` function that internally maps the `true`/`false` into `:ok`/`some_kind_of_error`. That way you don't just return `false` when this validation fails.

MegaRedHand (commented):
Good work!

lib/lambda_ethereum_consensus/fork_choice/helpers.ex
```diff
@@ -207,4 +212,58 @@ defmodule LambdaEthereumConsensus.ForkChoice.Helpers do
     current_epoch = Misc.compute_epoch_at_slot(current_slot)
     store.justified_checkpoint.epoch + 1 == current_epoch
   end
+
+  @spec root_by_id(atom() | Types.root() | Types.slot()) ::
+          {:ok, {Types.root(), boolean(), boolean()}} | {:error, String.t()} | atom()
+  def root_by_id(:head) do
```
MegaRedHand, line 218: We should make this `root_by_id` function return the block's root, instead of the state root. Then we can easily map that hash to the state root in the calling function. This has the bonus that it's easily reusable, and so it also removes most of the repeated code we can have.

f3r10: do you mean to return `Types.BeaconBlock` ?

MegaRedHand: No. Just `Types.root()`, but the block's hash (you shouldn't compute it, though), instead of the blocks's `state_root`.

f3r10: :thinking: so for example to get  `eth/v1/beacon/states/justified/root` instead of 
```elixir
    with {:ok, justified_checkpoint} <- ForkChoice.get_justified_checkpoint(),
         {:ok, signed_block} <- BlockStore.get_block(justified_checkpoint.root) do
      # TODO compute is_optimistic_or_invalid
      execution_optimistic = true
      {:ok, signed_block.message.state_root, execution_optimistic, false}
    end

```
would be:
```elixir
    with {:ok, justified_checkpoint} <- ForkChoice.get_justified_checkpoint() do
      # TODO compute is_optimistic_or_invalid
      execution_optimistic = true
      {:ok, justified_checkpoint.root, execution_optimistic, false}
    end

```
?

MegaRedHand: Yes. However, after calling `root_by_id` you'd need to call some other function `get_state_root(root)` to get the state root.

f3r10: :thinking: but `get_state_root` is the one that calls `root_by_id` :sweat_smile: 
a similar function that could be used is `StateStore.get_state_root_by_slot(slot)`

f3r10: I created this function that could be useful:
```elixir
  @spec get_state_root(Types.root()) :: {:ok, Types.root()} | {:error, String.t()} | :not_found
  def get_state_root(root) do
    with {:ok, signed_block} <- BlockStore.get_block(root) do
      {:ok, signed_block.message.state_root}
    end
  end
```

lib/beacon_api/utils.ex
```diff
@@ -10,4 +10,20 @@ defmodule BeaconApi.Utils do
...
     String.starts_with?(value, "0x") and byte_size(value) == 66
   end
+
+  @spec parse_id(binary) :: atom | binary
+  def parse_id("genesis"), do: :genesis
+  def parse_id("justified"), do: :justified
+  def parse_id("finalized"), do: :finalized
+  def parse_id("head"), do: :head
+
+  def parse_id("0x" <> hex_root) when byte_size(hex_root) == 64, do: Base.decode16!(hex_root)
```
MegaRedHand, line 20: We should use the fallible version (the one without the `!`) and map errors to `:invalid_id`

lib/lambda_ethereum_consensus/fork_choice/helpers.ex
```diff
@@ -207,4 +211,63 @@ defmodule LambdaEthereumConsensus.ForkChoice.Helpers do
...
+  def get_state_root(root) do
+    with {:ok, signed_block} <- BlockStore.get_block(root) do
+      {:ok, signed_block.message.state_root}
+    end
+  end
+
+  defp check_valid_slot(slot, current_slot) when slot < current_slot, do: :ok
+
+  defp check_valid_slot(slot, _current_slot),
+    do: {:error, "slot #{slot} cannot be greater that current slot"}
```
MegaRedHand, line 272: Typo:
```suggestion
    do: {:error, "slot #{slot} cannot be greater than current slot"}
```

MegaRedHand (commented):
LGTM. However, this PR also tackles the issue that #586 was trying to solve. Next time, try to keep the PR in scope so this doesn't happen, or at least mention it on the issue or telegram.

MegaRedHand (approved):

PR #583, opened 2024-01-08
fix: remove hardcoded fork digest values

### LINKED ISSUE #554: Use computed `fork_digest` instead of hardcoded one.
We have some places where we use a hardcoded fork digest, see https://github.com/search?q=repo%3Alambdaclass%2Flambda_ethereum_consensus%20BBA4DA96&type=code

This digest corresponds to capella in Mainnet, but we should be able to calculate it for any fork of any network.

We should compute the fork digest for the current slot, like lighthouse does: https://github.com/sigp/lighthouse/blob/441fc1691b69f9edc4bbdc6665f3efab16265c9b/consensus/types/src/chain_spec.rs#L218-L228

We also need to update this function with the fork digest for the current slot: https://github.com/lambdaclass/lambda_ethereum_consensus/blob/d1375bfc8767c0fd350c0fa92e9b3cc66b52be74/lib/lambda_ethereum_consensus/fork_choice/helpers.ex#L18

mpaulucci:
**Motivation**
We need to compute fork digest dynamically to support multiple networks and forks.

Closes #554

### DIFF
--- a/lib/lambda_ethereum_consensus/p2p/gossip/gossipsub.ex
+++ b/lib/lambda_ethereum_consensus/p2p/gossip/gossipsub.ex
@@ -27,7 +27,7 @@ defmodule LambdaEthereumConsensus.P2P.GossipSub do
       # {"sync_committee_0", Types.SyncCommitteeMessage}
     ]
 
-    fork_context = BeaconChain.get_fork_digest() |> Base.encode16!()
+    fork_context = BeaconChain.get_fork_digest() |> Base.encode16!(case: :lower)
 
     children =
       for {topic_msg, ssz_type, handler} <- topics do
--- a/lib/lambda_ethereum_consensus/p2p/gossip/gossipsub.ex
+++ b/lib/lambda_ethereum_consensus/p2p/gossip/gossipsub.ex
@@ -4,6 +4,7 @@ defmodule LambdaEthereumConsensus.P2P.GossipSub do
   """
   use Supervisor
 
+  alias LambdaEthereumConsensus.Beacon.BeaconChain
   alias LambdaEthereumConsensus.P2P.Gossip.Consumer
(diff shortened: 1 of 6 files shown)


lib/lambda_ethereum_consensus/p2p/gossip/gossipsub.ex
```diff
@@ -26,9 +27,11 @@ defmodule LambdaEthereumConsensus.P2P.GossipSub do
       # {"sync_committee_0", Types.SyncCommitteeMessage}
     ]
 
+    fork_context = BeaconChain.get_fork_digest() |> Base.encode16!()
```
MegaRedHand, line 30:
```suggestion
    fork_context = BeaconChain.get_fork_digest() |> Base.encode16!(case: :lower)
```

lib/lambda_ethereum_consensus/p2p/incoming_requests/handler.ex
```diff
@@ -130,12 +127,14 @@ defmodule LambdaEthereumConsensus.P2P.IncomingRequests.Handler do
   defp create_block_response_chunk({:ok, block}) do
     with {:ok, ssz_signed_block} <- Ssz.to_ssz(block),
          {:ok, snappy_ssz_signed_block} <- Snappy.compress(ssz_signed_block) do
+      fork_context = BeaconChain.get_fork_digest()
```
MegaRedHand, line 130: We should add a comment on how this doesn't take into account inter-fork requests. If we ask for blocks of an earlier hard-fork, we will fail because our fork digest uses the newer fork version.

mpaulucci: yes, make sense.

MegaRedHand (approved):
🚀

PR #586, opened 2024-01-09
feat: added get_state_root_by_root

### LINKED ISSUE #556: [Beacon API] Get state root by root should use the DB.
We should modify the `beacon_controller.ex` file to add a call to the DB when getting the state root by root.

karasakalmt:
Closes #556


MegaRedHand:
Closing this because it was already solved in #577. Sorry for the trouble!

karasakalmt:
> Closing this because it was already solved in #577. Sorry for the trouble!

Hey I guess the difference is the current implementation doesn't use DB as i understand
https://github.com/f3r10/lambda_ethereum_consensus/blob/ef2409dedb710ea186426145be8497dbd90bae4f/lib/lambda_ethereum_consensus/fork_choice/helpers.ex#L245
This one uses DB and checks the existence of the state root. I don't know if it was necessary though.

MegaRedHand:
The current implementation retrieves the state root from the block here:

https://github.com/lambdaclass/lambda_ethereum_consensus/blob/91469bcde7b01aee9ffa906d9d031a6cb1f11d85/lib/beacon_api/controllers/v1/beacon_controller.ex#L9-L11

But the behavior we'd want is the one explained in #619. I can assign that issue to you if you'd like!

karasakalmt:
> The current implementation retrieves the state root from the block here:
> 
> https://github.com/lambdaclass/lambda_ethereum_consensus/blob/91469bcde7b01aee9ffa906d9d031a6cb1f11d85/lib/beacon_api/controllers/v1/beacon_controller.ex#L9-L11
> 
> But the behavior we'd want is the one explained in #619. I can assign that issue to you if you'd like!

Assigned myself

PR #589, opened 2024-01-09
chore: rename ForkChoice.Store to ForkChoice.

mpaulucci:
**Description**
Also moved `get_current_status_message` to BeaconChain,

### DIFF
--- a/lib/lambda_ethereum_consensus/beacon/beacon_chain.ex
+++ b/lib/lambda_ethereum_consensus/beacon/beacon_chain.ex
@@ -48,6 +48,24 @@ defmodule LambdaEthereumConsensus.Beacon.BeaconChain do
     GenServer.call(__MODULE__, {:get_fork_digest, slot})
   end
 
+  @spec get_current_status_message() :: {:ok, Types.StatusMessage.t()} | {:error, any}
+  def get_current_status_message do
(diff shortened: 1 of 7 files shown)


lib/lambda_ethereum_consensus/beacon/beacon_chain.ex
```diff
@@ -48,6 +48,24 @@ defmodule LambdaEthereumConsensus.Beacon.BeaconChain do
     GenServer.call(__MODULE__, {:get_fork_digest, slot})
   end
 
+  @spec get_current_status_message() :: {:ok, Types.StatusMessage.t()} | {:error, any}
+  def get_current_status_message do
```
MegaRedHand, line 52: I don't think we should move this function. It's currently hardcoded, but because of how it's implemented it'd need to communicate with `ForkChoice`.

mpaulucci: The values in Beacon Chain might be a bit older compared to the ones in fork choice, but this way stuff like on_block doesn't block any getter.

MegaRedHand: Sounds good

MegaRedHand (approved):
LGTM

PR #592, opened 2024-01-09
docs: concurrency proposal - per-block processes

### LINKED ISSUE #614: Per-block processes proposal
We need to change the concurrency model. A documented proposal would be a good starting point.

Arkenan:
This is a description of the PoC under work: a single process per block for simpler concurrency and independent processing.

Closes #614

### DIFF
--- a/docs/concurrency_design.md
+++ b/docs/concurrency_design.md
@@ -188,7 +188,6 @@ We can see here that:
 sequenceDiagram
 
 participant bp as Block Processor (Genserver)
-participant store as Fork Choice Store <br> (Genserver)
 participant db as Persistent Store <br> (DB/Cache)
 participant dt as Download Task
 participant bpp as Parent Block Processor (Genserver)
--- a/docs/concurrency_design.md
+++ b/docs/concurrency_design.md
@@ -150,7 +150,7 @@ There are some tasks that, of course, will require interaction with the outside 
 
 - When a parent is missing, we need to download it. The block processor may pause until it receives a notification that the parent was downloaded and transitioned. After that, it may continue with its own state transition.
 - When validating it will need to interact with the global fork choice store. This just requires validating a few parameters like the epoch/slot, which is a fast process, so it’s a relatively simple call without passing the full block and without changing the store state.
-- The final step is updating the store, which still doesn’t need to pass around big state and is fast. If additional validations are performed or repeated, it doesn’t represent a big performance hit.
+- The final step is updating the store/beacon chain state, which still doesn’t need to pass around big state and is fast. If additional validations are performed or repeated, it doesn’t represent a big performance hit.
 
 ### Happy path diagram
 
@@ -159,7 +159,7 @@ sequenceDiagram
 
 participant port as LibP2P Port <br> (Genserver)
 participant bp as Block Processor (Genserver)
-participant store as Fork Choice Store <br> (Genserver)
+participant store as Fork Choice <br> (Genserver)
 participant db as Persistent Store <br> (DB/Cache)
 
 port ->> bp: gossip(block)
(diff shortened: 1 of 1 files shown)


MegaRedHand (dismissed):
LGTM. I already want to check the PoC 🔥

docs/concurrency_design.md
```diff
@@ -129,3 +129,90 @@ sequenceDiagram
 Some pending definitions:
 
 - The block tracker could eventually be a block cache, and maintain blocks and their state in an ETS that can be accessed easily by other processes.
+
+## Another possible design: per-block processes
+
+Conceptually, the core concurrency issue of the current design is that we use processes to represent steps instead of them representing independent pieces of computation/data. Most of the inter-process communication is call-based and sequential, effectively blocking processes while still adding the overhead of message passing and mailboxes with big objects like blocks and states.
+
+What we need is to have one process per independent task. In our case, our independent unit for is **the block processing**. We need to have a **single process per block**, that takes care of:
```
mpaulucci, line 137: so, you mean a single process per block, and not a separate process for block processing? That would only benefit us if we have a fork, in normal times (a straight line of blocks) it won't make a difference. But it makes sense to optimize for the worst scenario, not for the optimal one.

docs/concurrency_design.md
```diff
@@ -129,3 +129,90 @@ sequenceDiagram
...
+
+With this strategy in mind, the block won’t need to be passed around during its lifetime, and state transitions may happen in parallel without the need to spawn tasks on demand.
+
+There are some tasks that, of course, will require interaction with the outside world:
+
+- When a parent is missing, we need to download it. The block processor may pause until it receives a notification that the parent was downloaded and transitioned. After that, it may continue with its own state transition.
+- When validating it will need to interact with the global fork choice store. This just requires validating a few parameters like the epoch/slot, which is a fast process, so it’s a relatively simple call without passing the full block and without changing the store state.
+- The final step is updating the store, which still doesn’t need to pass around big state and is fast. If additional validations are performed or repeated, it doesn’t represent a big performance hit.
+
+### Happy path diagram
```
mpaulucci, line 155: just FYI, we also have a new abstraction called BeaconChain that should take some of the responsabilty that ForkChoice has today.

docs/concurrency_design.md
```diff
@@ -129,3 +129,90 @@ sequenceDiagram
...
+activate bpp
+bpp ->> db: store_block_if_not_present(block)
+bpp ->> bpp: validation and state transition
+bpp ->> db: put_state(new_parent_state)
+deactivate bpp
+db ->>+ bp: parent_ready()
+bp ->>- bp: validation and state transition
+```
+
+This is a simplified sequence diagram highlighting the differences when the parent block is not present. To summarize:
```
mpaulucci, line 211: you mean simplified because it doesn't show the recursive nature of the problem, right?

mpaulucci (dismissed):
🎉

docs/concurrency_design.md
Arkenan: Yes

docs/concurrency_design.md
Arkenan: Yes, this is one process per block, although there are several benefits:
1. It is true, works for the worst case scenario.
2. Naturally unblocks every step of the pipeline. That is, pending blocks won't block while it waits for a state transition to happen. The store or beacon chain won't block either.
3. If pending blocks is not blocked, then we can process all of the incoming gossip continuously, and discard the ones that are not valid early, without their validation depending on other blocks being processed. This also reduces the size of the processes mailbox.
4. Everything happens in parallel when needed (state transition, deserialization, decompression, etc).

docs/concurrency_design.md
```diff
@@ -129,3 +129,90 @@ sequenceDiagram
...
+- The final step is updating the store, which still doesn’t need to pass around big state and is fast. If additional validations are performed or repeated, it doesn’t represent a big performance hit.
+
+### Happy path diagram
+
+```mermaid
+sequenceDiagram
+
+participant port as LibP2P Port <br> (Genserver)
+participant bp as Block Processor (Genserver)
+participant store as Fork Choice Store <br> (Genserver)
```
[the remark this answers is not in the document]
Arkenan: Renamed here as well

MegaRedHand (approved):

PR #594, opened 2024-01-10
fix: remove hardcoded fork digest from libp2p code.

mpaulucci:
**Motivation**
We need to be able to have different fork digests depending on the network, hard fork, etc..

**Description**
This PR removes the last reference to a hardcoded fork digest in the Go code related to libp2p.


lib/lambda_ethereum_consensus/beacon/beacon_node.ex
```diff
@@ -55,8 +55,10 @@ defmodule LambdaEthereumConsensus.Beacon.BeaconNode do
       {LambdaEthereumConsensus.Beacon.BeaconChain, {anchor_state, time}},
       {LambdaEthereumConsensus.ForkChoice.Store, {anchor_state, anchor_block, time}},
       {LambdaEthereumConsensus.Beacon.PendingBlocks, []},
-      {LambdaEthereumConsensus.Beacon.SyncBlocks, []},
+      {LambdaEthereumConsensus.Libp2pPort, []},
```
mpaulucci, line 58: now `Libp2pPort` depends on BeaconChain to fetch the appropriate fork digest.

lib/lambda_ethereum_consensus/beacon/pending_blocks.ex
```diff
@@ -114,6 +114,12 @@ defmodule LambdaEthereumConsensus.Beacon.PendingBlocks do
     end)
   end
 
+  @impl true
```
mpaulucci, line 117: little bug, we were calling `BlockDownloader.request_blocks_by_root()` with an empty array sometimes

lib/lambda_ethereum_consensus/p2p/peerbook.ex
```diff
@@ -70,7 +70,7 @@ defmodule LambdaEthereumConsensus.P2P.Peerbook do
       |> min(len - @target_peers)
       |> max(0)
 
-    n = :rand.uniform(len)
+    n = if(len > 0, do: :rand.uniform(len), else: 0)
```
mpaulucci, line 73: if `len` is 0, :rand.uniform(len) crashes

lib/libp2p_port.ex
```diff
@@ -66,6 +70,7 @@ defmodule LambdaEthereumConsensus.Libp2pPort do
       should be started.
     * `:discovery_addr` - the address used by the discovery service.
     * `:bootnodes` - a list of bootnodes to use for discovery.
+    * `:fork_digest` - the fork digest to use for gossipsub.
```
mpaulucci, line 73: will update this documentation

lib/lambda_ethereum_consensus/beacon/beacon_node.ex
MegaRedHand: Doesn't `PendingBlocks` make Libp2p requests? Shouldn't it be started later?

lib/lambda_ethereum_consensus/beacon/pending_blocks.ex
MegaRedHand: This wouldn't work:
```elixir
iex(1)> %MapSet{} = MapSet.new([1])
MapSet.new([1])
iex(2)> %MapSet{} = MapSet.new([])
MapSet.new([])
```

[This SO question might have ways to do it](https://stackoverflow.com/questions/49284613/pattern-match-an-empty-mapset)

lib/lambda_ethereum_consensus/p2p/peerbook.ex
MegaRedHand: We should instead match for an empty `peerbook`, then.

```elixir
  @impl true
  def handle_info(:prune, peerbook) when map_size(peerbook) == 0 do
    schedule_pruning()
    {:noreply, peerbook}
  end

  @impl true
  def handle_info(:prune, peerbook) do
    ...
  end
```

lib/libp2p_port.ex
```diff
@@ -38,14 +40,16 @@ defmodule LambdaEthereumConsensus.Libp2pPort do
...
+    bootnodes: [],
+    fork_digest: <<>>
   ]
 
   @type init_arg ::
           {:listen_addr, String.t()}
           | {:enable_discovery, boolean()}
           | {:discovery_addr, String.t()}
           | {:bootnodes, [String.t()]}
+          | {:fork_digest, binary()}
```
MegaRedHand, line 52: Nit:
```suggestion
          | {:fork_digest, String.t()}
```

lib/libp2p_port.ex
```diff
@@ -224,7 +229,20 @@ defmodule LambdaEthereumConsensus.Libp2pPort do
...
+    config = Application.fetch_env!(:lambda_ethereum_consensus, :discovery)
+    port = Keyword.fetch!(config, :port)
+    bootnodes = Keyword.fetch!(config, :bootnodes)
+
+    args = [
+      listen_addr: [],
+      enable_discovery: true,
+      discovery_addr: "0.0.0.0:#{port}",
+      bootnodes: bootnodes,
+      #  BeaconChain.get_fork_digest()
```
MegaRedHand, line 242: We can remove this

Makefile
```diff
@@ -83,7 +83,7 @@ deps:
 proto: $(PROTOBUF_EX_FILES) $(PROTOBUF_GO_FILES)
 
 #🔨 compile-native: @ Compile C and Go artifacts.
-compile-native: $(OUTPUT_DIR)/libp2p_nif.so $(OUTPUT_DIR)/libp2p_port
+compile-native: proto $(OUTPUT_DIR)/libp2p_nif.so $(OUTPUT_DIR)/libp2p_port
```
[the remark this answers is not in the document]
MegaRedHand: Good catch!

MegaRedHand (changes requested):
See https://github.com/lambdaclass/lambda_ethereum_consensus/pull/594#discussion_r1447590615

lib/lambda_ethereum_consensus/beacon/pending_blocks.ex
mpaulucci: wow, good catch! thanks

lib/libp2p_port.ex
mpaulucci: It is a binary though

lib/lambda_ethereum_consensus/beacon/beacon_node.ex
mpaulucci: Done

lib/lambda_ethereum_consensus/beacon/pending_blocks.ex
MegaRedHand: Got bit by it more than one time 😞

lib/lambda_ethereum_consensus/beacon/pending_blocks.ex
```diff
@@ -115,7 +115,8 @@ defmodule LambdaEthereumConsensus.Beacon.PendingBlocks do
   end
 
   @impl true
-  def handle_info(:download_blocks, %{blocks_to_download: %MapSet{}} = state) do
+  def handle_info(:download_blocks, %{blocks_to_download: blocks} = state)
+      when map_size(blocks) == 0 do
```
MegaRedHand, line 119: This doesn't work. I suggest just using a module attribute like:

```elixir
@empty_mapset MapSet.new()

def handle_info(:download_blocks, %{blocks_to_download: @empty_mapset} = state)
```

lib/libp2p_port.ex
```diff
@@ -224,8 +214,18 @@ defmodule LambdaEthereumConsensus.Libp2pPort do
   ########################
 
   @impl GenServer
-  def init(args) do
-    {new_peer_handler, args} = Keyword.pop(args, :new_peer_handler, nil)
+  def init(_args) do
+    config = Application.fetch_env!(:lambda_ethereum_consensus, :discovery)
```
MegaRedHand, line 218: This change makes testing more difficult since it'd need to modify global state. More specifically, the `libp2p_port` unit tests will fail.

test/unit/pending_blocks_test.exs
```diff
@@ -1,7 +1,7 @@
 defmodule Unit.PendingBlocks do
   @moduledoc false
 
-  use ExUnit.Case, async: true
```
mpaulucci, line 4: we shouldn't do async: true when using Patch because it will complain of more than one Patch supervisor running

test/test_helper.exs
```diff
@@ -1,4 +1,5 @@
 ExUnit.start()
+Application.ensure_all_started(:telemetry)
```
MegaRedHand, line 2: 🚀

MegaRedHand (approved):
LGTM

PR #584, opened 2024-01-08
feat: store blocks and states only in DB

MegaRedHand:
Fixes #570

PR #605, opened 2024-01-11
feat: add blocks cache

MegaRedHand:
Depends on #584

### DIFF
--- a/lib/lambda_ethereum_consensus/application.ex
+++ b/lib/lambda_ethereum_consensus/application.ex
@@ -16,6 +16,7 @@ defmodule LambdaEthereumConsensus.Application do
     children = [
       {LambdaEthereumConsensus.Telemetry, []},
       {LambdaEthereumConsensus.Store.Db, []},
+      {LambdaEthereumConsensus.Store.Blocks, []},
       {LambdaEthereumConsensus.Beacon.BeaconNode, [checkpoint_sync]},
       {LambdaEthereumConsensus.P2P.Metadata, []},
       {BeaconApi.Endpoint, []}
--- a/lib/lambda_ethereum_consensus/application.ex
+++ b/lib/lambda_ethereum_consensus/application.ex
@@ -14,12 +14,12 @@ defmodule LambdaEthereumConsensus.Application do
       ]
 
     children = [
-      {LambdaEthereumConsensus.Telemetry, []},
-      {LambdaEthereumConsensus.Store.Db, []},
-      {LambdaEthereumConsensus.Store.Blocks, []},
+      LambdaEthereumConsensus.Telemetry,
+      LambdaEthereumConsensus.Store.Db,
+      LambdaEthereumConsensus.Store.Blocks,
       {LambdaEthereumConsensus.Beacon.BeaconNode, [checkpoint_sync]},
-      {LambdaEthereumConsensus.P2P.Metadata, []},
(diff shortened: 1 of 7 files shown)


lib/lambda_ethereum_consensus/store/blocks.ex
```diff
@@ -0,0 +1,68 @@
...
+  ##########################
+  ### Public API
+  ##########################
+
+  def start_link(_opts) do
+    GenServer.start_link(__MODULE__, [], name: __MODULE__)
+  end
+
+  def store_block(block_root, signed_block) do
+    cache_block(block_root, signed_block)
```
Arkenan, line 18: Let's put this inside the cast for now and make the ets protected. We can think about optimizing writes later on. I think (although we can check) that we should get way more reads than writes.

lib/lambda_ethereum_consensus/store/blocks.ex
```diff
@@ -0,0 +1,68 @@
...
+  ##########################
+  ### Private Functions
+  ##########################
+
+  defp lookup(block_root) do
+    with nil <- :ets.lookup_element(@ets_block_by_hash, block_root, 2, nil),
+         block when not is_nil(block) <- fetch_block(block_root) do
+      cache_block(block_root, block)
+      block
+    end
```
Arkenan, line 53: This comment is optional and subjective, but I think this is one of those cases where the `with` is less clear than a simple case. Maybe something like this is better?

```elixir
case :ets.lookup_element(@ets_block_by_hash, block_root, 2, nil) do
  nil -> cache_miss(block_root)
  block -> block
end
```

Then `cache_miss` itself can be the one that performs both the lookup and add it to the store.

Arkenan (dismissed):
LGTM

lib/lambda_ethereum_consensus/application.ex
```diff
@@ -16,6 +16,7 @@ defmodule LambdaEthereumConsensus.Application do
     children = [
       {LambdaEthereumConsensus.Telemetry, []},
       {LambdaEthereumConsensus.Store.Db, []},
+      {LambdaEthereumConsensus.Store.Blocks, []},
```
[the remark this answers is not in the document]
MegaRedHand: Yes

lib/lambda_ethereum_consensus/store/blocks.ex
```diff
@@ -0,0 +1,68 @@
...
+  @spec clear() :: any()
+  def clear, do: :ets.delete_all_objects(@ets_block_by_hash)
+
+  ##########################
+  ### GenServer Callbacks
+  ##########################
+
+  @impl GenServer
+  def init(_) do
+    :ets.new(@ets_block_by_hash, [:set, :public, :named_table])
```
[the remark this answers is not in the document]
MegaRedHand: I'm using `public` because I'm allowing anyone to write to the cache

lib/lambda_ethereum_consensus/store/blocks.ex
Arkenan: Discussed offline.

lib/lambda_ethereum_consensus/store/blocks.ex
MegaRedHand: Done

Arkenan (approved):
LGTM

PR #607, opened 2024-01-11
fix: set used config before running test

MegaRedHand:
Sometimes this test would fail because some other test put the `MinimalConfig` in the config.
Here's an example: https://github.com/lambdaclass/lambda_ethereum_consensus/actions/runs/7488525574/job/20383206386

### DIFF
--- a/test/unit/pending_blocks_test.exs
+++ b/test/unit/pending_blocks_test.exs
@@ -9,6 +9,8 @@ defmodule Unit.PendingBlocks do
   alias LambdaEthereumConsensus.Store.BlockStore
 
   setup do
+    Application.put_env(:lambda_ethereum_consensus, ChainSpec, config: MainnetConfig)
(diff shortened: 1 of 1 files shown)


test/unit/pending_blocks_test.exs
```diff
@@ -9,6 +9,8 @@ defmodule Unit.PendingBlocks do
   alias LambdaEthereumConsensus.Store.BlockStore
 
   setup do
+    Application.put_env(:lambda_ethereum_consensus, ChainSpec, config: MainnetConfig)
```
mpaulucci, line 12: why does this test fail with minimal config?

MegaRedHand: Because it uses a block with mainnet-sized values

mpaulucci (approved):

PR #609, opened 2024-01-11
feat: add some fork choice values to BeaconChain.

mpaulucci:
We don't want to have to calculate the new head everytime it is requested since it's expensive. Let's store a "cache" (better names welcomed) that gets updated every time we get a new block.


lib/lambda_ethereum_consensus/beacon/beacon_chain.ex
```diff
@@ -13,13 +13,20 @@ defmodule LambdaEthereumConsensus.Beacon.BeaconChain do
...
+      :fork_choice
     ]
 
     @type t :: %__MODULE__{
             genesis_time: Types.uint64(),
             genesis_validators_root: Types.bytes32(),
-            time: Types.uint64()
+            time: Types.uint64(),
+            fork_choice: %{
+              head_root: Types.root(),
```
mpaulucci, line 25: could prepend all these values with `last`, i.e `last_finalized_root`

MegaRedHand (commented):
LGTM. I'll let @Arkenan have a final look.

lib/lambda_ethereum_consensus/fork_choice/fork_choice.ex
```diff
@@ -70,7 +72,7 @@ defmodule LambdaEthereumConsensus.ForkChoice do
   @spec init({BeaconState.t(), SignedBeaconBlock.t(), Types.uint64()}) ::
           {:ok, Store.t()} | {:stop, any}
   def init({anchor_state = %BeaconState{}, signed_anchor_block = %SignedBeaconBlock{}, time}) do
-    case Helpers.get_forkchoice_store(anchor_state, signed_anchor_block, true) do
+    case Helpers.get_forkchoice_store(anchor_state, signed_anchor_block, false) do
```
mpaulucci, line 75: I've had to turn off db because otherwise the `get_head` function fails.

MegaRedHand (approved):
LGTM

lib/lambda_ethereum_consensus/fork_choice/fork_choice.ex
MegaRedHand: I'll take a look

lib/lambda_ethereum_consensus/fork_choice/fork_choice.ex
```diff
@@ -158,6 +162,11 @@ defmodule LambdaEthereumConsensus.ForkChoice do
     {:noreply, new_store}
   end
 
+  @impl GenServer
+  def handle_info(_msg, state) do
```
Arkenan, line 166: Instead of calling a `cast` to update the genserver, we can match the message against `{_ref, new_fork_choice_cache}` and receive the task result directly.

lib/lambda_ethereum_consensus/fork_choice/fork_choice.ex
```diff
@@ -194,4 +203,33 @@ defmodule LambdaEthereumConsensus.ForkChoice do
...
+
+    # TODO: do someting with the result from the execution client
+    # TODO: compute safe block hash
+    ExecutionClient.notify_forkchoice_updated(
+      head_execution_hash,
+      finalized_execution_hash,
+      finalized_execution_hash
+    )
+
+    BeaconChain.update_fork_choice_cache(
```
Arkenan, line 226: We can return a map or a tuple with this and it can be caught by handle_info instead of calling a cast.

Arkenan (approved):
LGTM! I suggested a minor change, but otherwise, feel free to merge.

Arkenan (approved):
LGTM after re-review and offline discussion.

PR #610, opened 2024-01-11
fix: avoid hard match in `BeaconNode.init`

MegaRedHand:
Although a really rare case, when fetching the latest state without a block for the same slot, the node would raise a match error instead of cleanly exiting.


mpaulucci (approved):
nice!

PR #611, opened 2024-01-11
feat: switchable store implementation PoC

MegaRedHand:
Depends on #584

This PR adds a PoC for switching the underlying storage for the `Store`


MegaRedHand:
Continued in #676

PR #612, opened 2024-01-11
fix: new endpoints were failing some requests

MegaRedHand:
A recently merged PR had some errors not caught in code review. We should probably add some tests to catch these kinds of errors.

### DIFF
--- a/lib/beacon_api/controllers/v1/beacon_controller.ex
+++ b/lib/beacon_api/controllers/v1/beacon_controller.ex
@@ -6,12 +6,11 @@ defmodule BeaconApi.V1.BeaconController do
 
   @spec get_state_root(Plug.Conn.t(), any) :: Plug.Conn.t()
   def get_state_root(conn, %{"state_id" => state_id}) do
-    case BeaconApi.Utils.parse_id(state_id) |> ForkChoice.Helpers.root_by_id() do
-      {:ok, {root, execution_optimistic, finalized}} ->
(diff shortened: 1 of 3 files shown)


lib/beacon_api/controllers/v1/beacon_controller.ex
```diff
@@ -6,12 +6,11 @@ defmodule BeaconApi.V1.BeaconController do
...
   def get_state_root(conn, %{"state_id" => state_id}) do
-    case BeaconApi.Utils.parse_id(state_id) |> ForkChoice.Helpers.root_by_id() do
-      {:ok, {root, execution_optimistic, finalized}} ->
-        with {:ok, state_root} <- ForkChoice.Helpers.get_state_root(root) do
-          conn |> root_response(state_root, execution_optimistic, finalized)
-        end
-
+    with {:ok, {root, execution_optimistic, finalized}} <-
+           BeaconApi.Utils.parse_id(state_id) |> ForkChoice.Helpers.root_by_id(),
+         {:ok, state_root} <- ForkChoice.Helpers.get_state_root(root) do
```
mpaulucci, line 11: we should try to avoid getters to ForkChoice, instead using BeaconChain

mpaulucci (approved):

lib/beacon_api/controllers/v1/beacon_controller.ex
MegaRedHand: This doesn't access the `ForkChoice` `GenServer`. It just uses some helper methods, although we probably should change the module's name 😅

PR #616, opened 2024-01-11
feat: mock execution client

### LINKED ISSUE #579: Add stubbed execution client
When in development, we'd like to be able to prototype some things quickly, so it'd be ideal to have a simple execution client embedded in our node. This functionality would be enabled by some flag, like `--mock-execution`

karasakalmt:
Closes #579


lib/lambda_ethereum_consensus/execution/engine_api/tesla.ex
```diff
@@ -0,0 +1,51 @@
+defmodule LambdaEthereumConsensus.Execution.EngineApi.Tesla do
```
mpaulucci, line 1: I would call it something other than Tesla, maybe `Api` or `Rcp`

mpaulucci (approved):

PR #617, opened 2024-01-12
feat: implement bit list utils module

### LINKED ISSUE #512: Bitlist utils
We should add a module with utils for manipulating bitstring-encoded bitlists.

Candidate functions for this module are:

https://github.com/lambdaclass/lambda_ethereum_consensus/blob/ae5c2a0195b9c285dfe6ab346e1cd2d944aebe2e/lib/lambda_ethereum_consensus/state_transition/operations.ex#L869-L873

f3r10:
**Motivation**

Closes #512


lib/utils/bit_list.ex
```diff
@@ -0,0 +1,92 @@
...
+  @spec new(bitstring) :: t
+  def new(bitstring) when is_bitstring(bitstring) do
+    # Change the byte order from little endian to big endian (reverse bytes).
+    num_bits = bit_size(bitstring)
+    len = length_of_bitlist(bitstring)
+
+    <<pre::integer-little-size(num_bits - 8), last_byte::integer-little-size(@bits_per_byte)>> =
+      bitstring
+
+    decoded = <<pre::integer-size(num_bits - 8), remove_trailing_bit(<<last_byte>>)::bitstring>>
```
Arkenan, line 20: This doesn't look right, the last byte should end up as the first one after removing the trailing bit. To test for this, let's add some e2e (encode-decode) tests with a bitlist of size 10.

lib/utils/bit_list.ex
```diff
@@ -0,0 +1,92 @@
...
+
+  @doc """
+  Sets a bit (turns it to 1).
+  Equivalent to bit_list[index] = 1.
+  """
+  @spec set(t, non_neg_integer) :: t
+  def set({bit_list, len}, index) do
+    skip = bit_size(bit_list) - index - 1
+    <<pre::bitstring-size(skip), _::size(1), rest::bitstring>> = bit_list
+    {<<pre::bitstring, 1::1, rest::bitstring>>, len}
```
Arkenan, line 52: These are all repeated from `BitVector`. Can we extract those to a single library called BitField and call these from both BitVector and BitList?

Arkenan (changes requested):
Thanks for the PR! Requested some changes

Arkenan (dismissed):
LGTM with some changes. Feel free to apply them and merge afterwards.

Arkenan (approved):
LGTM

PR #622, opened 2024-01-12
fix: metadata genserver not returning state

MegaRedHand:
Two bugs:
- 7b746f0ebc8943c75e3353fd4642b36429028532: `GenServer.handle_call` not returning state
- 85e3e161d8354c2eaeb93f78a0078206e31447e5: some setters (unused, so no problem) used `GenServer.call` instead of `GenServer.cast`, while the handlers used `handle_call`


Arkenan (approved):
LGTM, thanks for the fix

PR #628, opened 2024-01-15
fix: use pattern-matching instead of `String` when encoding `u256`s

### LINKED ISSUE #627: Pending blocks crash, probably ssz encode issue
We had this error on saturday, which made the `PendingBlocks` GenServer crash:

```
05:43:48.298 [error] GenServer LambdaEthereumConsensus.Beacon.PendingBlocks terminating
** (ArgumentError) argument error
    (erts 14.1) :erlang.++([169 | <<172, 16>>], [<<172, 16>>])
    (stdlib 5.1) unicode_util.erl:3233: :unicode_util.gc_ext_pict/3
    (elixir 1.15.6) lib/string.ex:2058: String.length/2
    (elixir 1.15.6) lib/string.ex:1378: String.pad/4
    (elixir 1.15.6) lib/map.ex:916: Map.update!/3
    (lambda_ethereum_consensus 0.1.0) lib/ssz.ex:133: anonymous fn/1 in Ssz.encode/1
    (elixir 1.15.6) lib/enum.ex:1701: anonymous fn/3 in Enum.map/2
    (stdlib 5.1) maps.erl:416: :maps.fold_1/4
```

Tracing down the ssz encode function, it looks like this section of the code:

```elixir
roots_to_remove =
  downloaded_blocks
  |> Enum.map(&Ssz.hash_tree_root!(&1.message))
  |> MapSet.new()
  |> MapSet.union(blocks_in_store)
```

This performs, in order:
- encoding of a struct.
- encoding of a map
- it goes through each field, and encodes it.

Inside of the SSZ encoding, it looks like there's a problem with padding. This is the code section in charge:

```elixir
  @spec encode_u256(non_neg_integer) :: binary
  def encode_u256(num) do
    num
    |> :binary.encode_unsigned(:little)
    |> String.pad_trailing(32, <<0>>)
  end
```

MegaRedHand:
Fixes #627

### DIFF
--- a/lib/ssz.ex
+++ b/lib/ssz.ex
@@ -165,7 +165,7 @@ defmodule Ssz do
   end
 
   @spec encode_u256(Types.uint256()) :: Types.bytes32()
-  def encode_u256(num) when num < @max_u256, do: <<num::little-size(256)>>
+  def encode_u256(num) when num <= @max_u256, do: <<num::little-size(256)>>
 
   @spec decode_u256(Types.bytes32()) :: Types.uint256()
   def decode_u256(<<num::little-size(256)>>), do: num
--- a/lib/ssz.ex
+++ b/lib/ssz.ex
@@ -164,9 +164,9 @@ defmodule Ssz do
     function_exported?(module, function, arity)
   end
 
-  @spec encode_u256(non_neg_integer) :: binary
+  @spec encode_u256(Types.uint256()) :: Types.bytes32()
   def encode_u256(num) when num < @max_u256, do: <<num::little-size(256)>>
 
-  @spec decode_u256(binary) :: non_neg_integer
+  @spec decode_u256(Types.bytes32()) :: Types.uint256()
   def decode_u256(<<num::little-size(256)>>), do: num
 end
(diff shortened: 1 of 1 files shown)


lib/ssz.ex
```diff
@@ -162,13 +164,9 @@ defmodule Ssz do
...
   end
 
-  @spec encode_u256(non_neg_integer) :: binary
-  def encode_u256(num) do
-    num
-    |> :binary.encode_unsigned(:little)
-    |> String.pad_trailing(32, <<0>>)
-  end
+  @spec encode_u256(Types.uint256()) :: Types.bytes32()
+  def encode_u256(num) when num < @max_u256, do: <<num::little-size(256)>>
```
Arkenan, line 168: You can use `<=` if you already did the `-1`. If you still want to use the `<`, then you can remove the `-1` from above

Arkenan (dismissed):
Boundary issue, LGTM otherwise

lib/ssz.ex
MegaRedHand: Good catch!

Arkenan (approved):
LGTM now

PR #630, opened 2024-01-15
fix: fix faulty MapSet pattern matching.

mpaulucci:
That is supposed to match an empty mapset but it actually matches any mapset.


MegaRedHand (approved):
🚀

PR #635, opened 2024-01-16
feat: add open api spex beacon endpoints

### LINKED ISSUE #165: Add OpenApiSpex to the phoenix endpoints

f3r10:
**Motivation**

Closes #165


lib/beacon_api/controllers/v1/beacon_controller.ex
```diff
@@ -1,8 +1,37 @@
...
+    StateNotFoundStateRootResponse,
+    SuccessStateRootResponse
+  }
+
   alias LambdaEthereumConsensus.ForkChoice
   alias LambdaEthereumConsensus.Store.BlockStore
   use BeaconApi, :controller
+  use OpenApiSpex.ControllerSpecs
+
+  operation(:get_state_root,
```
mpaulucci, line 16: There is already a spec provided by the ethereum foundation: https://ethereum.github.io/beacon-APIs/ . We should use this as our spec

lib/beacon_api/api_spec.ex
```diff
@@ -0,0 +1,13 @@
+defmodule BeaconApi.ApiSpec do
+  @moduledoc false
+  alias OpenApiSpex.OpenApi
+  @behaviour OpenApi
+
+  @ethspec "beacon-node-oapi.json"
```
mpaulucci, line 6: nice! But it would be nice to do like we do with spec tests, we have a file with the version (see https://github.com/lambdaclass/lambda_ethereum_consensus/blob/main/.spectest_version) and a operation in the makefile to download the current version from github releases

MegaRedHand: Also, we should throw an [`@external_resource <file>` tag](https://hexdocs.pm/elixir/1.14.5/Module.html#module-external_resource) for good measure.

Makefile
```diff
@@ -126,6 +126,17 @@ sepolia: compile-all
...
 	mix test --no-start --exclude spectest
 
+#### BEACON NODE OAPI ####
+OAPI_NAME = beacon-node-oapi
+OAPI_VERSION := $(shell cat .oapi_version)
+$(OAPI_NAME).json:
+	curl -L -o "$@" \
+		"https://ethereum.github.io/beacon-APIs/releases/${OAPI_VERSION}/beacon-node-oapi.json"
+
+OPENAPI_JSON := $(OAPI_NAME).json
```
mpaulucci, line 136: maybe we should add this as a dependency to `compile-all`. @MegaRedHand what do you think?

Makefile
```diff
@@ -126,6 +126,17 @@ sepolia: compile-all
 test: compile-all
 	mix test --no-start --exclude spectest
 
+#### BEACON NODE OAPI ####
+OAPI_NAME = beacon-node-oapi
+OAPI_VERSION := $(shell cat .oapi_version)
+$(OAPI_NAME).json:
```
MegaRedHand, line 132: We should make this depend on `.oapi_version`:
```suggestion
$(OAPI_NAME).json: .oapi_version
```

That way changes in the version cause `make` to update the json. Note that this wasn't needed for spec-tests due to them having their version appended to the file name.

Makefile
MegaRedHand: Yes, probably. The project wouldn't compile without it. Maybe we'll also have to update the CI 🤔

f3r10: yes, it is necessary to update the CI :sweat_smile: otherwise the build process fails.
I am going to try to update that part

.github/workflows/ci.yml
```diff
@@ -51,9 +51,25 @@ jobs:
         if: steps.output-cache.outputs.cache-hit != 'true'
         run: make compile-port compile-native
 
+  download-beacon-node-oapi:
+    name: Download Beacon Node OAPI
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+      - name: Cache compressed spectests
```
MegaRedHand, line 59:
```suggestion
      - name: Cache Beacon Node OAPI
```

lib/beacon_api/controllers/v1/beacon_controller.ex
```diff
@@ -1,11 +1,22 @@
...
+
   alias LambdaEthereumConsensus.ForkChoice
   alias LambdaEthereumConsensus.Store.BlockStore
   use BeaconApi, :controller
 
+  plug(OpenApiSpex.Plug.CastAndValidate, json_render_error_v2: true)
+
+  def open_api_operation(action) do
+    apply(__MODULE__, :"#{action}_operation", [])
+  end
```
MegaRedHand, line 13: What's `action`? Is it an atom? Is there a bounded amount of `action`s?

MegaRedHand (commented):
The CI additions look good 🚀. Left a question on the `OpenApiSpex` integration, but otherwise LGTM

lib/beacon_api/controllers/v1/beacon_controller.ex
f3r10: an `action` is an atom that is the MVC controller action's function name. In the Beacon API case, the actions are  `:get_state_root`, `:get_block_root`,  `:get_block`

https://github.com/open-api-spex/open_api_spex/blob/master/examples/phoenix_app/lib/phoenix_app_web/controllers/user_controller_with_struct_specs.ex#L17

mpaulucci (approved):
LGTM!

PR #639, opened 2024-01-16
feat: add loki to docker-compose

### LINKED ISSUE #633: [Observability] Add Loki to aggregate and visualize logs

MegaRedHand:
Closes #633

This PR adds Loki and Promtail containers to our docker-compose file. It also removes the login page on Grafana.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -4,15 +4,6 @@ import Config
 # Configure logging
 config :logger, level: :info, truncate: :infinity
 
-# Configures the phoenix endpoint
-config :lambda_ethereum_consensus, BeaconApi.Endpoint,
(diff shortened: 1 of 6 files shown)


config/config.exs
```diff
@@ -4,6 +4,18 @@ import Config
 # Configure logging
 config :logger, level: :info, truncate: :infinity
 
+## TODO: we might want to enable this with a CLI flag
```
mpaulucci, line 7: if we might enable this with a cli flag then it should probably be in runtime.ex

MegaRedHand: Yes. I'm leaving that for another PR since it's not necessary unless deploying with Grafana.

PR #643, opened 2024-01-16
refactor: use templates for configs and presets

MegaRedHand:
Depends on #642

This PR removes duplicate code in configs and presets.

### DIFF
--- a/lib/chain_spec/configs/mainnet.ex
+++ b/lib/chain_spec/configs/mainnet.ex
@@ -2,12 +2,5 @@ defmodule MainnetConfig do
   @moduledoc """
   Mainnet config constants.
   """
-  file = "config/networks/mainnet/config.yaml"
-
-  @external_resource file
-
-  @parsed_config ConfigUtils.load_config_from_file!(file)
-  @unified Map.merge(MainnetPreset.get_preset(), @parsed_config)
-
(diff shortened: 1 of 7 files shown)


lib/chain_spec/configs/mainnet.ex
```diff
@@ -2,9 +2,5 @@ defmodule MainnetConfig do
   @moduledoc """
   Mainnet config constants.
   """
-
-  @parsed_config ConfigUtils.load_config_from_file!("config/networks/mainnet/config.yaml")
-  @unified Map.merge(MainnetPreset.get_preset(), @parsed_config)
-
-  def get(key), do: Map.fetch!(@unified, key)
+  use ChainSpec.GenConfig, file: "config/networks/mainnet/config.yaml", preset: MainnetPreset
```
mpaulucci, line 5: The preset is a value in the config, can't we fetch it from config and parse it instead of having to select it here?

mpaulucci:
🎉

lib/chain_spec/configs/mainnet.ex
MegaRedHand: Yeah, we can

MegaRedHand: I added it in 1d6063b

mpaulucci (approved):

PR #644, opened 2024-01-17
feat: optimization on merkleization of chunks by precomputed hashes

### LINKED ISSUE #597: [SSZ] Optimization on merkleization of chunks by precomputed hashes.
## Description

Our [current implementation](https://github.com/lambdaclass/lambda_ethereum_consensus/pull/528) of merkleization of chunks needs some optimization and one way of doing that is by using precomputed hashes of zeros.

During merkleization, chunks are padded by zeros to reach the next value that's a power of 2. And, merkleization of zeros are known so basically we hard code these hashes and use them rather than computing them everytime. 

For example, the hash of:
- two zeros of 32 bytes is `f5a5fd42d16a20302798ef6ed309979b43003d2320d9f0e8ea9831a92759fb4b`
- four zeros of 32 bytes is `db56114e00fdd4c1f85c892bf35ac9a89289aaecb1ebd0a96cde606a748b5d71`

## references

- https://github.com/gballet/ssz.zig/blob/master/src/zeros.zig

Godspower-Eze:
## Outcome

```
##### With input args #####
Name                                                  ips        average  deviation         median         99th %
SszEx.merkleize_chunks_with_virtual_padding       73.71 K       13.57 μs    ±49.03%       12.42 μs       44.17 μs
SszEx.merkleize_chunks                            20.48 K       48.83 μs    ±35.14%       49.63 μs       65.63 μs

Comparison: 
SszEx.merkleize_chunks_with_virtual_padding       73.71 K
SszEx.merkleize_chunks                            20.48 K - 3.60x slower +35.26 μs
```

Closes #597


lib/ssz_ex.ex
```diff
@@ -15,6 +15,123 @@ defmodule LambdaEthereumConsensus.SszEx do
   @bits_per_byte 8
   @bits_per_chunk @bytes_per_chunk * @bits_per_byte
   @zero_chunk <<0::size(@bits_per_chunk)>>
+  @zero_hashes <<0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
```
MegaRedHand, line 18: Can we compute this at compile-time? That way it's clear what this is, and how it can be extended if needed.

lib/ssz_ex.ex
```diff
@@ -184,6 +333,16 @@ defmodule LambdaEthereumConsensus.SszEx do
...
 
+  def chunk_count({:list, {:int, size}, max_size}) do
+    size = size_of({:int, size})
+    (max_size * size + 31) |> div(32)
+  end
+
+  def chunk_count({:list, :bool, max_size}) do
+    size = size_of(:bool)
+    (max_size * size + 31) |> div(32)
+  end
```
MegaRedHand, line 344: Both of these seem to be doing the same thing:

```elixir
  def chunk_count({:list, type, max_size}) do
    size = size_of(type)
    (max_size * size + 31) |> div(32)
  end
```

lib/ssz_ex.ex
```diff
@@ -634,11 +783,16 @@ defmodule LambdaEthereumConsensus.SszEx do
...
-      0
-    else
-      n = ((len <<< 1) - 1) |> :math.log2() |> trunc()
-      2 ** n
+    cond do
+      len == 0 ->
+        0
+
+      len == 1 ->
+        1
```
MegaRedHand, line 791: These can be moved to the function signature. Also, I think the `len == 1` one isn't needed, since `(1 << 1) - 1 == 1` and `2 ** log(1) == 1`

lib/ssz_ex.ex
```diff
@@ -634,11 +783,16 @@ defmodule LambdaEthereumConsensus.SszEx do
...
+
+      true ->
+        n = ((len <<< 1) - 1) |> :math.log2() |> trunc()
```
MegaRedHand, line 794: Should we reuse `compute_pow` here?

lib/ssz_ex.ex
```diff
@@ -652,4 +806,69 @@ defmodule LambdaEthereumConsensus.SszEx do
...
+      parent_index = j |> div(2)
+      nodes = get_nodes(parent_index, i, j, height, acc_layers, acc_last_index)
+      hash_nodes_and_replace(nodes, acc_layers)
+    end)
+  end
+
+  defp get_nodes(parent_index, i, j, height, acc_layers, acc_last_index) do
+    nodes =
+      cond do
+        j < acc_last_index ->
```
MegaRedHand, line 833: Same here, these conditions can be moved to the function signature:
```elixir
defp get_nodes(parent_index, i, j, height, acc_layers, acc_last_index) when j < acc_last_index do ...
defp get_nodes(parent_index, i, j, height, acc_layers, acc_last_index) when j == acc_last_index do ...
defp get_nodes(_, _, _, _, _, _), do: :error

**Review by MegaRedHand — COMMENTED** [review]
Really good work!

File: lib/ssz_ex.ex (hunk 1 above)
Comment (Godspower-Eze) [reply lib/ssz_ex.ex:18]: How can I go about this?

Comment (MegaRedHand) [reply lib/ssz_ex.ex:18]: First of all, how did you get this string? Just to clarify, "compile-time" in Elixir is simple: just save it in a module attribute (`@some_name`); so doing something like `@zero_hashes compute_zero_hashes(...)` is enough.

File: lib/ssz_ex.ex [hunk 6]
```diff
@@ -784,18 +779,11 @@ defmodule LambdaEthereumConsensus.SszEx do
...
   end
 
-  defp next_pow_of_two(len) when is_integer(len) and len >= 0 do
-    cond do
-      len == 0 ->
-        0
-
-      len == 1 ->
-        1
+  defp next_pow_of_two(len) when len == 0, do: 0
```
MegaRedHand, line 782: Nit: this can be simplified to:
```elixir
defp next_pow_of_two(0), do: 0
```

lib/ssz_ex.ex
Godspower-Eze: Here: https://github.com/ralexstokes/ssz-rs/blob/1f94d5dfc70c86dab672e91ac46af04a5f96c342/ssz-rs/src/merkleization/mod.rs#L144

It's generated at compile time here.

Made an update. Compilation is slower rn though

MegaRedHand: This seems to be where it's computed:
https://github.com/ralexstokes/ssz-rs/blob/1f94d5dfc70c86dab672e91ac46af04a5f96c342/ssz-rs/build.rs#L14-L24

Let's try to add that to the `ZeroHashes` module. If the compilation turns out to be too slow, we can just go back to the pre-computed hashes, and leave the way to compute them as a comment.

Godspower-Eze: That function is only called once when the project is built. And, the point of this is so that we don't need to do the hashing again and that what the function does.

See another example:

https://github.com/gballet/ssz.zig/blob/master/src/zeros.zig

MegaRedHand: Yes. I'm just saying it might be useful as documentation. Both of those implementations have a function to compute those hashes. If we add that, future maintainers will more easily know what this magic string is. Also, just to clarify: as long as the module doesn't use any external functions, it won't need to be recompiled, so the code will run only the first time it's compiled (the same way `ssz-rs`'s implementation works).

Godspower-Eze: Gotcha.

Godspower-Eze: Added the function

lib/utils/zero_hashes.ex
```diff
@@ -0,0 +1,149 @@
...
+  @moduledoc """
+  Precomputed zero hashes
+  """
+
+  alias LambdaEthereumConsensus.SszEx
+
+  @bits_per_byte 8
+  @bytes_per_chunk 32
+  @max_merkle_tree_depth 64
+  @zero_hashes <<0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
```
mpaulucci, line 11:
```suggestion
    @buffer <<0::size(@bytes_per_chunk * @max_merkle_tree_depth * @bits_per_byte)>>

  @zero_hashes 0..(@max_merkle_tree_depth - 2)
               |> Enum.reduce(@buffer, fn index, acc_buffer ->
                 start = index * @bytes_per_chunk
                 stop = (index + 2) * @bytes_per_chunk
                 focus = acc_buffer |> :binary.part(start, stop - start)
                 left = focus |> :binary.part(0, @bytes_per_chunk)
                 hash = SszEx.hash_nodes(left, left)
                 change_index = (index + 1) * @bytes_per_chunk
                 SszEx.replace_chunk(acc_buffer, change_index, hash)
               end)
```

This way you can compute it at compile time

test/unit/ssz_ex_test.exs
```diff
@@ -145,6 +145,78 @@ defmodule Unit.SSZExTest do
...
+    assert root |> Base.encode16(case: :lower) == expected_value
+
+    ## Large Leaf Count
+
+    chunks = ones <> ones <> ones <> ones <> ones
+    root = chunks |> SszEx.merkleize_chunks_with_virtual_padding(2 ** 10)
+    expected_value = "2647cb9e26bd83eeb0982814b2ac4d6cc4a65d0d98637f1a73a4c06d3db0e6ce"
+    assert root |> Base.encode16(case: :lower) == expected_value
+
+    ## TOO HEAVY COMPUTATION!
```
mpaulucci, line 212: should we remove this code then?

lib/utils/zero_hashes.ex
Godspower-Eze: Isn't it faster to have as a constant than to compute this at compile time?

I added the function so @zero_hashes doesn't seem like a magic value.

mpaulucci: What matters is that it will be equally fast in runtime. I think it's much clearer to have it like this over a hardcoded value. The difference in compile time should be unnoticeable

Godspower-Eze: Okay. Updated it

PR #648, opened 2024-01-17
feat: add `Store.checkpoint_states` pruning

MegaRedHand:
Storing blocks and states in the DB reduces the amount of data we store in memory. However, we still have some other `Store` attributes that could grow indefinitely: `checkpoint_states`, `unrealized_justifications`, and `equivocating_indices`. This PR prunes `checkpoint_states` on each block, to address this problem.


lib/lambda_ethereum_consensus/fork_choice/fork_choice.ex
```diff
@@ -191,7 +191,7 @@ defmodule LambdaEthereumConsensus.ForkChoice do
          {:ok, new_store} <-
            signed_block.message.body.attester_slashings
            |> apply_handler(new_store, &Handlers.on_attester_slashing/2) do
-      {:ok, new_store}
+      {:ok, Handlers.prune_checkpoint_states(new_store)}
```
mpaulucci, line 194: shouldn't we do this at the end of an epoch instead of on every block?

MegaRedHand: I did it this way for simplicity. The map is actually _really_ small in size (~3 key-value pairs in the common case), and it doesn't impact performance.

PR #650, opened 2024-01-17
feat: added state_root -> state option to db

### LINKED ISSUE #637: Allow going from `state_root` to `state` in the DB
For some endpoints we need to fetch the state from the state's root. This is currently not possible, because we index the states by the corresponding block's root.

karasakalmt:
Closes #637


karasakalmt:
@MegaRedHand can you check the pr as draft

lib/lambda_ethereum_consensus/store/state_store.ex
```diff
@@ -6,61 +6,80 @@ defmodule LambdaEthereumConsensus.Store.StateStore do
...
   def store_state(%BeaconState{} = state) do
     # NOTE: due to how SSZ-hashing works, hash(block) == hash(header)
     block_root = Ssz.hash_tree_root!(state.latest_block_header)
+    state_root = Ssz.hash_tree_root!(state)
     {:ok, encoded_state} = Ssz.to_ssz(state)
 
-    key = state_key(block_root)
-    Db.put(key, encoded_state)
+    key_block = state_key(block_root, :block_root)
+    key_state = state_key(state_root, :state_root)
```
MegaRedHand, line 22: States are really heavy, so ideally we'd have them on one place only. I auggest just leaving the old block_hash -> state mapping (that way we don't impact block processing), and just add a state_hash -> block_hash mapping. Then when fetching the state by state hash we can just do state_hash -> block_hash -> state

lib/lambda_ethereum_consensus/store/state_store.ex
```diff
@@ -6,61 +6,80 @@ defmodule LambdaEthereumConsensus.Store.StateStore do
...
+    Db.put(key_block, encoded_state)
+    Db.put(key_state, encoded_state)
```
MegaRedHand, line 24: Do we need this? With slot -> block_hash and block_hash -> state mappings can't we already do slot -> state ? Also, I think we can just reuse the `BlockStates` function that goes from slot -> block_hash.

lib/lambda_ethereum_consensus/store/state_store.ex
```diff
@@ -6,61 +6,80 @@ defmodule LambdaEthereumConsensus.Store.StateStore do
...
+      get_state(root, root_type)
     end
   end
 
-  defp state_key(root), do: Utils.get_key(@state_prefix, root)
-  defp root_by_slot_key(slot), do: Utils.get_key(@stateslot_prefix, slot)
+  defp state_key(root, :block_root), do: Utils.get_key(@state_prefix_by_block, root)
+  defp state_key(root, :state_root), do: Utils.get_key(@state_prefix_by_state, root)
+  defp root_by_slot_key(slot, :block_root), do: Utils.get_key(@stateslot_prefix_by_block, slot)
+  defp root_by_slot_key(slot, :state_root), do: Utils.get_key(@stateslot_prefix_by_state, slot)
```
MegaRedHand, line 84: It'd be simpler to just have four different functions.

lib/lambda_ethereum_consensus/store/state_store.ex
```diff
@@ -6,61 +6,80 @@ defmodule LambdaEthereumConsensus.Store.StateStore do
...
-  def get_state_root_by_slot(slot),
-    do: slot |> root_by_slot_key() |> Db.get()
+  def get_state_root_by_slot(slot, root_type),
+    do: slot |> root_by_slot_key(root_type) |> Db.get()
 
-  @spec get_state_by_slot(Types.slot()) ::
+  @spec get_state_by_slot(Types.slot(), :block_root | :state_root) ::
           {:ok, BeaconState.t()} | {:error, String.t()} | :not_found
-  def get_state_by_slot(slot) do
+  def get_state_by_slot(slot, root_type) do
```
MegaRedHand, line 74: The result shouldn't change between root types

lib/lambda_ethereum_consensus/store/state_store.ex
karasakalmt: it will led to case .. do's in the other functions body but if you think so I can change it initially I have done it that way but things got messy

lib/lambda_ethereum_consensus/store/state_store.ex
karasakalmt: I was thinking about the same thing actually I will remove added it just to confirm

lib/lambda_ethereum_consensus/store/state_store.ex
```diff
@@ -7,43 +7,54 @@ defmodule LambdaEthereumConsensus.Store.StateStore do
...
+    block_root = Ssz.hash_tree_root!(state.latest_block_header)
+    state_root = Ssz.hash_tree_root!(state)
     {:ok, encoded_state} = Ssz.to_ssz(state)
 
-    key = state_key(block_root)
-    Db.put(key, encoded_state)
+    key_block = state_key(block_root)
+    key_state = block_key(state_root)
+    Db.put(key_block, encoded_state)
+    Db.put(key_state, key_block)
```
MegaRedHand, line 23: We should just store the `block_root` here. That saves on space, and we can go from `block_root` to `block_key` anyways

lib/lambda_ethereum_consensus/store/state_store.ex
```diff
@@ -7,43 +7,54 @@ defmodule LambdaEthereumConsensus.Store.StateStore do
...
+    end
   end
 
-  @spec get_state(Types.root()) ::
+  @spec get_state(Types.root(), :block_root | :state_root) ::
           {:ok, BeaconState.t()} | {:error, String.t()} | :not_found
-  def get_state(root) do
-    with {:ok, bin} <- root |> state_key() |> Db.get() do
+  def get_state(root, root_type) do
+    with {:ok, bin} <- root |> get_block_root(root_type) |> state_key() |> Db.get() do
```
MegaRedHand, line 43: For API simplicity, we should have two functions for this: one for `:block_root` and another for `:state_root` and leave this function private.

lib/lambda_ethereum_consensus/store/state_store.ex
```diff
@@ -7,43 +7,54 @@ defmodule LambdaEthereumConsensus.Store.StateStore do
...
+
+  defp get_block_root(root, type) do
+    if type == :state_root do
+      with {:ok, bin} <- root |> block_key() |> Db.get() do
+        bin
+      end
+    else
+      root
+    end
   end
```
MegaRedHand, line 38: Nit: you can match the type in the function signature

lib/lambda_ethereum_consensus/store/state_store.ex
```diff
@@ -7,43 +7,54 @@ defmodule LambdaEthereumConsensus.Store.StateStore do
...
+    with {:ok, bin} <- root |> get_block_root(root_type) |> state_key() |> Db.get() do
       Ssz.from_ssz(bin, BeaconState)
     end
   end
 
-  @spec get_latest_state() ::
+  @spec get_latest_state(:block_root | :state_root) ::
           {:ok, BeaconState.t()} | {:error, String.t()} | :not_found
-  def get_latest_state do
+  def get_latest_state(root_type) do
```
MegaRedHand, line 50: Both `:block_root` and `:state_root` should be returning the same, so having a `root_type` parameter isn't needed.

lib/lambda_ethereum_consensus/store/state_store.ex
```diff
@@ -7,43 +7,54 @@ defmodule LambdaEthereumConsensus.Store.StateStore do
...
     last_key = root_by_slot_key(0xFFFFFFFFFFFFFFFF)
 
     with {:ok, it} <- Db.iterate(),
          {:ok, _key, _value} <- Exleveldb.iterator_move(it, last_key),
-         {:ok, @stateslot_prefix <> _slot, root} <- Exleveldb.iterator_move(it, :prev),
+         {:ok, @state_prefix <> _slot, root} <- Exleveldb.iterator_move(it, :prev),
```
MegaRedHand, line 55: This would match anything in the three tables, including `{block_root, state}` pairs, so `root` isn't necessarily a root.

test/integration/fork_choice/handlers_test.exs
```diff
@@ -16,7 +16,7 @@ defmodule Integration.ForkChoice.HandlersTest do
   test "on_block w/data from DB" do
     # NOTE: this test requires a DB with a state, and blocks for the state's slot and the next slot.
     # WARN: sometimes fails with "OffsetOutOfBounds" errors. Re-run the test in those cases.
-    {:ok, state} = StateStore.get_latest_state()
+    {:ok, state} = StateStore.get_latest_state(:block_root)
```
MegaRedHand, line 19: This change can be reverted

MegaRedHand (commented):
LGMT. Left some small comment, but after it's resolved we can merge.

PR #653, opened 2024-01-18
fix: remove db iteration in `get_head`

MegaRedHand:
We had some problems with DB concurrency because `get_head` was iterating the whole DB. This PR replaces that with a tree-shaped cache embedded in `Store` and keeps only the blocks used by fork-choice.


mpaulucci (dismissed):
great!

lib/lambda_ethereum_consensus/fork_choice/fork_choice.ex
```diff
@@ -40,8 +40,7 @@ defmodule LambdaEthereumConsensus.ForkChoice do
 
   @spec has_block?(Types.root()) :: boolean()
   def has_block?(block_root) do
-    block = get_block(block_root)
-    block != nil
+    GenServer.call(__MODULE__, {:has_block?, block_root}, @default_timeout)
```
MegaRedHand, line 43: I had to make this change because the tree lives only on memory. We can improve this by:
1. persist metadata about blocks (currently we only have implicit info from the tree)
2. load previous state when restarting the node (with metadata)
3. differentiate between `has_block_in_db?`, `can_process_block?`, and `block_was_processed?`

config/runtime.exs
```diff
@@ -61,3 +62,10 @@ block_time_ms =
 
 config :lambda_ethereum_consensus, LambdaEthereumConsensus.Telemetry,
   block_processing_buckets: [0.5, 1.0, 1.5, 2, 4, 6, 8] |> Enum.map(&(&1 * block_time_ms))
+
+if is_nil(jwt_secret) do
+  Logger.warning(
```
mpaulucci, line 67: is Logger ready in runtime.ex?

lib/lambda_ethereum_consensus/fork_choice/fork_choice.ex
mpaulucci: can you create a ticket?

config/runtime.exs
MegaRedHand: The warning works 🤔. Let me check if it's ready or not

PR #654, opened 2024-01-19
refactor: improve return values from execution client.

mpaulucci:
**Motivation**
Eventually we will need to store if the block was optimistic or not, so we need to pass that back to the state transition function


lib/lambda_ethereum_consensus/execution/execution_client.ex
```diff
@@ -9,21 +9,19 @@ defmodule LambdaEthereumConsensus.Execution.ExecutionClient do
   Verifies the validity of the data contained in the new payload and notifies the Execution client of a new payload
   """
   @spec verify_and_notify_new_payload(Types.ExecutionPayload.t()) ::
-          {:ok, boolean()} | {:error, any}
+          {:ok, atom()} | {:error, String.t()}
```
MegaRedHand, line 12: We should list the possible values in the spec: `:valid | :optimistic | :invalid`

MegaRedHand (dismissed):

MegaRedHand (approved):

PR #660, opened 2024-01-22
feat: add custom console logger.

mpaulucci:
**Motivation**
We want a logger in console that shows the information in a clear manner and looks professional.

**Description**
Screenshot 2024-01-22 at 18 39 47


config/runtime.exs
```diff
@@ -62,10 +61,3 @@ block_time_ms =
 
 config :lambda_ethereum_consensus, LambdaEthereumConsensus.Telemetry,
   block_processing_buckets: [0.5, 1.0, 1.5, 2, 4, 6, 8] |> Enum.map(&(&1 * block_time_ms))
-
```
mpaulucci, line 65: At this point the logger is not initialized, moved to Application start.

lib/lambda_ethereum_consensus/logger/console_logger.ex
```diff
@@ -0,0 +1,46 @@
+defmodule ConsoleLogger do
+  @moduledoc """
+  Custom logger formatter for console output.
+  """
+
+  def format(level, message, timestamp, metadata) do
+    pattern = Logger.Formatter.compile(" $time $message ")
```
MegaRedHand, line 7: You could move this to a `@` module constant. That way it's only run at compile time.

lib/lambda_ethereum_consensus/logger/console_logger.ex
```diff
@@ -0,0 +1,46 @@
...
+            Atom.to_string(key) <> IO.ANSI.reset() <> "=" <> format_metadata(key, value)
+        end
+      )
+
+    [format_level(level)] ++
+      [Logger.Formatter.format(pattern, level, message, timestamp, [])] ++
+      [formatted_metadata] ++ ["\n"]
+  rescue
+    err ->
+      inspect(err)
```
MegaRedHand, line 24: Why this?

lib/lambda_ethereum_consensus/logger/console_logger.ex
```diff
@@ -0,0 +1,46 @@
...
+    pattern = Logger.Formatter.compile(" $time $message ")
+
+    formatted_metadata =
+      Enum.map_join(
+        metadata,
+        " ",
+        fn {key, value} ->
+          IO.ANSI.bright() <>
+            Atom.to_string(key) <> IO.ANSI.reset() <> "=" <> format_metadata(key, value)
+        end
```
MegaRedHand, line 16: Nit: we could put this in a helper function

MegaRedHand (approved):
LGTM

lib/lambda_ethereum_consensus/logger/console_logger.ex
mpaulucci: If there is an error in the logger code, then it gets swallowed otherwise. With this, at least we log the error of the logger code.

lib/lambda_ethereum_consensus/logger/console_logger.ex
```diff
@@ -0,0 +1,46 @@
...
+
+  def format(level, message, timestamp, metadata) do
+    pattern = Logger.Formatter.compile(" $time $message ")
+
+    formatted_metadata =
+      Enum.map_join(
+        metadata,
+        " ",
+        fn {key, value} ->
+          IO.ANSI.bright() <>
```
MegaRedHand, line 14: We should disable colors when logging to a file (we can check with `IO.ANSI.enabled?()`)

mpaulucci: But this is just the console logger though

MegaRedHand: Yes. But it's the one used when piping data to a file (i.e. when doing `mix run > some_file.txt`).

MegaRedHand: Sure! In fact, [according to docs](https://hexdocs.pm/logger/1.16.0/Logger.html#module-backends-and-backwards-compatibility), it's the right way to do it.

MegaRedHand (approved):
🗿

PR #676, opened 2024-01-23
refactor: allow switching `Store.blocks` implementation

MegaRedHand:
Second version of #611

### DIFF
--- a/lib/lambda_ethereum_consensus/store/blocks/in_memory.ex
+++ b/lib/lambda_ethereum_consensus/store/blocks/in_memory.ex
@@ -1,0 +1,26 @@
+defmodule LambdaEthereumConsensus.Store.Blocks.InMemory do
+  @moduledoc false
+  alias Types.SignedBeaconBlock
+
+  @behaviour LambdaEthereumConsensus.Store.BlocksImpl
+
+  defstruct inner: %{}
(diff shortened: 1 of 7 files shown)


lib/lambda_ethereum_consensus/store/blocks/in_memory.ex
```diff
@@ -0,0 +1,26 @@
...
+
+  ##########################
+  ### Public API
+  ##########################
+
+  @impl true
+  @spec store_block(t(), Types.root(), SignedBeaconBlock.t()) :: t()
+  def store_block(%{inner: blocks} = state, block_root, signed_block) do
+    new_blocks = Map.put(blocks, block_root, signed_block)
+    %{state | inner: new_blocks}
```
mpaulucci, line 20: I think we should find a different solution that doesn't store the blocks in `store` when they are in memory. There should be a common api to store blocks if it's on memory or on db, and they should not leak like in this case IMO.

mpaulucci: I understand that it is done like this to not break spec test, but there has to be a better way.

MegaRedHand:
Closing this since we dropped support for in-memory storage

PR #682, opened 2024-01-24
test: run tests with db

MegaRedHand:


lib/mix/tasks/generate_spec_tests.ex
```diff
@@ -59,16 +59,19 @@ defmodule Mix.Tasks.GenerateSpecTests do
     module_name = "Elixir.#{c}.#{f}.#{r}Test" |> String.to_atom()
     runner_module = "Elixir.#{r}TestRunner" |> String.to_atom()
 
+    # TODO: we can isolate tests that use the DB from each other by using ExUnit's tmp_dir context option.
     header = """
     defmodule #{module_name} do
       use ExUnit.Case, async: false
 
       setup_all do
+        start_link_supervised!({LambdaEthereumConsensus.Store.Db, db_location: "test/generated/#{config}_#{fork}_#{runner}_test_db"})
```
mpaulucci, line 68: In favor of this change, but this would work even if we didn't do this? Since the fork choice tree contains the info of every block that is currently in the store?

mpaulucci, line 68: This would prevent tests to run in parallel right? Since we would try to spawn more than one genserver with the same name? (LambdaEthereumConsensus.Store.Db)

MegaRedHand: Yes, but I was having some problems when running it in CI vs _my machine_. It seems the DBs took some time to remove the lock file they use, and so failed to initialize. I intend to fix that in another PR.

MegaRedHand: Exactly! I'll try to start this only once in `test_helper.exs` in another PR

MegaRedHand: No. Different CI jobs don't share storage space (or at least I hope so!). The issue is with race conditions when going from a test module to the next one. We re-open the DB and sometimes it fails because the lockfile is still there.

PR #683, opened 2024-01-24
feat: enable ssz generic spec tests

### LINKED ISSUE #626: [SSZ] Enable all the spec tests for `SszGenericTestRunner`
# Description
Now that BitLists, BitVectors, and Vectors are ready, it is possible to enable all the spec tests for SszGenericTestRunner
Specs: [SSZ, generic tests](https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/ssz_generic)

f3r10:
**Motivation**

Closes #626


lib/ssz_ex.ex
```diff
@@ -5,6 +5,7 @@ defmodule LambdaEthereumConsensus.SszEx do
   alias LambdaEthereumConsensus.Utils.BitList
   alias LambdaEthereumConsensus.Utils.BitVector
   import alias LambdaEthereumConsensus.Utils.BitVector
```
MegaRedHand, line 7:
```suggestion
 alias LambdaEthereumConsensus.Utils.BitVector
```

Godspower-Eze:
The error is because `hash_tree_root()!` is not implemented for bitlist and bitvector and that's what I am supposed to work on so I would just make updates to this PR.

Arkenan (changes requested):
Partial review for the test files so that we don't manually tweak each container. I'll review the bitlist edge cases while these are addressed.

lib/ssz_ex.ex
```diff
@@ -367,26 +377,75 @@ defmodule LambdaEthereumConsensus.SszEx do
...
+        <<bit::integer-size(1), _rest::bitstring>> = bit_vector
+
+        if bit == 0 do
+          {:ok, BitVector.new(bit_vector, size)}
+        else
+          {:error, "ExcessBits"}
+        end
+
+      size == 0 and num_bytes == 0 ->
+        {:error, "ExcessBits"}
```
Arkenan, line 397: According to the specs:

> Empty vector types (`Vector[type, 0]`, `Bitvector[0]`) are illegal.

Can't we just return an error on `size == 0` regardless of the content?

Arkenan (changes requested):
Some more comments, will continue reviewing tomorrow!

lib/spec/runners/ssz_generic.ex
```diff
@@ -97,6 +101,46 @@ defmodule SszGenericTestRunner do
...
   end
 
+  defp assert_ssz(
+         "valid",
+         {:bitlist, _size} = schema,
+         real_serialized,
+         real_deserialized,
+         _expected_hash_tree_root
+       ) do
+    {:ok, deserialized_value} = SszEx.decode(real_deserialized, schema)
```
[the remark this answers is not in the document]
f3r10: There is a problem with the cases of bitlists and bitvectors.  The `value` placed on the value.yaml that is supposed to be the result after decoding the serialized value is the same as the serialized one. :sweat_smile:   

I am checking the lighthouse implementation and the `value.yaml` is decoded first. 
For the case `bitlist_16_lengthy_0`   the result of reading the `value.yaml`  file is 
```Some(Bitfield { bytes: [160, 92], len: 16, _phantom: PhantomData<ssz_types::bitfield::Variable<typenum::uint::UInt<typenum::uint::UInt<typenum::uint::UInt<typenum::uint::UInt<typenum::uint::UInt<typenum::uint::UTerm, typenum::bit::B1>, typenum::bit::B0>, typenum::bit::B0>, typenum::bit::B0>, typenum::bit::B0>>> })```

So for this reason I am also decoding that value first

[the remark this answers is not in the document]
Comment (f3r10) [reply lib/spec/runners/ssz_generic.ex:110]: oh I just realized that with what you write below `def sanitize(_k, bitlist, {:bitlist, size} = schema)` this issue would be solved :tada:

File: lib/ssz_ex.ex [hunk 4]
```diff
@@ -53,13 +56,17 @@ defmodule LambdaEthereumConsensus.SszEx do
   def decode(binary, {:int, size}), do: decode_uint(binary, size)
   def decode(value, {:bytes, _}), do: {:ok, value}
 
-  def decode(binary, {:list, basic_type, size}) do
+  def decode(binary, {:list, basic_type, size} = schema) do
     if variable_size?(basic_type),
       do: decode_variable_list(binary, basic_type, size),
-      else: decode_list(binary, basic_type, size)
+      else: decode_list(binary, basic_type, size, schema)
```
Arkenan, line 62: The `decode_list` is a very confusing function, most likely because:

1. It's called `list` but it's used for both list and vector. It should be called `decode_collection`, or something like that. 
2. It passes the whole schema instead of just passing :list or :vector.

We will delete this function. In general it's very weird to pass an atom like :list or vector UNLESS there will be a dynamic or unpredictable situation for that parameter. Here, we'll always pass :list when it's a list or :vector when it's a vector. This is very similar to what I commented on assert_ssz_valid vs assert_ssz_invalid.

As a matter of fact, this is over-reutilization makes the code exceedenly complex, as the only thing these functions share is getting the fixed size of a variable and then the decoding sections. All checks are different, or may not even exist.

Let's have `decode_fixed_list` and `decode_fixed_vector` functions. I'll comment below some tips on how to do that.

lib/ssz_ex.ex
```diff
@@ -367,26 +377,67 @@ defmodule LambdaEthereumConsensus.SszEx do
...
     end
   end
 
+  def check_valid_list_size_prev_decode({:vector, _, _}, fixed_size, size, binary)
+      when fixed_size * size == byte_size(binary),
+      do: true
+
+  def check_valid_list_size_prev_decode({:vector, _, _}, _fixed_size, _size, _binary), do: false
+
+  def check_valid_list_size_prev_decode({:list, _, _}, _fixed_size, _size, _binary), do: true
```
Arkenan, line 427: The fact that list don't have a pre-check is a big evidence towards this being over-reutilized.

lib/ssz_ex.ex
```diff
@@ -367,26 +377,67 @@ defmodule LambdaEthereumConsensus.SszEx do
...
+    else
+      {:error, "invalid list binary size"}
     end
   end
 
+  def check_valid_list_size_prev_decode({:vector, _, _}, fixed_size, size, binary)
+      when fixed_size * size == byte_size(binary),
+      do: true
+
+  def check_valid_list_size_prev_decode({:vector, _, _}, _fixed_size, _size, _binary), do: false
```
Arkenan, line 425: After the proposed refactor, this ones should return `:ok` or an error tuple.

Arkenan (changes requested):
Some more comments, this time on list vs vector decoding. The only thing left to review is the decoding of containers. I'm finding some strange things in the sanitize_offset funciton that I'll be adding comments on.

lib/ssz_ex.ex
```diff
@@ -527,45 +578,89 @@ defmodule LambdaEthereumConsensus.SszEx do
   defp decode_container(binary, module) do
     schemas = module.schema()
     fixed_length = get_fixed_length(schemas)
-    <<fixed_binary::binary-size(fixed_length), variable_binary::bitstring>> = binary
 
-    with {:ok, fixed_parts, offsets} <- decode_fixed_section(fixed_binary, schemas, fixed_length),
-         {:ok, variable_parts} <- decode_variable_section(variable_binary, offsets) do
-      {:ok, struct!(module, fixed_parts ++ variable_parts)}
+    if variable_size?(module) do
```
Arkenan, line 582: Maybe we can benefit of having variable_size? be a guard and use it to pattern match in the decode_container definition? e.g.

```elixir
defp decode_container(binary, module) when variable_size?(module) do
 #...
end
```

If that's not possible we can do the same we did for vectors and lists, and have the if + `decode_variable_container` or `decode_fixed_container`

lib/ssz_ex.ex
```diff
@@ -582,14 +677,22 @@ defmodule LambdaEthereumConsensus.SszEx do
   end
 
   # https://notes.ethereum.org/ruKvDXl6QOW3gnqVYb8ezA?view
-  defp sanitize_offset(offset, previous_offset, num_bytes, num_fixed_bytes) do
+  defp sanitize_offset(offset, previous_offset, _num_bytes, num_fixed_bytes)
```
Arkenan, line 680: If we're always going to re-assign to the same variable, instead of sanityze I'd call this `check_offset_valid` and return `:ok` if that's the case.

lib/ssz_ex.ex
```diff
@@ -402,40 +406,48 @@ defmodule LambdaEthereumConsensus.SszEx do
...
-             |> flatten_results(),
-           :ok <- check_valid_list_size(schema, size, length(decoded_list)) do
-        result
-      end
-    else
-      {:error, "invalid list binary size"}
+    with {:ok, decoded_list} = result <-
+           binary
+           |> decode_chunk(fixed_size, basic_type)
+           |> flatten_results(),
```
Arkenan, line 415: Let's have these three lines inside a function called `def decode_fixed_collection(collection, size, base_type)`, this is the only actual reusable piece of code between the two functions. This new function should replace the `decode_chunk/3` helper.

Arkenan (dismissed):
LGTM

Arkenan (approved):
LGTM

PR #685, opened 2024-01-24
refactor: remove blocks and states from store

MegaRedHand:
Depends on #682

This PR drops support for storing blocks and states in memory.

### DIFF
--- a/lib/beacon_api/controllers/v1/beacon_controller.ex
+++ b/lib/beacon_api/controllers/v1/beacon_controller.ex
@@ -23,13 +23,13 @@ defmodule BeaconApi.V1.BeaconController do
   def get_state_root(conn, %{state_id: state_id}) do
     with {:ok, {root, execution_optimistic, finalized}} <-
            BeaconApi.Utils.parse_id(state_id) |> ForkChoice.Helpers.root_by_id(),
-         {:ok, state_root} <- ForkChoice.Helpers.get_state_root(root) do
+         %{} = state_root <- ForkChoice.Helpers.get_state_root(root) do
       conn |> root_response(state_root, execution_optimistic, finalized)
     else
       {:error, error_msg} ->
         conn |> ErrorController.internal_error("Error: #{inspect(error_msg)}")
 
-      :not_found ->
+      nil ->
         conn |> ErrorController.not_found(nil)
 
       :empty_slot ->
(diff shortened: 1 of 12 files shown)


lib/beacon_api/controllers/v1/beacon_controller.ex
```diff
@@ -3,13 +3,14 @@ defmodule BeaconApi.V1.BeaconController do
...
   alias LambdaEthereumConsensus.ForkChoice
+  alias LambdaEthereumConsensus.Store.Blocks
   alias LambdaEthereumConsensus.Store.BlockStore
   use BeaconApi, :controller
 
   plug(OpenApiSpex.Plug.CastAndValidate, json_render_error_v2: true)
 
   @doc """
-  action is an atom that correspond to the controller action's function atoms declared on `BeaconApi.Router` 
+  action is an atom that correspond to the controller action's function atoms declared on `BeaconApi.Router`
```
mpaulucci, line 13: would remove this @doc

MegaRedHand: I'll leave it as a comment. It's important to document this, since the function creates (a bounded anount of) new atoms at runtime.

lib/beacon_api/controllers/v1/beacon_controller.ex
```diff
@@ -3,13 +3,14 @@ defmodule BeaconApi.V1.BeaconController do
   alias BeaconApi.ErrorController
 
   alias LambdaEthereumConsensus.ForkChoice
+  alias LambdaEthereumConsensus.Store.Blocks
   alias LambdaEthereumConsensus.Store.BlockStore
```
mpaulucci, line 7: What is the difference between `LambdaEthereumConsensus.Store.Blocks` and `LambdaEthereumConsensus.Store.BlockStore`?

MegaRedHand: `Blocks` is the cache. We should rename these 😅

MegaRedHand: I opened #686 for that

mpaulucci: nice!

mpaulucci (approved):

PR #692, opened 2024-01-26
feat: add "DB-only" mode

### LINKED ISSUE #684: Allow setting up the node with DB and caches only
This would be useful for interacting with the DB in tests or when debugging interactively

MegaRedHand:
Closes #684

This PR introduces a new flag `--mode`, which sets the application mode. "db" starts the application with DB-related modules only, while "full" is the default and starts all features. This is useful for manual debugging and for running tests.

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -1,8 +1,8 @@
 import Config
 
-{args, _remaining_args, _errors} =
-  OptionParser.parse(System.argv(),
-    switches: [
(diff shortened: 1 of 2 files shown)


mpaulucci (commented):
Not sure about this one, it seems we do would want a BeaconChain in test environment, but we don't want the LibP2P part (gossip, incoming requests, etc..)

Arkenan (approved):

PR #696, opened 2024-01-26
perf: make `verify_and_notify_new_payload` calls async.

### LINKED ISSUE #694: Calling the execution client from the block processing should be a non blocking operation
Currently, when doing this call https://github.com/lambdaclass/lambda_ethereum_consensus/blob/fba703f4d15b499a2a92931aed432138a10edfe0/lib/lambda_ethereum_consensus/state_transition/state_transition.ex#L126 we're blocking the Process from being able to continue processing the block. We should make this a non blocking operation by using `Task.async`.

Acceptance criteria:
- Calling `ExecutionClient.verify_and_notify_new_payload` should be non-blocking.
- We probably want to move this call to https://github.com/lambdaclass/lambda_ethereum_consensus/blob/fba703f4d15b499a2a92931aed432138a10edfe0/lib/lambda_ethereum_consensus/fork_choice/handlers.ex#L158
- Before the function ends, we need to await the result and store the execution status in the Block DB

mpaulucci:
**Motivation**
We want to make this call async for performance reasons, so it allows the block processor to perform the state transition.

Closes #694


lib/spec/runners/operations.ex
```diff
@@ -97,18 +97,20 @@ defmodule OperationsTestRunner do
       YamlElixir.read_from_file!(case_dir <> "/execution.yaml")
       |> SpecTestUtils.sanitize_yaml()
 
-    status = if execution_valid, do: :valid, else: :invalid
+    _status = if execution_valid, do: :valid, else: :invalid
 
-    result =
-      Operations.process_execution_payload(pre, body, fn _payload -> {:ok, status} end)
```
MegaRedHand, line 103: Doesn't this change the behaviour? Why do these tests pass?

lib/spec/runners/operations.ex
```diff
@@ -97,18 +97,20 @@ defmodule OperationsTestRunner do
       YamlElixir.read_from_file!(case_dir <> "/execution.yaml")
       |> SpecTestUtils.sanitize_yaml()
 
-    status = if execution_valid, do: :valid, else: :invalid
+    _status = if execution_valid, do: :valid, else: :invalid
```
MegaRedHand, line 100: We can delete this if unused

lib/spec/runners/operations.ex
mpaulucci: What do you mean? There were some tests that return invalid from the execution client (trivial tests tbh) that we can't have anymore because we moved the call to another place in the code.

So we filter those tests out

MegaRedHand (approved):

PR #699, opened 2024-01-30
test(spec): move spec test runners to test folder

mpaulucci:
**Motivation**
We shouldn't have test runner code in our lib folder. I understand that we did this because we thought there was no other way, but there is.


lib/lambda_ethereum_consensus/p2p/block_downloader.ex
```diff
@@ -50,7 +50,6 @@ defmodule LambdaEthereumConsensus.P2P.BlockDownloader do
          {:ok, chunks} <- parse_response(response_chunk),
          {:ok, blocks} <- decode_chunks(chunks),
          :ok <- verify_batch(blocks, slot, count) do
-      # TODO: handle cases where slot is empty
```
mpaulucci, line 53: unrelated, but this comment doesn't apply anymore.

mix.exs
```diff
@@ -12,7 +12,8 @@ defmodule LambdaEthereumConsensus.MixProject do
       elixirc_paths: compiler_paths(Mix.env()),
       warn_test_pattern: "_remove_warning.exs",
       preferred_cli_env: [
-        dialyzer: :test
+        dialyzer: :test,
+        generate_spec_tests: :test
```
mpaulucci, line 16: this is the main change, running the task in :test env so that the runners are included as part of the compilation.

Arkenan (approved):
LGTM

Arkenan (approved):

PR #702, opened 2024-02-01
docs: refine per-block-process design

Arkenan:
After some feedback:

- Removed the downloader task, it can be a block processor stage too. It removes one more passing around of the block and one more independent thing to take care about.
- The GenServer now exits instead of waiting, and is respawned at a latter stage instead of receiving a message.
- Some typos fixed.


MegaRedHand (dismissed):
LGTM 👍

PR #703, opened 2024-02-01
feat: merklelization of list of composite types

### LINKED ISSUE #678: [SSZ] Support for merklelization of list of composite objects

Godspower-Eze:
Closes #678


test/unit/ssz_ex_test.exs
```diff
@@ -217,6 +218,102 @@ defmodule Unit.SSZExTest do
...
+  end
+
+  test "hash tree root of vector of composite objects" do
+    ## list of containers
+    checkpoint = %Checkpoint{
+      epoch: 12_345,
+      root: Base.decode16!("0100000000000000000000000000000000000000000000000000000000000001")
+    }
+
+    module = SszEx.get_module(checkpoint)
```
MegaRedHand, line 285: This should be the same as using `Checkpoint`. Also, you can just do `%module{} = checkpoint` here

lib/ssz_ex.ex
```diff
@@ -24,6 +24,8 @@ defmodule LambdaEthereumConsensus.SszEx do
   @spec hash_nodes(binary(), binary()) :: binary()
   def hash_nodes(left, right), do: :crypto.hash(:sha256, left <> right)
 
+  def get_module(%module{} = _map), do: module
```
MegaRedHand, line 27: I think we should remove this.

lib/ssz_ex.ex
```diff
@@ -151,6 +150,30 @@ defmodule LambdaEthereumConsensus.SszEx do
     end
   end
 
+  def hash_tree_root_list_composite_type(list, inner_schema, limit, len) do
+    root =
+      list
+      |> Enum.reduce(<<>>, fn value, acc_roots ->
+        acc_roots <> hash_tree_root!(value, inner_schema)
```
MegaRedHand, line 157: This should use `hash_tree_root` as it could fail, right?

lib/ssz_ex.ex
```diff
@@ -112,29 +114,26 @@ defmodule LambdaEthereumConsensus.SszEx do
...
-      # TODO
-      # hash_tree_root_list_complex_type(list, {:list, type, size}, limit)
-      {:error, "Not implemented"}
-    else
-      packed_chunks = pack(list, {:list, type, size})
-      limit = chunk_count({:list, type, size})
-      len = length(list)
+  def hash_tree_root(list, {:list, type, _size} = schema) do
+    limit = chunk_count(schema)
+    len = length(list)
```
MegaRedHand, line 119: We should check that `len < size` here, to avoid checks in the following functions

lib/ssz_ex.ex
```diff
@@ -112,29 +114,26 @@ defmodule LambdaEthereumConsensus.SszEx do
...
-      {:error, "Not implemented"}
-    else
-      packed_chunks = pack(vector, {:list, type, size})
+  def hash_tree_root(vector, {:vector, type, _size} = schema) do
+    if basic_type?(type) do
+      packed_chunks = pack(vector, schema)
       hash_tree_root_vector_basic_type(packed_chunks)
+    else
+      hash_tree_root_vector_composite_type(vector, type)
     end
```
MegaRedHand, line 137: We can simplify this by doing:

```elixir
    if basic_type?(type) do
      pack(vector, schema)
    else
      # this is just the `Enum.reduce` in both `hash_tree_root_*_composite_type` functions
      list_hash_tree_roots(vector, schema)
    end
    # we can drop the `_basic_type` here
    |> hash_tree_root_vector_basic_type()
```

MegaRedHand (changes requested):

lib/ssz_ex.ex
MegaRedHand: What I mean is that `hash_tree_root!` can raise errors, so we should use `hash_tree_root` directly, and propagate errors.

lib/ssz_ex.ex
MegaRedHand: Yes. Since both are Elixir lists, you should be able to reuse the `list_hash_tree_roots` function. The idea is to move the `Enum.reduce` into that single function, to avoid duplicating code. After that, you can further reduce duplication by hashing lists with the same vector function, but mixing their length at the end (the Rust SSZ lib we use does something similar).

Godspower-Eze: I don't understand the second part. Please explain

lib/ssz_ex.ex
Godspower-Eze: So, we return the error instead of it failing, right?

MegaRedHand: Yes

lib/ssz_ex.ex
MegaRedHand: Actually, forget I mentioned that second part. `merkleize_chunks_with_virtual_padding` already does this.

lib/ssz_ex.ex
Godspower-Eze: I don't think this should be the case here.

It shouldn't fail and this is inside of `Enum.reduce` so if it fails then there's probably an issue with the values or a bug.

lib/ssz_ex.ex
```diff
@@ -159,6 +156,22 @@ defmodule LambdaEthereumConsensus.SszEx do
...
   end
 
+  def hash_tree_root_list_composite_type(list, inner_schema, limit, len) do
+    root =
+      list_hash_tree_root(list, inner_schema)
+      |> merkleize_chunks_with_virtual_padding(limit)
+      |> mix_in_length(len)
+
+    {:ok, root}
+  end
```
MegaRedHand, line 166: This repeats code from `hash_tree_root_list_basic_type`.

lib/ssz_ex.ex
Godspower-Eze: I am now handling this in `hash_tree_root_list` and it's cleaner. What do you think?

lib/ssz_ex.ex
```diff
@@ -84,14 +84,14 @@ defmodule LambdaEthereumConsensus.SszEx do
   end
 
   @spec hash_tree_root!(list(), {:list, any, non_neg_integer}) :: Types.root()
-  def hash_tree_root!(list, {:list, type, size}) do
-    {:ok, root} = hash_tree_root(list, {:list, type, size})
+  def hash_tree_root!(list, {:list, _type, _size} = schema) do
```
MegaRedHand, line 87: Can't we just call `hash_tree_root` regardless of the schema?
 ```elixir
def hash_tree_root!(value, schema) do
  {:ok, root} = hash_tree_root(value, schema)
  root
end
```

File: lib/ssz_ex.ex (hunk 4 above)
Comment (MegaRedHand) [reply lib/ssz_ex.ex:117]: Looks good!

File: lib/ssz_ex.ex [hunk 8]
```diff
@@ -112,35 +112,36 @@ defmodule LambdaEthereumConsensus.SszEx do
...
-      hash_tree_root_vector_basic_type(packed_chunks)
+      list_hash_tree_root(vector, type)
     end
+    |> hash_tree_root_vector()
   end
 
-  @spec hash_tree_root_list_basic_type(binary(), non_neg_integer, non_neg_integer) ::
-          {:ok, Types.root()} | {:error, String.t()}
-  def hash_tree_root_list_basic_type(chunks, limit, len) do
+  def hash_tree_root_vector(chunks) do
```
MegaRedHand, line 138: We have to check that the vector is the right size

lib/ssz_ex.ex
MegaRedHand: You can use `reduce_while` and an `{:ok, value}` accumulator to handle errors here. I understand this would fail only in rare circumstances, but we should keep `hash_tree_root` as an exception-free alternative to `hash_tree_root!`

lib/ssz_ex.ex
Godspower-Eze: Something like this

```
    if chunks_len > limit do
      {:error, "chunk size exceeds limit"}
```

??

MegaRedHand: More like `chunks_len != limit`, since vectors are of fixed size

Godspower-Eze: I don't think that's needed for vector. It doesn't have `limit`.

lib/ssz_ex.ex
Godspower-Eze: Okay. What is the ideal scenario to use `hash_tree_root!`?

lib/ssz_ex.ex
MegaRedHand: The vector has a `size`. We should check that `length(vector) == size`, just for consistency, since we don't have a way to check that at the type level

lib/ssz_ex.ex
MegaRedHand: When we are sure that there won't be any errors.

lib/ssz_ex.ex
Godspower-Eze: Gotcha

lib/ssz_ex.ex
Godspower-Eze: Updated it. Is this okay?

MegaRedHand: Yes 👍

Godspower-Eze:
I get this when I try to replicate the 'start and stop the node` check on my machine:

```bash
INFO 14:41:43.043 [Checkpoint sync] Initiating checkpoint sync 
%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}%ArgumentError{message: "invalid ANSI sequence specification: :default"}** (Mix) Could not start application lambda_ethereum_consensus: LambdaEthereumConsensus.Application.start(:normal, []) returned an error: shutdown: failed to start child: LambdaEthereumConsensus.Beacon.BeaconNode
    ** (EXIT) exited in: Task.await_many([%Task{mfa: {LambdaEthereumConsensus.Beacon.CheckpointSync, :get_state, 1}, owner: #PID<0.1400.0>, pid: #PID<0.1401.0>, ref: #Reference<0.0.179203.2421924730.938016772.120703>}, %Task{mfa: {LambdaEthereumConsensus.Beacon.CheckpointSync, :get_block, 1}, owner: #PID<0.1400.0>, pid: #PID<0.1402.0>, ref: #Reference<0.0.179203.2421924730.938016772.120704>}], 60000)
        ** (EXIT) time out
```

MegaRedHand:
That's an unrelated bug. It's already fixed in #725, though.

lib/ssz_ex.ex
```diff
@@ -80,78 +80,97 @@ defmodule LambdaEthereumConsensus.SszEx do
...
+  @spec hash_tree_root!(any, any) :: Types.root()
+  def hash_tree_root!(value, schema) do
+    {:ok, root} = hash_tree_root(value, schema)
+    root
+  end
 
-  @spec hash_tree_root!(non_neg_integer, {:int, non_neg_integer}) :: Types.root()
-  def hash_tree_root!(value, {:int, size}), do: pack(value, {:int, size})
+  @spec hash_tree_root(boolean, atom) :: Types.root()
+  def hash_tree_root(value, :bool), do: {:ok, pack(value, :bool)}
```
MegaRedHand, line 90: Don't we need to merkleize after packing?

Godspower-Eze: For ints and bools, we don't because it's a single chunk

MegaRedHand: But the spec says:

> * `merkleize(pack(value))` if `value` is a basic object or a vector of basic objects.

Here we're only doing `pack(value)` 🤔

Godspower-Eze: Here:

> Then, merkleize the chunks (empty input is padded to 1 zero chunk):
- If 1 chunk: the root is the chunk itself.
- If > 1 chunks: merkleize as binary tree.

MegaRedHand: So `hash_tree_root(3) == <<3::little-size(256>>` ?

Godspower-Eze: Not exactly, but yes

MegaRedHand: 🤯

MegaRedHand (approved):
Really impressive work!

PR #708, opened 2024-02-05
chore: alias `mix compile` and `mix clean` to `make` targets

MegaRedHand:
This PR changes the mix compile and clean tasks to also compile and clean other artifacts, via mix aliases.


MegaRedHand:
Seems to clash with CI. We should revisit this

PR #709, opened 2024-02-05
refactor: make block processing async inside PendingBlocks

mpaulucci:
**Motivation**
Before this change, PendingBlocks was waiting for ForkChoice to process the block before updating it's status. This sometimes results in a timeout if there are many blocks that need to be processed, causing the PendingBlocks genserver to crash.

**Description**
- Refactored PendingBlocks state to store a map of blocks with it's corresponding status, instead of having one map per status.
- Adapt PendingBlocks to the fact that all calls to ForkChoice must be async


lib/lambda_ethereum_consensus/beacon/pending_blocks.ex
```diff
@@ -68,44 +71,39 @@ defmodule LambdaEthereumConsensus.Beacon.PendingBlocks do
...
           state
 
-        # If already in fork choice, remove from pending
-        ForkChoice.has_block?(block_root) ->
-          state |> Map.update!(:pending_blocks, &Map.delete(&1, block_root))
-
         # If parent is not in fork choice, download parent
-        not ForkChoice.has_block?(parent_root) ->
-          state |> Map.update!(:blocks_to_download, &MapSet.put(&1, parent_root))
+        !Blocks.get_block(parent_root) ->
```
MegaRedHand, line 99: This introduces a bug: if the parent is in the DB this would be false, and we'd try to process the block even if its parent wasn't processed, causing it to fail and marking it as an invalid block

mpaulucci: But this assumes that a block is stored in the DB only if it is processed (and valid). We should not store invalid (or potentially invalid) blocks.

Do we store blocks before processing them today?

MegaRedHand: No. But this can happen if the node restarts after processing some blocks

mpaulucci: I think that this is probably out of scope from this PR

MegaRedHand: For sure. We could patch it up by just nuking the database in the makefile to avoid this error when debugging, but we need to find a better way to handle this.

mpaulucci: Would you like me to add this to the PR?

lib/lambda_ethereum_consensus/beacon/pending_blocks.ex
```diff
@@ -114,65 +112,45 @@ defmodule LambdaEthereumConsensus.Beacon.PendingBlocks do
...
-
-    roots_to_remove =
+    new_state =
       downloaded_blocks
-      |> Enum.map(&Ssz.hash_tree_root!(&1.message))
-      |> MapSet.new()
-      |> MapSet.union(blocks_in_store)
+      |> Enum.reduce(state, fn signed_block, state ->
+        block_root = Ssz.hash_tree_root!(signed_block.message)
+        state |> Map.put(block_root, {signed_block, :pending})
```
Arkenan, line 140: Maybe both here and when casting `:add_block` we should call a single function that does the ssz hashing + adding the block to the state as pending? Also in that function we could validate that the block is not already present in the state, as in some cases we might get the block from two different sources (e.g. through gossip and manual download) and we probably don't want to replace an existing entry.

lib/lambda_ethereum_consensus/beacon/pending_blocks.ex
```diff
@@ -68,44 +71,39 @@ defmodule LambdaEthereumConsensus.Beacon.PendingBlocks do
...
       parent_root = signed_block.message.parent_root
 
       cond do
+        # If already processed, remove it
+        Blocks.get_block(block_root) ->
+          state |> Map.delete(block_root)
+
         # If parent is invalid, block is invalid
-        state.invalid_blocks |> Map.has_key?(parent_root) ->
+        get_block_status(state, parent_root) == :invalid ->
```
Arkenan, line 87: nit: can we assign `status = get_block_status(state, parent_root)` before the cond so we don't call it multiple times?

lib/lambda_ethereum_consensus/beacon/pending_blocks.ex
```diff
@@ -114,65 +112,45 @@ defmodule LambdaEthereumConsensus.Beacon.PendingBlocks do
...
-      |> case do
-        {:ok, signed_blocks} ->
-          signed_blocks
-
-        {:error, reason} ->
-          Logger.debug("Block download failed: '#{reason}'")
-          []
+      if Enum.empty?(blocks_to_download) do
+        []
+      else
```
Arkenan, line 122: nit and optional: maybe the empty case should be automatically handled by the block downloader with a 

```elixir
BlockDownloader.request_blocks_by_root([]), do: []
```

It's nicer to use 

```elixir
blocks_to_download
|> Enum.take(16)
|> BlockDownloader.request_blocks_by_root()
#...
```

directly and it makes the downloader module more robust.

Arkenan (dismissed):
LGTM, left some nit comments.

lib/lambda_ethereum_consensus/beacon/pending_blocks.ex
Arkenan: I think this is a more complex discussion outside of the scope of the PR, which is supporting recoverability. To do so, we should save the state of a block in the DB.

MegaRedHand: Just to clarify, I meant we need to find a better way to handle this _in a following PR_. For now, we can just go for the nuclear option.

MegaRedHand (approved):

PR #710, opened 2024-02-05
feat: snappy framing format (decompression)

### LINKED ISSUE #224: [Snappy] Implement Snappy framing format in Elixir

MegaRedHand:
Closes #224
Depends on #713

This PR continues the work done in #543. What's missing is implementing Snappy streaming compression, and running some benchmarks comparing it to the `Snappy` NIF.


test/unit/snappyex_test.exs
```diff
@@ -0,0 +1,108 @@
+defmodule Unit.SnappyExTest do
+  use ExUnit.Case
+  use ExUnitProperties
+  doctest SnappyEx
+
+  @empty_stream <<0xFF, 6::little-size(24)>> <> "sNaPpY"
+
+  def assert_snappy_decompress(compressed, uncompressed) do
```
Arkenan, line 8: defp?

test/unit/snappyex_test.exs
```diff
@@ -0,0 +1,108 @@
...
+
+    assert_snappy_decompress(compressed_payload, expected)
+
+    # Compressed chunks
+    msg = "0011FF060000734E61507059000A0000B3A056EA1100003E0100"
+    "00" <> "11" <> compressed_payload = msg
+    expected = Base.decode16!("0000000000000000000000000000000000")
+
+    assert_snappy_decompress(compressed_payload, expected)
+  end
```
Arkenan, line 66: These should be 3 separate tests.

test/unit/snappyex_test.exs
```diff
@@ -0,0 +1,108 @@
...
+    stream = Enum.join([@empty_stream, header, checksum, compressed_data])
+
+    assert SnappyEx.decompress(stream) == {:ok, data}
+  end
+
+  test "decompress GetMetadata responses" do
+    # Uncompressed chunks
+    msg = "0011FF060000734E6150705901150000F1D17CFF0008000000000000FFFFFFFFFFFFFFFF0F"
+    # status <> length <> ...
+    "00" <> "11" <> compressed_payload = msg
```
Arkenan, line 49: Why do we have "0011" in the message if we remove it afterwards?

test/unit/snappyex_test.exs
```diff
@@ -0,0 +1,108 @@
...
+  end
+
+  test "decompress Ping responses" do
+    for {msg, expected} <- [
+          {"0008FF060000734E61507059010C0000B18525A04300000000000000", "4300000000000000"},
+          {"0008FF060000734E61507059010C00000175DE410100000000000000", "0100000000000000"},
+          {"0008FF060000734E61507059010C0000EAB2043E0500000000000000", "0500000000000000"},
+          {"0008FF060000734E61507059010C0000290398070000000000000000", "0000000000000000"}
+        ] do
+      "00" <> "08" <> compressed_payload = msg
```
Arkenan, line 75: Same here.

test/unit/snappyex_test.exs
```diff
@@ -0,0 +1,108 @@
...
+      expected = Base.decode16!(expected)
+
+      assert_snappy_decompress(compressed_payload, expected)
+    end
+
+    # Error response
+    msg =
+      "011CFF060000734E6150705900220000EF99F84B1C6C4661696C656420746F20756E636F6D7072657373206D657373616765"
+
+    "01" <> "1C" <> compressed_payload = msg
```
Arkenan, line 85: Same here. If this is a non-snappy related header (e.g. gossip identifier) we may either not put it in the test or have a function that removes the first four bytes in every test, but we're not using any of those for testing.

Arkenan (approved):
LGTM, added some nit comments on testing.

test/unit/snappyex_test.exs
MegaRedHand: The test was copied from the first snappy tests we had. This was intended as an example of manually decoding a message from a peer. Do you think we should move it somewhere else? Or should I just remove it?

MegaRedHand: I just removed it. We can get new samples from peers anyways.

Arkenan (approved):
LGTM

PR #722, opened 2024-02-06
perf: inline `SszEx.hash` function

MegaRedHand:
Since the function is just an "alias" to `:crypto.hash`, we can safely inline all calls. This gives a ~5% performance increase when hashing, according to benchmarks.

### DIFF
--- a/lib/ssz_ex.ex
+++ b/lib/ssz_ex.ex
@@ -18,6 +18,7 @@ defmodule LambdaEthereumConsensus.SszEx do
   @zero_chunk <<0::size(@bits_per_chunk)>>
   @zero_hashes ZeroHashes.compute_zero_hashes()
 
+  @compile {:inline, hash: 1}
   @spec hash(iodata()) :: binary()
   def hash(data), do: :crypto.hash(:sha256, data)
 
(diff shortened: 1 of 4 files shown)


lib/types/store.ex
```diff
@@ -39,6 +39,7 @@ defmodule Types.Store do
           proposer_boost_root: Types.root() | nil,
           equivocating_indices: MapSet.t(Types.validator_index()),
           checkpoint_states: %{Checkpoint.t() => BeaconState.t()},
+          # NOTE: the `Checkpoint` values in latest_messages are `LatestMessage`s
           latest_messages: %{Types.validator_index() => Checkpoint.t()},
```
MegaRedHand, line 43: I was puzzled by this when working on other things, so I added this comment

lib/ssz_ex.ex
```diff
@@ -18,6 +18,7 @@ defmodule LambdaEthereumConsensus.SszEx do
   @zero_chunk <<0::size(@bits_per_chunk)>>
   @zero_hashes ZeroHashes.compute_zero_hashes()
 
+  @compile {:inline, hash: 1}
```
mpaulucci, line 21: TIL

lib/ssz_ex.ex
```diff
@@ -18,6 +18,7 @@ defmodule LambdaEthereumConsensus.SszEx do
...
   @spec hash(iodata()) :: binary()
```
mpaulucci, line 22: shouldnt this function live outside of SSZ? I see it's being used in other places too, like shuffling

mpaulucci (approved):

lib/ssz_ex.ex
MegaRedHand: Maybe. We just added it here since adding a new module for this function seemed like too much

PR #729, opened 2024-02-07
feat: persist Store

### LINKED ISSUE #662: Persist `Store.tree_cache` information

MegaRedHand:
Closes #662

This PR persists the whole state in the DB, by dumping the Store object in parallel each time a block is processed. It also makes some other small changes to do this:

- adds hardcoded genesis validator roots for mainnet and sepolia, which are used for validating checkpoint-sync anchors
- refactors the way arguments are passed from `BeaconNode` to its children


lib/types/store.ex
```diff
@@ -137,4 +138,15 @@ defmodule Types.Store do
       {:error, :not_found} -> store
     end
   end
+
+  def fetch_store do
```
mpaulucci, line 142: I would put this in `LambdaEthereumConsensus.Store.StoreStore` do we don't overload this module. And we should rename (DB) Store to Storage or something xD

mpaulucci (approved):
LGTM

PR #732, opened 2024-02-08
docs: add bitvector and bitlist documentation

Arkenan:


docs/bitvectors.md
```diff
@@ -0,0 +1,214 @@
...
+00000011 00000001
+```
+
+In this representation, thinking of it as a byte array, we get `le = [3, 1]`. The lowest index, `le[0] = 3` means the lowest significant byte, and the highest index, `le[1] = 1` is the most significant byte. So while little endian is less readable, it is frequently used to represent integers as binaries because of this property.
+
+## Bit vectors
+
+### Little endian bit order
+
+Why would we need a third representation? Let's first pose the problem. We want to represent a set of booleans. Imagine we have a fixed amount of validators, equal to 9, and we want to represent wether they attested in a block or not. We may represent this as follows:
```
MegaRedHand, line 39:
```suggestion
Why would we need a third representation? Let's first pose the problem. We want to represent a set of booleans. Imagine we have a fixed amount of validators, equal to 9, and we want to represent whether they attested in a block or not. We may represent this as follows:
```

Fun fact: a wether is a castrated ram

docs/bitvectors.md
```diff
@@ -0,0 +1,214 @@
...
+11000000 10000000
+```
+
+Moving it to little endian byte order (we go byte by byte and reverse the bits):
+
+```
+00000011 00000001
+```
+
+Which is what I'll send over the network. This is what SSZ calls `bitvectors`, which is a binary representing an array of booleans of constant size. We know that this array is of size 9 beforehand, so we know what bits are padding and should be ignored. For variable sized bit arrays we'll use `bitlists`, which we'll talk about later.
```
MegaRedHand, line 91:
```suggestion
Which is what SSZ calls `bitvectors`, and what nodes send over the network: a binary representing an array of booleans of constant size. We know that this array is of size 9 beforehand, so we know what bits are padding and should be ignored. For variable-sized bit arrays, we'll use `bitlists`, which we'll talk about later.
```

docs/bitvectors.md
```diff
@@ -0,0 +1,214 @@
...
+
+### Sentinel bits
+
+In reality, there's not a fixed amount of validators, if someone deposits 32ETH in the deposit contract, a new validator will join the set. `bitlists` are used to represent boolean arrays of variable size like this one. Conceptually, they use the little-endian bit order too, but they use a strategy called `sentinel bit` to mark where it ends. Let's imagine, again, that we're representing the same set of 9 validators as before. We start with the following 9 bits:
+
+```
+110000001
+```
+
+To serialize this and send it over the network, I do the following:
```
MegaRedHand, line 133: \*insert mega red joke here\*

```suggestion
To serialize this and send it over the network, we do the following:
```

docs/bitvectors.md
```diff
@@ -0,0 +1,214 @@
...
+
+1. Add an extra bit = 1:
+
+```
+1100000011
+```
+
+2. Add padding to complete the full bytes
```
MegaRedHand, line 141:
```suggestion
2. Add padding to complete the full bytes:
```

docs/bitvectors.md
```diff
@@ -0,0 +1,214 @@
...
+11000000 11000000
+```
+
+3. Move to little-endian byte order (reverse bits within each byte):
+
+```
+00000011 00000011
+```
+
+When deserializing, we'll look closely at the last byte, and realize that there's 6 trailing 0s (padding), and discard those and the 7th bit (the sentinel 1).
```
MegaRedHand, line 153:
```suggestion
When deserializing, we'll look closely at the last byte, realize that there are 6 trailing 0s (padding), and discard those along with the 7th bit (the sentinel 1).
```

docs/bitvectors.md
```diff
@@ -0,0 +1,214 @@
...
+11000000
+```
+
+When adding the trailing bit and padding, it will look like this:
+
+```
+11000000 10000000
+```
+
+This means that the sentinel bit is, effectively, adding a full new byte. After reversing the bits:
```
MegaRedHand, line 169:
```suggestion
This means that the sentinel bit is, effectively, adding a new full byte. After reversing the bits:
```

docs/bitvectors.md
```diff
@@ -0,0 +1,214 @@
...
+
+```
+00000011 00000001
+```
+
+When parsing this, we still take care about the last byte, but we will realize that it's comprised of 7 trailing 0s and a sentinel bit, so we'll discard it fully.
```
MegaRedHand, line 175:
```suggestion
When parsing this, we still take care of the last byte, but we will realize that it's comprised of 7 trailing 0s and a sentinel bit, so we'll discard it fully. 
```

docs/bitvectors.md
```diff
@@ -0,0 +1,214 @@
...
+00000011 00000001
+```
+
+When parsing this, we still take care about the last byte, but we will realize that it's comprised of 7 trailing 0s and a sentinel bit, so we'll discard it fully. 
+
+This also shows the importance of the sentinel bit: if it wasn't for it it wouldn't be obvious to the parser that `00000011` represented 8 elements: it could be a set of two validators where both voted.
+
+### Internal representation
+
+For bitlists, in this client we do the same as with bitvectors. We represent them using big endian, for the same reasons. That is, the first thing we do is reverse the bytes, and then remove the first zeroes of the first byte. The code doing that is the following:
```
MegaRedHand, line 181:
```suggestion
For bitlists, in this client we do the same as with bitvectors, and for the same reasons: we represent them using big-endian. That is, the first thing we do is reverse the bytes, and then remove the first zeroes of the first byte. The code doing that is the following:
```

docs/bitvectors.md
```diff
@@ -0,0 +1,214 @@
...
+defp remove_trailing_bit(<<0::7, 1::1>>), do: <<0::0>>
+defp remove_trailing_bit(<<0::8>>), do: <<0::0>>
+
+# This last case should never happen, the last byte should always
+# have a sentinel bit.
+```
+
+We see that we perform two things at the same time:
+
+1. We read as a little endian and then represent it as big endian.
```
MegaRedHand, line 213:
```suggestion
1. We read as a little-endian and then represent it as big-endian.
```

docs/bitvectors.md
```diff
@@ -0,0 +1,237 @@
...
+
+In this representation, we reverse the bytes around. 259 is represented as follows:
+
+```
+00000011 00000001
+```
+
+Representing it as a byte array, we get `bytes = [3, 1]`. The lowest index, `bytes[0] = 3` means the lowest significant byte, and the highest index, `bytes[1] = 1` is the most significant byte. So, while little-endian is less readable, it is frequently used to represent integers as binaries because of this property.
+
+### Little-endian bit order
```
MegaRedHand, line 45: We should put this in quotes or something, since it isn't a standardized name

MegaRedHand (approved):
LGTM. It has some typos, but we can fix that in another PR

PR #735, opened 2024-02-08
feat: add flag for logging to a file

### LINKED ISSUE #652: Add CLI option for logging to a file
Currently, we have some commented code in `config/config.exs` that ideally should be configured at runtime via a CLI flag

MegaRedHand:
Closes #652


config/runtime.exs
```diff
@@ -86,3 +87,37 @@ block_time_ms =
...
+  config :logger, :default_formatter,
+    format: {LogfmtEx, :format},
+    colors: [enabled: false],
+    metadata: [:mfa]
+
+  config :logfmt_ex, :opts,
+    message_key: "msg",
+    timestamp_key: "ts",
+    timestamp_format: :iso8601
+end
```
Arkenan, line 123: Maybe this would be tidier? 

```elixir

case Keyword.get(args, :log_file) do
  nil ->
    config :logger, :default_formatter, format: {ConsoleLogger, :format}, metadata: [:slot, :root]
  log_file ->
   #...
end
```

Arkenan (approved):

config/runtime.exs
MegaRedHand: Good idea!

PR #740, opened 2024-02-11
feat: enable ssz static deposit data container using new ssz_ex library

### LINKED ISSUE #739: [SSZ] DepositData

f3r10:
**Motivation**

Closes #739


test/spec/runners/ssz_static.ex
```diff
@@ -40,47 +78,57 @@ defmodule SszStaticTestRunner do
     expected =
       YamlElixir.read_from_file!(case_dir <> "/value.yaml")
       |> SpecTestUtils.sanitize_yaml()
+      |> sanitize(schema)
```
Arkenan, line 81: Let's move both the `sanitize` function and the `parse_type` to a spec test utils. We shouldn't be duplicating this if we already have it in the `ssz_generic` tests.

test/spec/runners/ssz_static.ex
```diff
@@ -40,47 +78,57 @@ defmodule SszStaticTestRunner do
...
-  defp to_struct_checked(_actual, expected), do: expected
+  def sanitize(vector_elements, {:vector, :bool, _size} = _schema), do: vector_elements
 
-  defp parse_type(%SpecTestCase{handler: handler}) do
-    Module.concat(Types, handler)
-  end
+  def sanitize(vector_elements, {:vector, module, _size} = _schema) when is_atom(module),
+    do:
+      vector_elements
+      |> Enum.map(&struct!(module, &1))
```
Arkenan, line 117: Don't use the pipe operator for just one operation, it makes it less readable. It's easier to just write the following:

```elixir
Enum.map(vector_elements, &struct!(module, &1))
```

test/spec/runners/ssz_static.ex
```diff
@@ -40,47 +78,57 @@ defmodule SszStaticTestRunner do
...
+  def sanitize(bitlist, {:bitlist, _size} = _schema), do: elem(BitList.new(bitlist), 0)
+  def sanitize(bitvector, {:bitvector, size} = _schema), do: BitVector.new(bitvector, size)
+
+  def sanitize(bytelist, {:list, {:int, 8}, _size} = _schema)
+      when is_integer(bytelist) and bytelist > 0,
+      do: :binary.encode_unsigned(bytelist) |> :binary.bin_to_list()
+
+  def sanitize(bytelist, {:list, {:int, 8}, _size} = _schema)
+      when is_integer(bytelist) and bytelist == 0,
+      do: []
```
Arkenan, line 128: Let's have this clause be the first and be matched against the value directly. That is:

```elixir
def sanitize(0, {:list, {:int, 8}, _size} = _schema), do: []
```

If you're matching against a value you don't need to match against its type, it's redundant. That also allows you to remove the `and bytelist > 0` from the other clause (that will now be the second).

test/spec/runners/ssz_static.ex
```diff
@@ -40,47 +78,57 @@ defmodule SszStaticTestRunner do
...
+
+  def sanitize(bytelist, {:list, {:int, 8}, _size} = _schema),
+    do: :binary.bin_to_list(bytelist)
```
Arkenan, line 131: When is this clause called? Can you add it in a comment?

Arkenan (changes requested):
Thanks for the PR Fernando! Have you checked if there are any other containers that pass this test? You can try if there are others by:

1. Commenting the `skip?` content (line 66) so every container is tested
2. Running the test with `--trace` so you see which cases succeed with their name.

If the only one passing is deposit data, that's perfectly fine, I just want to make sure it's the only one.

test/spec/runners/ssz_static.ex
f3r10: this clause is called when an element of the container is a bytelist that is parsed as a binary and not as a list of bytes

Arkenan (approved):
LGTM

PR #741, opened 2024-02-12
refactor: removed all the froms and made the pid maps

### LINKED ISSUE #665: [LibP2P refactor] Delete "from" pids
Currently, when subscribing to a topic, we send `self` over to our go implementation, and then get it back in the notifications. That `self` is the `Subscriber`, which is a genstage that handles the gossip notifications. The problems are:

1. Sending this in each notification is an unnecessary overhead. We can have this configured for gossip topic, as it doesn't vary per-message.
2. This means that we can't transition to a per-message process. It assumes there's only a single process receiving the messages (per topic).

karasakalmt:
Closes #665


Arkenan:
We'll be going a different route with this change. Thanks for taking care of this!

PR #760, opened 2024-02-15
feat: pass the rest of containers ssz spec static tests

### LINKED ISSUE #724: [SSZ] Enable all SSZ spec tests with our SSZEx library
- [x] #739
- [ ] DepositMessage
- [ ] Eth1Block
- [ ] Eth1Data
- [ ] ExecutionPayload
- [ ] ExecutionPayloadHeader
- [ ] Fork
- [ ] ForkData
- [ ] HistoricalBatch
- [ ] IndexedAttestation
- [ ] LightClientBootstrap
- [ ] LightClientOptimisticUpdate
- [ ] LightClientUpdate
- [ ] PendingAttestation
- [ ] PowBlock
- [ ] ProposerSlashing
- [ ] SignedBeaconBlock
- [ ] SignedBeaconBlockHeader
- [ ] SignedContributionAndProof
- [ ] SignedVoluntaryExit
- [ ] SigningData
- [ ] SyncAggregate
- [ ] SyncAggregatorSelectionData
- [ ] SyncCommittee
- [ ] SyncCommitteeContribution
- [ ] SyncCommitteeMessage
- [ ] Validator
- [ ] VoluntaryExit

This is a list of non-priority ones, it's important to not get stuck on those, although ideally we want them to work too:

- ContributionAndProof
- Eth1Block
- LightClientBootstrap
- LightClientFinalityUpdate
- LightClientHeader
- LightClientOptimisticUpdate
- LightClientUpdate
- PowBlock
- SignedContributionAndProof
- SyncAggregatorSelectionData
- SyncCommitteeContribution
- SyncCommitteeMessage

f3r10:
**Motivation**

Closes #724


lib/types/beacon_chain/execution_payload.ex
```diff
@@ -32,14 +32,14 @@ defmodule Types.ExecutionPayload do
...
-          logs_bloom: binary(),
+          logs_bloom: list(Types.uint8()),
           prev_randao: Types.bytes32(),
           block_number: Types.uint64(),
           gas_limit: Types.uint64(),
           gas_used: Types.uint64(),
           timestamp: Types.uint64(),
           # size MAX_EXTRA_DATA_BYTES
-          extra_data: binary(),
+          extra_data: list(Types.bytes1()),
```
mpaulucci, line 42: I did a test with a mainnet block and this part is not working it seems. I get

```
** (EXIT from #PID<0.98.0>) an exception was raised:
    ** (FunctionClauseError) no function clause matching in LambdaEthereumConsensus.SszEx.encode_fixed_size_list/3
        (lambda_ethereum_consensus 0.1.0) lib/ssz_ex.ex:342: LambdaEthereumConsensus.SszEx.encode_fixed_size_list("@penguinbuild.org", {:int, 8}, 32)
        (elixir 1.15.6) lib/enum.ex:1693: Enum."-map/2-lists^map/1-1-"/2
        (lambda_ethereum_consensus 0.1.0) lib/ssz_ex.ex:602: LambdaEthereumConsensus.SszEx.encode_schemas/1
        (lambda_ethereum_consensus 0.1.0) lib/ssz_ex.ex:571: LambdaEthereumConsensus.SszEx.encode_container/2
        (elixir 1.15.6) lib/enum.ex:1693: Enum."-map/2-lists^map/1-1-"/2
        (elixir 1.15.6) lib/enum.ex:1693: Enum."-map/2-lists^map/1-1-"/2
        (lambda_ethereum_consensus 0.1.0) lib/ssz_ex.ex:602: LambdaEthereumConsensus.SszEx.encode_schemas/1
        (lambda_ethereum_consensus 0.1.0) lib/ssz_ex.ex:571: LambdaEthereumConsensus.SszEx.encode_container/2
```

Weird this hasn't been caught by the spec test?

mpaulucci: Here is the branch, https://github.com/lambdaclass/lambda_ethereum_consensus/compare/ssz-ex-bench?expand=1

you can run it with `mix run --no-start bench/ssz_ex_bench.exs`

test/spec/utils.ex
```diff
@@ -135,16 +141,35 @@ defmodule SpecTestUtils do
   def sanitize_ssz(bitlist, {:bitlist, _size} = _schema), do: elem(BitList.new(bitlist), 0)
   def sanitize_ssz(bitvector, {:bitvector, size} = _schema), do: BitVector.new(bitvector, size)
 
+  def sanitize_ssz(
```
mpaulucci, line 144: careful with this "sanitation", it's probably producing unintended bugs

We need to ensure that `SszEx.decode(SszEx.encode(container)) == container` for every container.

test/fixtures/block.ex
```diff
@@ -66,13 +66,13 @@ defmodule Fixtures.Block do
...
       receipts_root: Random.root(),
-      logs_bloom: Random.binary(256),
+      logs_bloom: :binary.bin_to_list(Random.binary(256)),
       prev_randao: Random.hash32(),
       block_number: Random.uint64(),
       gas_limit: Random.uint64(),
       gas_used: Random.uint64(),
       timestamp: Random.uint64(),
-      extra_data: Random.binary(30),
+      extra_data: :binary.bin_to_list(Random.binary(30)),
```
mpaulucci, line 75: not sure about this change, I think we want this to be a binary (a string) in elixir

native/ssz_nif/src/elx_types/beacon_chain.rs
```diff
@@ -302,13 +302,13 @@ gen_struct_with_config!(
         fee_recipient: ExecutionAddress<'a>,
```
mpaulucci, line 302: we probably want to rollback the changes in this file

test/fixtures/block.ex
f3r10: I think the problem with using only binaries is that it is not possible to control the size of that field.
For example, according to SSZ specs `logs_bloom` has to be a fixed vector with a default size of 256 bytes.
On the other hand, `extra_data` is a variable list (a byte list) with a maximum of 32 bytes.
I am not sure how it would be possible to make a difference between those types only using binaries :thinking:

f3r10: one possible solution would be to add these two extra functions in the library:
```defp decode_fixed_list(binary, {:int, 8}, size)```
```defp encode_fixed_size_list(binary, {:int, 8}, _size)```
That would keep the binary and it would not be necessary to change it to a list of bytes. Also, some sanitize_ssz would be not necessary anymore. 
What do you think @mpaulucci @Arkenan ?

Comment (f3r10) [reply test/fixtures/block.ex:75]: that same would apply to a vector of bytes like `logs_bloom`

Comment (f3r10) [reply test/fixtures/block.ex:75]: although, this solution breaks other types list `current_epoch_participation` that is indeed a list of u8 :sweat_smile: 
I think it would be necessary to add additional types to the schemas -> 
:byte_list and :vector_list

Comment (MegaRedHand) [reply test/fixtures/block.ex:75]: Can't we just assume that `(binary, {:list, :uint8, size})` is a byte list (and the same for byte vectors)?

Comment (mpaulucci) [reply test/fixtures/block.ex:75]: so I think we should have `list of byte`, and `list of u8`. One is <<1, 2, 3 >> in Elixir, the other one [1, 2, 3]. They both end up being the same when serialized but the mapping is different when they unserialize. Those this make sense?

https://github.com/ethereum/consensus-specs/blob/dev/ssz/simple-serialize.md#basic-types

Comment (mpaulucci) [reply test/fixtures/block.ex:75]: The size of the field is controlled by the `schema` definition, when they are an Elixir map both a vector and a list would of bytes would look like <<1, 2, 3>>.

Comment (f3r10) [reply test/fixtures/block.ex]: Ok, I think it is done now.

File: test/spec/runners/ssz_static.ex [hunk 5]
```diff
@@ -84,14 +84,15 @@ defmodule SszStaticTestRunner do
     assert_ssz(schema, decompressed, expected, expected_root)
   end
 
-  defp assert_ssz(schema, real_serialized, real_deserialized, expected_root) do
+  defp assert_ssz(schema, real_serialized, real_deserialized, _expected_root) do
     {:ok, deserialized} = SszEx.decode(real_serialized, schema)
```
mpaulucci, line 88: Can we encode both with Ssz and Ssz and assert that both results are equal? For encoding and also for decoding?

We should make sure that we can switch one implementation for the other.

lib/types/beacon_chain/execution_payload.ex
f3r10: @mpaulucci I pulled the latest changes to the new branch `ssz-ex-bench` and the bench works.

f3r10: The encoding is also slower :disappointed:

test/spec/runners/ssz_static.ex
f3r10: it looks like there are discrepancies between bit_list and bit_vectors. The new library already decodes those elements.
For example:
```%Types.Attestation{aggregation_bits: <<0::size(2)>> ...```
and the rust implementation
```%Types.Attestation{aggregation_bits: <<4>> ....```

Comment (f3r10) [reply test/spec/runners/ssz_static.ex:88]: oh I think the problem is the `sanitize_ssz` function :thinking:

Comment (f3r10) [reply test/spec/runners/ssz_static.ex:88]: I am going to create a function that sanitizes the bit_list and bit_vector fields from the rust decode part

Comment (f3r10) [reply test/spec/runners/ssz_static.ex:98]: Ok, I think that is done. Comparing the `decode` results was tricky because of bit_list and bit_vectors and the Aja.Vector fields. 
The `encoding` part worked right away. Let me know if the process done for the decode part is ok.

File: test/spec/runners/ssz_static.ex [hunk 6]
```diff
@@ -73,28 +75,102 @@ defmodule SszStaticTestRunner do
...
+    |> Enum.reverse()
+  end
+
+  defp convert(other), do: other
+
   defp parse_type(%SpecTestCase{handler: handler}) do
     Module.concat(Types, handler)
   end
+
+  defp to_struct_checked(actual, expected) when is_list(actual) and is_list(expected) do
```
Arkenan, line 157: What does this function do exactly? The name is not very clear. We might want to add it in a comment. From what I'm guessing:

- If it's a vec or a list it just follows recursively.
- If it's a base element, it just discards `actual` and takes from `expected`.
- When receiving a struct and a map it uses the struct name of `actual` but the data from `expected`, effectively discarding `actual`.

If this is the case, it means that `real_deserialized` in the calling function is actually the parsed value from the yaml with the struct names of the ssz rust library? If that's the case, why do we do this? We are not comparing the deserialized value from both libraries, we are doing extremely complex processing to compare the new deserialized value with the parsed value again. Shouldn't we take the actual values from `old_deserialized` and try to make the format compatible with the new one to compare instead?

Arkenan (changes requested):
Hi Fernando, thank you very much for working on this! I asked some questions about why the decoded structs are different. We should make sure we're all on the same page about that, as we'll not be able to integrate if we don't solve that.

test/spec/runners/ssz_static.ex
f3r10: this method `to_struct_checked`  was already present with the NIF Ssz Rust :sweat_smile:  I removed it when I started to migrate the spec test to the Elixir Lib. Then, I added it again to try to compare the result of the two libraries

f3r10: From what I understand this method is similar to `sanitize_ssz` because it converts maps to the corresponding structs and also tries to fix some compatibilities between list and the Aja library.

test/spec/runners/ssz_static.ex
```diff
@@ -73,28 +75,102 @@ defmodule SszStaticTestRunner do
...
   end
 
-  defp assert_ssz(schema, real_serialized, real_deserialized, expected_root) do
+  defp assert_ssz(
+         schema,
+         real_serialized,
+         real_deserialized_sanitized,
+         real_deserialized,
+         _expected_root
+       ) do
```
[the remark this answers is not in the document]
f3r10: the difference between `parsed_for_ssz_ex`  and `parsed_for_nif` is that `parsed_for_ssz_ex` goes through `sanitize_ssz` trying to fix the incompatibilities that the yaml decoding does.

lib/ssz_ex.ex
```diff
@@ -797,8 +832,10 @@ defmodule LambdaEthereumConsensus.SszEx do
   defp get_fixed_size(:bool), do: 1
   defp get_fixed_size({:int, size}), do: div(size, @bits_per_byte)
   defp get_fixed_size({:bytes, size}), do: size
+  defp get_fixed_size({:vector, :bytes, size}), do: size
   defp get_fixed_size({:vector, basic_type, size}), do: size * get_fixed_size(basic_type)
-  defp get_fixed_size({:bitvector, _}), do: 1
+
+  defp get_fixed_size({:bitvector, size}), do: ceil(size / 8)
```
MegaRedHand, line 838: We should use `div(size + 7, 8)` to avoid using floats.

lib/types/beacon_chain/beacon_state.ex
```diff
@@ -131,6 +131,19 @@ defmodule Types.BeaconState do
     end)
   end
 
+  def decode_ex(%__MODULE__{} = map) do
+    map
+    |> Map.update!(:validators, &Aja.Vector.new/1)
+    |> Map.update!(:balances, &Aja.Vector.new/1)
+    |> Map.update!(:randao_mixes, &Aja.Vector.new/1)
+    |> Map.update!(:previous_epoch_participation, &Aja.Vector.new/1)
+    |> Map.update!(:current_epoch_participation, &Aja.Vector.new/1)
```
Arkenan, line 140: Instead of having these here, shouldn't we be able to match against `vec` in the ssz_ex file instead?

lib/types/beacon_chain/beacon_state.ex
```diff
@@ -131,6 +131,19 @@ defmodule Types.BeaconState do
...
+    |> Map.update!(:latest_execution_payload_header, &Types.ExecutionPayloadHeader.decode/1)
+    |> Map.update!(:justification_bits, fn bits ->
+      BitVector.new(bits, Constants.justification_bits_length())
```
Arkenan, line 143: Why do we need this? Isn't this already covered in ssz_ex?

lib/types/beacon_chain/beacon_state.ex
```diff
@@ -131,6 +131,19 @@ defmodule Types.BeaconState do
...
   end
 
+  def decode_ex(%__MODULE__{} = map) do
+    map
+    |> Map.update!(:validators, &Aja.Vector.new/1)
+    |> Map.update!(:balances, &Aja.Vector.new/1)
+    |> Map.update!(:randao_mixes, &Aja.Vector.new/1)
+    |> Map.update!(:previous_epoch_participation, &Aja.Vector.new/1)
+    |> Map.update!(:current_epoch_participation, &Aja.Vector.new/1)
+    |> Map.update!(:latest_execution_payload_header, &Types.ExecutionPayloadHeader.decode/1)
```
Arkenan, line 141: Isn't this covered by ssz_ex as well? This is just normal decoding + special decoding of the u256. If u256 is a special case it should be covered in ssz_ex.

lib/types/beacon_chain/beacon_state.ex
f3r10: This is the decoding part. So the result after decoding is not an `Aja.Vector` but a simple `list`.
In this case, for the `beacon_state` a decoded `list` should be converted into an `Aja.Vector`

lib/types/beacon_chain/beacon_state.ex
f3r10: yes

test/spec/utils.ex
```diff
@@ -119,16 +120,17 @@ defmodule SpecTestUtils do
     end
   end
 
+  def sanitize_ssz(vec(_) = other, {:list, module, _size} = _schema) when is_atom(module),
```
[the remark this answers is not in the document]
f3r10: the match of the `Aja.Vector`  is recognized as a `map` and falls right on the next clause of the container `def sanitize_ssz(container, module) when is_map(container)`

test/spec/utils.ex
```diff
@@ -119,16 +120,17 @@ defmodule SpecTestUtils do
...
+    do: Aja.Vector.map(other, fn elem -> struct!(module, elem) end)
+
+  def sanitize_ssz(vec(_) = other, _schema), do: other
```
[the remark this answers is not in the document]
f3r10: the same as the last problem with the pattern matching of the `vec(_) = vector` of the last comment

Arkenan (approved):
LGTM

PR #762, opened 2024-02-15
feat: enable ssz static merklelization

Godspower-Eze:
Enables merklelization for the enabled ssz static spec tests


MegaRedHand (approved):
🔥 🔥 🔥

PR #763, opened 2024-02-15
fix: make spec test generation depend on selected version

MegaRedHand:
I was having some problems when changing the version listed in `.spectest_version`. This PR fixes that.


mpaulucci (approved):
nice!

PR #766, opened 2024-02-15
fix: use `tmp/` for test DBs

MegaRedHand:
When running `make spec-test` it sometimes clashes with db files, trying to load them as test files, while failing to run. This fixes that by using a different folder for the DBs


mpaulucci (approved):
do you want to add that to .gitignore?

PR #768, opened 2024-02-15
chore: add Deneb configs

MegaRedHand:
This PR adds the new configuration and presets for the latest hard-fork, Deneb. Since no mainnet config was modified, we do not need any further changes. However, `MIN_PER_EPOCH_CHURN_LIMIT` was lowered to 2 in minimal, which conflicts with our current spectest version. We need to update our code and spec-tests to the latest version (I'm leaving this for another PR).


mpaulucci (approved):
nice!

PR #771, opened 2024-02-15
ci: add spellchecker

MegaRedHand:


MegaRedHand:
Seems like too much of a chore to maintain... We could add something like this in the future, but for now, I'm closing this PR.

PR #779, opened 2024-02-19
feat: add `ExecutionPayload`, `BeaconBlock`, and related Deneb containers

### LINKED ISSUE #747: Update SSZ containers for deneb
See:
https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/beacon-chain.md#containers
https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/fork-choice.md#containers
https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/p2p-interface.md#containers

Containers to be modified: 
- `BeaconBlockBody`
- `ExecutionPayload`
- `ExecutionPayloadHeader`
- `PayloadAttributes`

We could append these new types with `Deneb`, like `BeaconBlockBodyDeneb` to avoid conflicting with the Cappella ones, which will be deprecated eventually.

Containers to be added:
- `BlobSidecar`
- `BlobIdentifier`

Acceptance criteria:
- Avoid breaking the Cappella SSZ static spec tests.
- Include the Deneb SSZ static spec tests.

MegaRedHand:
Closes #747


lib/types/beacon_chain/execution_payload_deneb.ex
```diff
@@ -0,0 +1,85 @@
...
+    :gas_limit,
+    :gas_used,
+    :timestamp,
+    :extra_data,
+    :base_fee_per_gas,
+    :block_hash,
+    :transactions,
+    :withdrawals,
+    :blob_gas_used,
+    :excess_blob_gas
```
Arkenan, line 25: We need to add the new fields in `docs/specs/beacon-chain.md`.

Arkenan (approved):
LGTM, although open debate for a different PR. What do you think about, instead of having "Types.ExecutionPayload" and "Types.ExecutionPayloadDeneb", we can have "Types.Capella.ExecutionPayload" and "Types.Deneb.ExecutionPayload". This would make it simpler to change the code (only change aliases instead of changing the names everywhere, plus the changes in the use cases).

lib/types/beacon_chain/execution_payload_deneb.ex
MegaRedHand: Should we add those now or when we update the whole docs to the latest fork? I opened #775 for that

Arkenan: Ah, I didn't know that was pending, we can leave it for then.

PR #780, opened 2024-02-19
feat: add NIF to use KZG functions

### LINKED ISSUE #746: Add NIF to use KZG functions
See the Ethereum C implementation here: https://github.com/ethereum/c-kzg-4844/tree/main

We could either use the Rust bindings https://github.com/ethereum/c-kzg-4844/blob/main/bindings/rust/README.md with `rustler` or the C implementation.

The goal is to be able to make these tests pass: https://github.com/ethereum/consensus-specs/tree/dev/tests/formats/kzg

Functions to implement:
- `blob_to_kzg_commitment`
- `compute_kzg_proof`
- `compute_blob_kzg_proof`
- `verify_blob_kzg_proof`
- `verify_blob_kzg_proof_batch`

It might make sense to divide this ticket into one per function. Keeping it like this for now, but feel free to change it.

Godspower-Eze:
Closes #746


lib/types/mod.ex
```diff
@@ -43,4 +43,8 @@ defmodule Types do
   @type execution_address :: bytes20
   @type withdrawal_index :: uint64
   @type payload_id :: bytes8
+  @type kzg_commitment :: bytes48
+  @type kzg_proof :: bytes48
+  ## 131072 is FIELD_ELEMENTS_PER_BLOB(i.e 4096) * BYTES_PER_FIELD_ELEMENT(i.e 32)
+  @type blob :: <<_::131_072>>
```
MegaRedHand, line 49: This varies between chains, so we should leave it as `binary()`. Also, let's move all three to `Kzg` and try marking them as `@opaque`.

native/kzg_nif/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[lib]
+name = "kzg_nif"
+path = "src/lib.rs"
+crate-type = ["cdylib"]
+
+[dependencies]
+rustler = "0.31.0"
+c-kzg = { git = "https://github.com/ethereum/c-kzg-4844", rev = "748283cced543c486145d5f3f38684becdfe3e1b"}
```
MegaRedHand, line 14: Can we use a tag instead of the commit hash? The latest one is `v0.4.2`

native/kzg_nif/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "kzg_nif"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
MegaRedHand, line 6:
```suggestion
```

native/kzg_nif/src/lib.rs
```diff
@@ -0,0 +1,182 @@
...
+    // This cannot fail because bin size equals bytes len
+    binary.as_mut_slice().write_all(bytes).unwrap();
+    binary.into()
+}
+
+fn get_trusted_setup() -> KzgSettings {
+    let lib_path = std::env::current_dir().unwrap().join(file!());
+    let src_path = lib_path.parent().unwrap();
+    let main_path = src_path.parent().unwrap();
+    let trusted_setup_path = main_path.join("native/kzg_nif/official_trusted_setup.txt");
```
MegaRedHand, line 19: Does this file depend on the network (mainnet, sepolia, gnosis)?

native/kzg_nif/src/lib.rs
```diff
@@ -0,0 +1,182 @@
...
+    let trusted_setup_file: &Path = trusted_setup_path.as_path();
+    assert!(trusted_setup_file.exists());
```
MegaRedHand, line 21: We should avoid adding this on release builds by using `debug_assert!`. Also, let's add a comment to make the error's source clearer.
```suggestion
    debug_assert!(trusted_setup_file.exists(), "Missing trusted setup file");
```

native/kzg_nif/src/lib.rs
```diff
@@ -0,0 +1,182 @@
...
+#[rustler::nif]     
+fn blob_to_kzg_commitment<'env>(
+    env: Env<'env>,
+    blob: Binary,
+) -> Result<Binary<'env>, String> {
+    let blob = match Blob::from_bytes(blob.as_slice()) {
+        Ok(blob) => blob,
+        Err(e) => return Err(format!("{:?}", e))
+    };
+    let trusted_setup = get_trusted_setup();
```
MegaRedHand, line 34: We should find a way to initialize the trusted setup only once.

MegaRedHand (commented):
Good work! Left some minor comments.

lib/types/mod.ex
Godspower-Eze: Can you please reference where this is stated(that the size of a blob varies across chains)?

MegaRedHand: Since it's part of the preset, it can change between chains. However both minimal and mainnet use the same number, so we can leave it hardcoded for now.

native/kzg_nif/src/lib.rs
MegaRedHand: I opened #848 for this

MegaRedHand (approved):
Great work! 🚀

PR #785, opened 2024-02-20
feat: add bitlists and bitvectors natively in SSZ nif library

### LINKED ISSUE #784: Add bitlists and bitvectors to the ssz nif library
### Goal

The goal would be that the end result is the same so switching between them is easier. 

### Current situation

Currently the SszEx and the Ssz (rust nif) libraries encode bitlist and bitvector fields differently. 

- SszEx is more advanced: it already uses our bitlists and bitvectors with their opaque types and full operation support
- Ssz contains plain binaries, so we have to convert the fields to those types, operate and then turn them back to binaries. 
 
We should be able to make Ssz adapt to this change so that implementations are compatible and easy to change between each other.

Arkenan:
## Changes

- Remove the length from the bitlist internal representation.
- Add encode/decode functions to attestation and pending attestation containers, so that the old library can handle bitlists.
- Add encode/decode functions to beacon state and sync aggregate so the old library can handle bitvectors.
- In all points integrating with bitlists and bitvectors (mostly operations and epoch processing), remove the creation and serialization of bitlists and bitvectors, as they are now part of the encoding and decoding process.

Closes #784


test/unit/ssz_ex_test.exs
```diff
@@ -465,8 +465,8 @@ defmodule Unit.SSZExTest do
 
   test "serialize and deserialize bitlist" do
     encoded_bytes = <<160, 92, 1>>
-    assert {:ok, decoded_bytes} = SszEx.decode(encoded_bytes, {:bitlist, 16})
-    assert {:ok, ^encoded_bytes} = SszEx.encode(decoded_bytes, {:bitlist, 16})
+    assert {:ok, decoded_bytes} = SszEx.decode(encoded_bytes, {:bitlist, 30})
```
MegaRedHand, line 468: Why this change?

MegaRedHand (approved):
LGTM

test/unit/ssz_ex_test.exs
Arkenan: Good finds, thanks for checking! I undid this change, It wasn't necessary. It was something that I was testing manually and forgot to change back.

PR #786, opened 2024-02-20
refactor: simplify `PendingBlocks`

MegaRedHand:
This PR:
- avoids checking if a block's parent was processed if we know it's being downloaded
- checks preemptively if a block was already processed when adding the block to `PendingBlocks`
- restricts the `state` type to expected values
- simplifies the code a bit (adding a `has_block?` helper, and removing an `if`)

The first two could potentially solve a message queue overflow we're having with the `Blocks` cache.


Arkenan (approved):
LGTM

PR #820, opened 2024-02-27
feat: implement `get_commitee_assignment` function

MegaRedHand:
This function will be used for computing validator committee assignments.

### DIFF
--- a/lib/lambda_ethereum_consensus/validator/validator.ex
+++ b/lib/lambda_ethereum_consensus/validator/validator.ex
@@ -1,0 +1,52 @@
+defmodule LambdaEthereumConsensus.Validator do
+  @moduledoc """
+  Functions for performing validator duties.
+  """
+  alias LambdaEthereumConsensus.StateTransition.Accessors
+  alias LambdaEthereumConsensus.StateTransition.Misc
+  alias Types.BeaconState
+
+  @doc """
+    Return the committee assignment in the ``epoch`` for ``validator_index``.
+    ``assignment`` returned is a tuple of the following form:
+        * ``assignment[0]`` is the list of validators in the committee
+        * ``assignment[1]`` is the index to which the committee is assigned
+        * ``assignment[2]`` is the slot at which the committee is assigned
+    Return None if no assignment.
+  """
+  @spec get_committee_assignment(BeaconState.t(), Types.epoch(), Types.validator_index()) ::
+          {:ok, {[Types.validator_index()], Types.uint64(), Types.slot()}} | {:error, String.t()}
+  def get_committee_assignment(%BeaconState{} = state, epoch, validator_index) do
+    next_epoch = Accessors.get_current_epoch(state) + 1
+
+    if epoch > next_epoch do
+      {:error, "epoch must be <= next_epoch"}
+    else
+      start_slot = Misc.compute_start_slot_at_epoch(epoch)
+      committee_count_per_slot = Accessors.get_committee_count_per_slot(state, epoch)
+      end_slot = start_slot + ChainSpec.get("SLOTS_PER_EPOCH")
+
+      start_slot..end_slot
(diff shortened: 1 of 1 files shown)


lib/lambda_ethereum_consensus/validator/validator.ex
```diff
@@ -0,0 +1,52 @@
...
+        |> Stream.map(&compute_duties(state, slot, validator_index, &1))
+        |> Enum.reject(&is_nil/1)
+        |> List.first()
+      end)
+      |> Enum.reject(&is_nil/1)
+      |> then(&{:ok, List.first(&1)})
+    end
+  end
+
+  defp compute_duties(state, slot, validator_index, index) do
```
Arkenan, line 41: Let's change `index` for `committee_index` for clarity.

lib/lambda_ethereum_consensus/validator/validator.ex
```diff
@@ -0,0 +1,52 @@
...
+      start_slot = Misc.compute_start_slot_at_epoch(epoch)
+      committee_count_per_slot = Accessors.get_committee_count_per_slot(state, epoch)
+      end_slot = start_slot + ChainSpec.get("SLOTS_PER_EPOCH")
+
+      start_slot..end_slot
+      |> Stream.map(fn slot ->
+        0..(committee_count_per_slot - 1)
+        |> Stream.map(&compute_duties(state, slot, validator_index, &1))
+        |> Enum.reject(&is_nil/1)
+        |> List.first()
```
Arkenan, line 34: Maybe something like this could be useful?

```elixir
def first_not_nil(stream) do
  stream
  |> Stream.reject(&is_nil/1)
  |> Stream.take(1)
  |> Enum.to_list()
  |> List.first()
end
```

My only concern with the current version is that it will compute all duties regardless of if the element is found. The same at a slot level, even if there was a committee found in a slot, it will compute every other slot. In that case, using Stream does not provide a great advantage.

lib/lambda_ethereum_consensus/validator/validator.ex
```diff
@@ -0,0 +1,52 @@
...
+      end)
+      |> Enum.reject(&is_nil/1)
+      |> then(&{:ok, List.first(&1)})
```
Arkenan, line 37: If nothing is found, then is nil returned? That wouldn't be great for the error format. Maybe something like this instead?

```elixir
|> case do
  nil -> {:error, "No assignments found for validator #{validator_index} in epoch #{epoch}"}
  result -> result
end
```

Arkenan (approved):
LGTM with some small improvement comments

lib/lambda_ethereum_consensus/validator/validator.ex
MegaRedHand: Great catch!

lib/lambda_ethereum_consensus/validator/validator.ex
MegaRedHand: This is in line with what the spec uses. A validator having no assignments isn't really an error. Maybe it just wasn't activated yet. And it's easier to differentiate an `{:ok, nil}` from an `{:error, "..."}`. We could change it to a `:not_found`, though. wdyt?

Arkenan: `nil` is ok if it's well documented (e.g. today the type specs don't include `nil` as a possible result). Also, unrelated but I think that `|> Enum.find(& not is_nil(&1))` is equivalent to `first_not_nil`. It might simplify the file.

Arkenan (approved):

lib/lambda_ethereum_consensus/validator/validator.ex
MegaRedHand: Done. Good catch! I'm disappointed dialyzer didn't catch that one...

PR #826, opened 2024-02-29
feat: compute duties for a validator

MegaRedHand:
This PR adds a `GenServer` that computes validator duties and logs them to the console.


lib/lambda_ethereum_consensus/validator/utils.ex
```diff
@@ -0,0 +1,64 @@
...
+        nil
+    end
+  end
+
+  @doc """
+    Compute the correct subnet for an attestation.
+  """
+  @spec compute_subnet_for_attestation(Types.uint64(), Types.slot(), Types.uint64()) ::
+          Types.uint64()
+  def compute_subnet_for_attestation(committees_per_slot, slot, committee_index) do
```
MegaRedHand, line 58: This function isn't used in this PR but will be used in the future. I intended to add the code that subscribes to the subnet in this PR, but it seems we'll need to refactor the code a bit to allow that.

mpaulucci (approved):

Issue #796, opened 2024-02-22
Attester committee assignments lookahead


Arkenan:
Implement `get_committee_assignment(state, epoch, index)`. At the beginning of each epoch, the validator should check to which committee it will be assigned in the next one. It should join the attestation subnets and discover new peers using discovery + ENR.

spec ref: [link](https://github.com/lambdaclass/lambda_ethereum_consensus/blob/main/docs/specs/validator.md#lookahead)

PR #828, opened 2024-03-01
feat: subscribe to necessary attestation subnets

MegaRedHand:
Closes: #796

 This PR adds logic to subscribe to attestation subnets according to the hardcoded validator's duties.


lib/lambda_ethereum_consensus/p2p/gossip/attestation.ex
```diff
@@ -0,0 +1,87 @@
...
+  end
+
+  def leave(subnet_id) do
+    topic = get_topic_name(subnet_id)
+    # TODO: update ENR
+    Libp2pPort.leave_topic(topic)
+    P2P.Metadata.set_attestation_subnet(subnet_id, false)
+  end
+
+  def collect(subnet_id), do: GenServer.call(__MODULE__, {:collect, subnet_id})
```
mpaulucci, line 32: who calls these?

lib/lambda_ethereum_consensus/p2p/gossip/gossipsub.ex
```diff
@@ -43,6 +43,8 @@ defmodule LambdaEthereumConsensus.P2P.GossipSub do
         {Consumer, %{topic: topic, ssz_type: ssz_type, handler: handler}}
       end
 
+    children = [LambdaEthereumConsensus.P2P.Gossip.Attestation] ++ children
```
mpaulucci, line 46: This seems to be quite different to the way we were approaching gossip subscriptions? Is the idea to migrate all to a "separate genserver" format?

lib/lambda_ethereum_consensus/p2p/gossip/attestation.ex
MegaRedHand: No-one yet. I added these while working on subscriptions, but turns out it wasn't needed. We'll need them when implementing the aggregation of attestations, however.

lib/lambda_ethereum_consensus/p2p/gossip/gossipsub.ex
MegaRedHand: I don't think so. In this particular case, we needed to dynamically join and leave topics, including subscribing to a particular subnet.

lib/lambda_ethereum_consensus/p2p/gossip/attestation.ex
MegaRedHand: I removed them

lib/lambda_ethereum_consensus/p2p/gossip/gossipsub.ex
mpaulucci: But I imagine this would be the case for most if not all the topics? For example, when a fork changes then we need to join a new topic based on the new fork digest?

mpaulucci: (not in scope for this PR, just discussing)

mpaulucci (approved):

lib/lambda_ethereum_consensus/p2p/gossip/gossipsub.ex
MegaRedHand: You're right. We'll need to support multiple topics when transitioning to a new hard-fork, but apart from that, there are some topics that remain static, like blocks.

PR #829, opened 2024-03-01
feat: add `--metrics` flag for enabling metrics

### LINKED ISSUE #813: Add flag to enable metrics endpoint

MegaRedHand:
Closes #813

This PR adds a `--metrics` flag that enables metrics and optionally overrides the default port (e.g. `--metrics 9569`).

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -8,31 +8,17 @@ switches = [
   mock_execution: :boolean,
   mode: :string,
   datadir: :string,
-  metrics: :string,
+  metrics: :boolean,
+  metrics_port: :integer,
   log_file: :string
 ]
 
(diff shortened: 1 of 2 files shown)


config/runtime.exs
```diff
@@ -8,15 +8,31 @@ switches = [
   mock_execution: :boolean,
   mode: :string,
   datadir: :string,
+  metrics: :string,
```
mpaulucci, line 11: its counterintuitive for me that it is called metrics but receives a port. Maybe two config options would be clearer?

MegaRedHand: Sure!

mpaulucci (approved):

PR #843, opened 2024-03-04
fix: encode requests with `encode_request`

MegaRedHand:
This PR changes `BlockDownloader` to use `ReqResp.encode_request` instead of `encode_ok` for request encoding and changes the `BeaconBlocksByRootRequest` container, which wasn't correct, for a list schema. It also removes unused SSZ containers from both the Elixir and the NIF's side.


mpaulucci (approved):
nice!

PR #850, opened 2024-03-06
fix: bugs in Libp2pPort and IncomingRequests.Handler

MegaRedHand:
This PR fixes six bugs:
- match on old `BeaconBlocksByRoot` struct
- not matching new possible results in `map_block_result`
- use of old enum value `:unsubscribe` in `Lib2pPort`
- validator not handling `StateTransition.process_slots` results
- BLS library not allowing all-balls as secret key
- default logger not logging any metadata

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -148,7 +148,7 @@ case Keyword.get(args, :log_file) do
     config :logger, :default_formatter,
       format: {LogfmtEx, :format},
       colors: [enabled: false],
-      metadata: [:mfa]
+      metadata: [:mfa, :slot]
 
     config :logfmt_ex, :opts,
       message_key: "msg",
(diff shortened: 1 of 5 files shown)


config/runtime.exs
```diff
@@ -148,7 +148,7 @@ case Keyword.get(args, :log_file) do
     config :logger, :default_formatter,
       format: {LogfmtEx, :format},
       colors: [enabled: false],
-      metadata: [:mfa]
+      metadata: [:mfa, :slot]
```
mpaulucci, line 151: do we want `:root` here too?

MegaRedHand: Yes, but I'm leaving it for another PR. Reason: just one of the two gives enough context, and `:root` is given as a raw binary that's processed by the other formatter, however the default one can't.

PR #853, opened 2024-03-07
ci: run the spectests for deneb hard-fork

MegaRedHand:
This PR adds the deneb spectests to the CI along with an "All spec-tests passed" job to simplify required checks. Before merging this, we'll need to modify them to un-require the old spectests, and require the new job.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -2,8 +2,20 @@
 import Config
 
 # Configure fork
-# Available: :capella, :deneb
-config :lambda_ethereum_consensus, :fork, :capella
+# Available: "capella", "deneb"
+# Used only for testing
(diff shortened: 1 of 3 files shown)


config/config.exs
```diff
@@ -2,8 +2,20 @@
 import Config
 
 # Configure fork
-# Available: :capella, :deneb
-config :lambda_ethereum_consensus, :fork, :capella
+# Available: "capella", "deneb"
+# Used only for testing
+fork_raw = System.get_env("FORK", "capella")
```
mpaulucci, line 7: this is at compile time?

MegaRedHand: Locally, you'd need to recompile manually to change forks.

PR #856, opened 2024-03-08
test: generate spec tests for current fork

mpaulucci:
**Motivation**
We want to generate spec tests for both deneb and capella, but we only want to run related to a specific fork if it was compiled for that fork

**Description**
Created `.fork_version` file to be able to use it as a target to generate spec tests.


MegaRedHand (approved):
LGTM

PR #858, opened 2024-03-11
feat: check enabled tests task

### LINKED ISSUE #799: [Spec tests] Coverage script
This script will not measure the coverage of the code, but the spec test coverage, with a detailed report, either in CLI or html format.

Currently, implemented and skipped tests are clearly output when running make spec-test, but we don't have any measure of unimplemented runners. We should have a script that checks:

- Which test vectors have runners implemented.
- Which test cases are skipped vs pass.

This would allow us to not miss any spec tests because of test runners not being implemented.

Some configs could be:

- Filters:
  - Per config (mainnet, minimal, general).
  - Per fork.
  - Per runner.
- Show only unimplemented.

Arkenan:
Changes:

- `check_enabled_tests` task that generates a report on skipped vs enabled tests.
- New makefile target to run iex in test mode.
- MetaUtils module for code inspection.

Closes #799


test/spec/tasks/check_enabled_tests.ex
```diff
@@ -0,0 +1,30 @@
...
+  require Logger
+
+  alias Spec.MetaUtils
+  @output_dir "reports"
+
+  def run(_args) do
+    # TODO: pretty print.
+    res = MetaUtils.check_enabled() |> inspect(limit: :infinity)
+    File.mkdir_p!(@output_dir)
+    File.write!(filename(), res)
```
MegaRedHand, line 21: I think we should just `IO.puts` this, for simplicity.

Arkenan: Removed the file, added pretty printing.

test/spec/tasks/check_enabled_tests.ex
```diff
@@ -0,0 +1,58 @@
...
+  alias Spec.MetaUtils
+
+  def run(_args) do
+    MetaUtils.check_enabled() |> print(0)
+    :ok
+  end
+
+  defp print({name, report}, spaces) do
+    print_spaced(name, spaces)
+    new_spaces = spaces + String.length(name)
```
MegaRedHand, line 23: To account for long names, I think it's better to use a fixed size here.

```elixir
    new_spaces = spaces + 4  # may be a module attribute
```

test/spec/tasks/check_enabled_tests.ex
```diff
@@ -0,0 +1,58 @@
...
+
+  Run with MIX_ENV=test mix check_enabled_tests
+  The output will be in a reports/spec_test_report.exs file, that can be formatted using mix format.
+  """
+
+  require Logger
+
+  alias Spec.MetaUtils
+
+  def run(_args) do
```
MegaRedHand, line 16: I haven't been able to run this from the CLI (tried with `mix check_enabled_tests`). I noticed it's missing the usual `use Mix.Task` and `@impl Mix.Task` tags, but even with that it doesn't work. Is it working for you?

MegaRedHand (changes requested):

test/spec/tasks/check_enabled_tests.ex
Arkenan: Added 3-space indents + 🟡 emoji for the partial runs. This way, the partial runs are aligned with other titles and lines are not that long.

test/spec/tasks/check_enabled_tests.ex
Arkenan: Fixed. You had to use MIX_ENV=test, but I added the env preference so you can skip the manual env config.

MegaRedHand (approved):
Great work!! 🔥

PR #874, opened 2024-03-12
test: un-skip passing spectests

MegaRedHand:


test/spec/runners/finality.ex
```diff
@@ -19,6 +19,12 @@ defmodule FinalityTestRunner do
     Enum.member?(@disabled_cases, testcase)
   end
 
+  @impl TestRunner
```
mpaulucci, line 22: these are passing for me

test/spec/runners/random.ex
```diff
@@ -30,6 +30,11 @@ defmodule RandomTestRunner do
     Enum.member?(@disabled_cases, testcase)
   end
 
+  def skip?(%SpecTestCase{fork: "deneb", case: _testcase}) do
```
mpaulucci, line 33: these are passing for me

PR #879, opened 2024-03-14
feat: check blob data availability

### LINKED ISSUE #765: Check blob data availability in fork choice `on_block` handler
See:
- https://eips.ethereum.org/EIPS/eip-4844
- https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/fork-choice.md#on_block

Depends on:
- https://github.com/lambdaclass/lambda_ethereum_consensus/issues/746
- Probably https://github.com/lambdaclass/lambda_ethereum_consensus/issues/754

Acceptance criteria:
- Fork choice spec tests should work for Deneb fork that use blobs: https://github.com/ethereum/consensus-specs/blob/dev/tests/formats/fork_choice/README.md#on_block-execution-step

MegaRedHand:
Closes #765

### DIFF
--- a/test/spec/runners/fork_choice.ex
+++ b/test/spec/runners/fork_choice.ex
@@ -8,7 +8,6 @@ defmodule ForkChoiceTestRunner do
 
   alias LambdaEthereumConsensus.ForkChoice.Handlers
   alias LambdaEthereumConsensus.ForkChoice.Helpers
-  alias LambdaEthereumConsensus.SszEx
   alias LambdaEthereumConsensus.Store.BlobDb
   alias LambdaEthereumConsensus.Store.Blocks
   alias Types.BeaconBlock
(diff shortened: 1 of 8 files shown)


test/spec/runners/fork_choice.ex
```diff
@@ -16,21 +17,9 @@ defmodule ForkChoiceTestRunner do
 
   use HardForkAliasInjection
 
-  # TODO: implement blob checks
```
mpaulucci, line 19: 🎉

test/spec/tasks/generate_spec_tests.ex
```diff
@@ -92,6 +92,7 @@ defmodule Mix.Tasks.GenerateSpecTests do
 
   defp generate_case(runner_module, testcase) do
     """
+      @tag :tmp_dir
```
mpaulucci, line 95: whats this for?

MegaRedHand: The [`tmp_dir`](https://hexdocs.pm/ex_unit/main/ExUnit.Case.html#module-tmp-dir) tag tells `ExUnit` it should create a temporary directory for the test and adds its path to the context with `tmp_dir` key. The blob tests reuse the same blocks on different testcases, which causes the tests to fail if some of those testcases aren't supposed to have blobs. That's why we need to nuke the DB after finishing each test.

mpaulucci (approved):

test/spec/tasks/generate_spec_tests.ex
mpaulucci: nice!

PR #880, opened 2024-03-14
feat: set `.fork_version` to "deneb"

MegaRedHand:
Since the transition to deneb was completed, our CI fails in the "Start and stop the node" job. This PR fixes this by moving on to deneb as the default fork.


mpaulucci (approved):
let's go!

Issue #629, opened 2024-01-15
[Observability] Add downtime alerts


Arkenan:

MegaRedHand:
Done in #889

PR #889, opened 2024-03-18
feat: add sentry support

MegaRedHand:
Related to #629

To more easily catch errors happening to our node on the server, this PR adds sentry support (disabled by default).


config/config.exs
```diff
@@ -39,3 +39,18 @@ config :lambda_ethereum_consensus, BeaconApi.Endpoint,
...
+config :sentry,
+  dsn: System.get_env("SENTRY_DSN"),
+  environment_name: Mix.env(),
+  enable_source_code_context: true,
+  root_source_code_paths: [File.cwd!()]
+
+config :lambda_ethereum_consensus, :logger, [
+  {:handler, :sentry_handler, Sentry.LoggerHandler,
+   %{
+     config: %{metadata: [:file, :line]}
```
mpaulucci, line 54: should we add additional metadata like block hash or slot?

MegaRedHand: Sure!

MegaRedHand: I added the slot one (I also found we weren't capturing normal ERROR messages 😅), but the root one doesn't seem to work. I guess since it's a raw binary the handler doesn't know how to format it.

mpaulucci (approved):

PR #892, opened 2024-03-18
feat: construct and publish aggregated attestations

### LINKED ISSUE #839: Aggregated attestation creation
An aggregator chosen by is_aggregator() is also an attester in the committee, and generates an attestation. Let's say that we have a variable `attestation` with the value calculated as specified in #803. Then we also collect attestations throughout the assigned slot in the `collected_attestations` variable.

We need to aggregate the attestations into one, and then construct and AggregateAndProof object.

in pseudoelixir:

```elixir
matching_attestations = Enum.filter(attestations, & &1.attestation_data == attestation.attestation_data) 

aggregation_bits = 
  matching_attestations 
  |> Enum.map(& &1.aggregation_bits) 
  |> BitList.aggregate(length(committee))

aggregate = %Attestation{
  attestation_data: attestation.attestation_data,
  aggregation_bits: aggregation_bits
  signature: bls.Aggregate(Enum.map(matching_attestations, & &1.signature))
}

agg = get_aggregate_and_proof(state, aggregator_index, aggregate, privkey)
sig = get_aggregate_and_proof_signature(state, agg, privkey)
%SignedAggregateAndProof{message: agg, signature: sig}
```

Maybe we can have this helper:

```elixir
module BitList do
  def aggregate(bitlist_collection, size) do
    Enum.reduce(
      bitlist_collection, 
      BitList.default(size), 
      fn bits, agg -> BitList.agg(agg, bits) end)
  end
end
```

spec ref: [link](https://github.com/lambdaclass/lambda_ethereum_consensus/blob/main/docs/specs/validator.md#construct-aggregate)

### LINKED ISSUE #840: Broadcast Attestation Aggregate
After creating the SignedAggregateAndProof it must be sent to the global aggregate channel (`beacon_aggregate_and_proof`) when the second slot interval finishes.

Depends on #812 for testing.

Spec refs: [link](https://github.com/lambdaclass/lambda_ethereum_consensus/blob/main/docs/specs/validator.md#broadcast-aggregate)

MegaRedHand:
Depends on #890
Closes #839
Closes #840

This PR adds the logic of `SignedAggregateAndProof` creation and broadcasting. Also: it adds some typespecs that were missing, and filters incoming attestations leaving only those matching our vision.

### DIFF
--- a/lib/lambda_ethereum_consensus/p2p/gossip/attestation.ex
+++ b/lib/lambda_ethereum_consensus/p2p/gossip/attestation.ex
@@ -17,6 +17,7 @@ defmodule LambdaEthereumConsensus.P2P.Gossip.Attestation do
     GenServer.start_link(__MODULE__, init_arg, name: __MODULE__)
   end
 
+  @spec join(non_neg_integer()) :: :ok
   def join(subnet_id) do
     topic = get_topic_name(subnet_id)
     Libp2pPort.join_topic(topic)
@@ -25,6 +26,7 @@ defmodule LambdaEthereumConsensus.P2P.Gossip.Attestation do
     update_enr()
   end
 
+  @spec leave(non_neg_integer()) :: :ok
   def leave(subnet_id) do
     topic = get_topic_name(subnet_id)
     Libp2pPort.leave_topic(topic)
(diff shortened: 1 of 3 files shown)


mpaulucci (approved):
LGTM

Issue #821, opened 2024-02-27
Implement EIP-2335 compliant keystore


MegaRedHand:
Other validator clients store validator private keys in the format described in [EIP/ERC-2335](https://eips.ethereum.org/EIPS/eip-2335). We should add support for it, including  both JSON de/serialization and private key decryption

PR #897, opened 2024-03-19
feat: load validator keys from file (prototype)

MegaRedHand:
First step towards #821

This PR adds the `--validator-file` flag, that accepts the path to a file with a validator's keys. From those, the validator index is computed (with an appropriate error message if not found).

An example file would be:

```
0xa03c1e287ccc4d457f5e71e9dc769294835945561e6f236ac7de210d2e614eee8a85e21dfb46e2143c68de22ccee8660
0x5274a06d46a9c6ecc83bdea918066e1a3bc80143bb1d99113ddee4192c6efbf1
```
The first line is the pubkey for validator 55 of sepolia, while the second is a random private key (for obvious reasons).

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -11,6 +11,7 @@ switches = [
   testnet_dir: :string,
   metrics: :boolean,
   metrics_port: :integer,
+  validator_file: :string,
   log_file: :string
 ]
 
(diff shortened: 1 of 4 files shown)


mpaulucci (approved):
great work!

Issue #918, opened 2024-03-22
Collect `BLSToExecutionChange` objects and include them in a block


mpaulucci:
Spec: https://github.com/lambdaclass/lambda_ethereum_consensus/blob/main/docs/specs/validator.md#bls-to-execution-changes

Depends on: https://github.com/lambdaclass/lambda_ethereum_consensus/issues/910

PR #920, opened 2024-03-22
feat: collect `BLSToExecutionChange` from gossip

MegaRedHand:
Related to #918

### DIFF
--- a/lib/lambda_ethereum_consensus/fork_choice/fork_choice.ex
+++ b/lib/lambda_ethereum_consensus/fork_choice/fork_choice.ex
@@ -8,6 +8,7 @@ defmodule LambdaEthereumConsensus.ForkChoice do
 
   alias LambdaEthereumConsensus.Beacon.BeaconChain
   alias LambdaEthereumConsensus.ForkChoice.{Handlers, Helpers}
+  alias LambdaEthereumConsensus.P2P.Gossip.OperationsCollector
   alias LambdaEthereumConsensus.Store.Blocks
   alias LambdaEthereumConsensus.Store.StoreDb
   alias LambdaEthereumConsensus.Validator
@@ -166,6 +167,7 @@ defmodule LambdaEthereumConsensus.ForkChoice do
 
     Handlers.notify_forkchoice_update(store, head_block)
 
+    OperationsCollector.notify_new_block(head_block)
     Validator.notify_new_block(head_block.slot, head_root)
 
     BeaconChain.update_fork_choice_cache(
--- a/lib/lambda_ethereum_consensus/p2p/gossip/handler.ex
+++ b/lib/lambda_ethereum_consensus/p2p/gossip/handler.ex
@@ -7,9 +7,17 @@ defmodule LambdaEthereumConsensus.P2P.Gossip.Handler do
 
   alias LambdaEthereumConsensus.Beacon.BeaconChain
   alias LambdaEthereumConsensus.Beacon.PendingBlocks
+  alias LambdaEthereumConsensus.P2P.Gossip.OperationsCollector
   alias LambdaEthereumConsensus.Store.BlobDb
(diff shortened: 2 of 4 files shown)


lib/lambda_ethereum_consensus/fork_choice/fork_choice.ex
```diff
@@ -166,6 +167,7 @@ defmodule LambdaEthereumConsensus.ForkChoice do
 
     Handlers.notify_forkchoice_update(store, head_block)
 
+    OperationsCollector.notify_new_block(head_block)
```
mpaulucci, line 170: nice! Seems this `recompute_head` has grown its responsabilities, basically doing every task that is async. We may want to do a refactor soon (out of scope of this PR)

lib/lambda_ethereum_consensus/p2p/gossip/handler.ex
```diff
@@ -31,7 +39,7 @@ defmodule LambdaEthereumConsensus.P2P.Gossip.Handler do
     root = aggregate.data.beacon_block_root |> Base.encode16()
 
     # We are getting ~500 attestations in half a second. This is overwhelming the store GenServer at the moment.
-    # Store.on_attestation(aggregate)
+    # ForkChoice.on_attestation(aggregate)
```
mpaulucci, line 42: Oh, I didn't know this was still commented out. Created: https://github.com/lambdaclass/lambda_ethereum_consensus/issues/924

lib/lambda_ethereum_consensus/p2p/gossip/operations_collector.ex
```diff
@@ -0,0 +1,49 @@
...
+      operations.bls_to_execution_changes
+      |> MapSet.new(& &1.message.validator_index)
+
+    bls_to_execution_changes =
+      state.bls_to_execution_change
+      |> Enum.reject(&MapSet.member?(indices, &1.message.validator_index))
+
+    %{state | bls_to_execution_change: bls_to_execution_changes}
+  end
+end
```
mpaulucci, line 49: we need a getter here as well for the validator to query, right?

mpaulucci, line 49: I like this design. I guess we want to also be able to remove these as slots/epochs pass by. What do you think of Beacon Chain emiting an event every time a slot (or epoch) passes and genservers that are interesed in slot changes subscribing to that event? (again, probably out of scope of this PR)

lib/lambda_ethereum_consensus/fork_choice/fork_choice.ex
MegaRedHand: 👍 #925

lib/lambda_ethereum_consensus/p2p/gossip/operations_collector.ex
MegaRedHand: I was thinking the `Validator` needs something like that (but for every third of a slot). I think it's something we want to tackle (although maybe after we finish the `Validator`)

mpaulucci (approved):

lib/lambda_ethereum_consensus/p2p/gossip/operations_collector.ex
MegaRedHand: #926

MegaRedHand:
@mpaulucci I'm gonna add the getter before merging

lib/lambda_ethereum_consensus/p2p/gossip/operations_collector.ex
MegaRedHand: > I guess we want to also be able to remove these as slots/epochs pass by.

Actually, this kind of operation should be valid until it's added in a block (or a similar one is).

PR #921, opened 2024-03-22
feat: add block serializer to JSON

### LINKED ISSUE #891: [Beacon API] Get block by ID

avilagaston9:
**Motivation**

This PR adds the **to_json** function to serialize a **SignedBeaconBlock** into JSON format, in order to be used by the BeaconAPI.

Closes #891


lib/beacon_api/controllers/v2/beacon_controller.ex
```diff
@@ -61,14 +63,31 @@ defmodule BeaconApi.V2.BeaconController do
...
+  def to_json(map) when is_map(map) do
+    map
+    |> Map.from_struct()
+    |> Stream.map(&to_json/1)
+    |> Map.new()
+  end
+
+  def to_json(list) when is_list(list), do: Enum.map(list, &to_json/1)
+
+  def to_json({:aggregation_bits, v}) do
```
mpaulucci, line 80: we probably want a generic way of encoding a bitlist instead of having to specify the field (in this case `aggregation_bits`).

mpaulucci: Maybe we want to do something similar to the `encode` function https://github.com/lambdaclass/lambda_ethereum_consensus/blob/main/lib/types/beacon_chain/beacon_state.ex#L152

mpaulucci: Generic/recursive would be ideal!

test/unit/beacon_api/beacon_api_v2_test.exs
```diff
@@ -0,0 +1,78 @@
...
+
+  @opts Router.init([])
+
+  setup do
+    start_supervised!(Blocks)
+    :ok
+  end
+
+  test "get block by id" do
+    start_link_supervised!({Db, dir: "tmp/beacon_api_v2/1"})
```
MegaRedHand, line 21: We could use [the `:tmp_dir` tag functionality from `ExUnit`](https://hexdocs.pm/ex_unit/main/ExUnit.Case.html#module-tmp-dir). Also, we should add this functionality to the original `v1` tests.

test/unit/beacon_api/beacon_api_v1_test.exs
```diff
@@ -80,6 +82,7 @@ defmodule BeaconApiTest do
     assert conn.resp_body == encoded_resp_body_json
   end
 
+  @tag :tmp_dir
```
MegaRedHand, line 85: We can change this for a single `@moduletag :tmp_dir`, if I recall correctly

lib/beacon_api/controllers/v2/beacon_controller.ex
```diff
@@ -61,14 +64,37 @@ defmodule BeaconApi.V2.BeaconController do
...
-      data: %{
-        # TODO: return block as JSON
-        message: inspect(block)
-      }
+      data: to_json(block)
     })
   end
 
-  defp block_not_found(conn) do
+  defp to_json(attribute, module) when is_struct(attribute) do
```
mpaulucci, line 71: you probably want to move this function to a more generic module?

avilagaston9: Could it be, 'lib/beacon_api/utils'?

lib/beacon_api/controllers/v2/beacon_controller.ex
```diff
@@ -61,14 +63,11 @@ defmodule BeaconApi.V2.BeaconController do
...
-      data: %{
-        # TODO: return block as JSON
-        message: inspect(block)
-      }
+      data: Utils.to_json(block)
     })
   end
 
-  defp block_not_found(conn) do
+  def block_not_found(conn) do
```
MegaRedHand, line 70: We can leave this one as `defp`

PR #923, opened 2024-03-22
feat: add release metadata to sentry

### LINKED ISSUE #906: Add `release` metadata to Sentry
https://docs.sentry.io/platforms/elixir/configuration/releases/

We should be adding the commit SHA so we can easily pin point an issue to a release. Also, maybe we can connect Github so we can get the source code related to the error.

MegaRedHand:
Closes #906

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -195,3 +195,11 @@ case Keyword.get(args, :log_file) do
       timestamp_key: "ts",
       timestamp_format: :iso8601
 end
+
+# Sentry
+
+{git_sha, 0} = System.cmd("git", ["rev-parse", "--short", "HEAD"])
+
+config :sentry,
+  dsn: System.get_env("SENTRY_DSN"),
+  release: String.trim(git_sha)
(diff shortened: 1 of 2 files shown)


config/runtime.exs
```diff
@@ -195,3 +195,11 @@ case Keyword.get(args, :log_file) do
       timestamp_key: "ts",
       timestamp_format: :iso8601
 end
+
+# Sentry
+
+{git_sha, 0} = System.cmd("git", ["rev-parse", "--short", "HEAD"])
```
mpaulucci, line 201: You probably want to use this instead? https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables

`GITHUB_SHA`?

MegaRedHand: No. That would only work for GitHub workflows. We want this to work automatically on the server (i.e. without us manually fetching the commit's hash).

mpaulucci: ok, makes sense

Issue #916, opened 2024-03-22
Collect voluntary exits and include them in a block


mpaulucci:
Spec: https://github.com/lambdaclass/lambda_ethereum_consensus/blob/main/docs/specs/validator.md#voluntary-exits

Depends on: https://github.com/lambdaclass/lambda_ethereum_consensus/issues/910

Issue #913, opened 2024-03-22
Collect attester slashings and include them in a block


mpaulucci:
Spec: https://github.com/lambdaclass/lambda_ethereum_consensus/blob/main/docs/specs/validator.md#attester-slashings

Depends on: https://github.com/lambdaclass/lambda_ethereum_consensus/issues/910

PR #927, opened 2024-03-25
feat: collect slashings and voluntary exits

MegaRedHand:
Related to #912 #913 #916

### DIFF
--- a/lib/lambda_ethereum_consensus/p2p/gossip/operations_collector.ex
+++ b/lib/lambda_ethereum_consensus/p2p/gossip/operations_collector.ex
@@ -2,10 +2,13 @@ defmodule LambdaEthereumConsensus.P2P.Gossip.OperationsCollector do
   @moduledoc """
   Module that stores the operations received from gossipsub.
   """
+  use GenServer
+
+  alias Types.AttesterSlashing
   alias Types.BeaconBlock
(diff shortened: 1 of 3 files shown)


lib/lambda_ethereum_consensus/p2p/gossip/operations_collector.ex
```diff
@@ -56,6 +105,30 @@ defmodule LambdaEthereumConsensus.P2P.Gossip.OperationsCollector do
...
+        &MapSet.member?(slashed_proposers, &1.signed_header_1.message.proposer_index)
+      )
+
+    exited = operations.voluntary_exits |> MapSet.new(& &1.message.validator_index)
+
+    voluntary_exits =
+      state.voluntary_exit |> Enum.reject(&MapSet.member?(exited, &1.message.validator_index))
+
+    %{
+      state
```
mpaulucci, line 127: not asking to change this now, but wonder if it would be better to have one GenServer per operation to benefit from parallelization.

mpaulucci (approved):

lib/lambda_ethereum_consensus/p2p/gossip/operations_collector.ex
MegaRedHand: Yeah, probably. It depends on how slow the filtering is.

PR #935, opened 2024-03-26
feat: add block construction boilerplate and calculate some fields

mpaulucci:
**Motivation**
Have the initial boilerplate for block creation to start working over.

**Description**
Initial block construction. It now calculates `randao_reveal` and calculates the signature. 

All the rest is hardcoded for now. I copied the values from the block taken from the spec tests to ensure that the block signature matches.


MegaRedHand (approved):
LGTM

PR #938, opened 2024-03-27
feat: add block proposing workflow

MegaRedHand:


test/unit/validator/proposer_test.exs
```diff
@@ -29,7 +31,11 @@ defmodule Unit.Validator.ProposerTests do
       <<0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
         0, 64>>
 
-    {:ok, signed_block} = Proposer.construct_block(pre_state, privkey)
+    validator_index = 63
+
+    {:ok, state} = StateTransition.process_slots(pre_state, 1)
+
+    {:ok, signed_block} = Proposer.construct_block(state, validator_index, privkey)
```
mpaulucci, line 38: let's call this `state` pre. Does it verify that `state_transition(pre, block)` is valid?

mpaulucci: We might want to add that either to the test or to the `construct_block` function.

lib/lambda_ethereum_consensus/validator/proposer.ex
```diff
@@ -4,14 +4,17 @@ defmodule LambdaEthereumConsensus.Validator.Proposer do
...
-  @spec construct_block(BeaconState.t(), Bls.privkey()) :: {:ok, Types.SignedBeaconBlock.t()}
-  def construct_block(state, privkey) do
+  @spec construct_block(BeaconState.t(), Types.validator_index(), Bls.privkey()) ::
+          {:ok, Types.SignedBeaconBlock.t()}
+  def construct_block(%BeaconState{} = state, proposer_index, privkey) do
+    # NOTE: the state is at the start of the block's slot
     block = %Types.BeaconBlock{
-      slot: 1,
-      proposer_index: 63,
+      slot: state.slot,
```
mpaulucci, line 16: maybe we should add a new argument to the function called `slot` instead of messing with the state?

MegaRedHand: Done

test/unit/validator/proposer_test.exs
MegaRedHand: Done

mpaulucci (approved):
:fire

PR #940, opened 2024-03-27
feat: add `state_root` to proposed block.

mpaulucci:


MegaRedHand (approved):
Left some nits. LGTM otherwise

MegaRedHand:
With this, we're already producing valid blocks, right?

mpaulucci:
We're missing the execution side of things.

PR #946, opened 2024-04-02
feat: allow genesis from file for local testnets

### LINKED ISSUE #804: Support loading custom config files from YAML
Context: https://github.com/lambdaclass/lambda_ethereum_consensus/issues/172#issuecomment-1961920310

This includes loading information from:
- [ ] `config.yaml`
- [ ] `boot_enr.yaml` (#807)
- [ ] `genesis.ssz` (depends on #806)
- [ ] ~`deploy_block.txt`~ (this one's optional)

And also adding a flag `--testnet-dir <DIR>` that loads all this automatically from a directory.

### LINKED ISSUE #806: Load genesis state with custom config
Context: https://github.com/lambdaclass/lambda_ethereum_consensus/issues/172#issuecomment-1961920310

~Blocked by: #805~

The flag `--testnet-dir` should parse the `genesis.ssz` file and load the state contained within.

Arkenan:
**Motivation**

Local testnets have hardcoded genesis state in a ssz file. It needs to be loaded so it can be then used in the beacon node module.

**Description**

- Adds a StoreSetup module to make a strategy (file, url, or db) and then use it when bringing up the node.
- Add runtime config for StoreSetup.
- Add CustomConfig fixes to parse the local testnet yaml correctly (merge the network config, parse integers correctly).

Closes #806
Closes #804


config/runtime.exs
```diff
@@ -41,8 +44,10 @@ validator_file = Keyword.get(args, :validator_file)
 enable_beacon_api = Keyword.get(args, :beacon_api, false)
 beacon_api_port = Keyword.get(args, :beacon_api_port, 4000)
 
-config :lambda_ethereum_consensus, LambdaEthereumConsensus.ForkChoice,
-  checkpoint_sync_url: checkpoint_sync_url
+if not is_nil(testnet_dir) and not is_nil(checkpoint_sync_url) do
+  IO.puts("Both checkpoint sync and testnet url specified (only one should be specified).")
```
mpaulucci, line 48: should we do this validation in another place so that we have access to Logger?

config/runtime.exs
```diff
@@ -88,9 +93,30 @@ chain_config =
...
   end
 
-config :lambda_ethereum_consensus, ChainSpec,
+# We use put_env here as we need this immediately after to read the state.
+Application.put_env(:lambda_ethereum_consensus, ChainSpec,
   config: Map.fetch!(chain_config, :config),
   genesis_validators_root: Map.fetch!(chain_config, :genesis_validators_root)
+)
+
+genesis =
```
mpaulucci, line 102: maybe do this inside Genesis module?

lib/lambda_ethereum_consensus/beacon/genesis.ex
```diff
@@ -0,0 +1,138 @@
+defmodule Genesis do
```
mpaulucci, line 1: nit: LambdaEthereumConsensus.Beacon.Genesis ?

Also maybe the genesis naming is a bit confusing if we're doing checkpoint sync which is not technically a "genesis" state

mpaulucci: What about something like StoreSetup or similar?

lib/ssz_ex.ex
```diff
@@ -231,7 +231,7 @@ defmodule LambdaEthereumConsensus.SszEx do
   end
```
mpaulucci, line 231: remove changes to this file?

lib/chain_spec/configs/gen_config.ex
```diff
@@ -21,6 +21,7 @@ defmodule ChainSpec.GenConfig do
 
       @impl unquote(__MODULE__)
       def get(key), do: Map.fetch!(@__unified, key)
+      def get_all, do: @__unified
```
MegaRedHand, line 24: We should add this to the behavior

lib/lambda_ethereum_consensus/beacon/genesis.ex
Arkenan: Used StoreSetup

lib/ssz_ex.ex
Arkenan: Removed

config/runtime.exs
Arkenan: Hmmm in general, other validations (like the one for mode) in this file that need to make the system crash are of the same style. Where could we move it?

MegaRedHand: I think using `IO.puts` for these input errors is the best way to handle them. We can add the ERROR prefix manually or even the whole log format by invoking the formatter directly.

lib/chain_spec/utils.ex
```diff
@@ -7,7 +7,7 @@ defmodule ConfigUtils do
   def load_config_from_file!(path) do
     path
     |> File.read!()
-    |> String.replace(~r/(0x[0-9a-fA-F]+)/, "'\\g{1}'")
+    |> String.replace(~r/\ (0x[0-9a-fA-F]+)/, " '\\g{1}'")
```
MegaRedHand, line 10: Does the `\` before the space do anything?

lib/lambda_ethereum_consensus/beacon/store_setup.ex
```diff
@@ -0,0 +1,164 @@
...
+  def setup!, do: setup!(get_strategy!())
+
+  @spec setup!(store_setup_strategy()) :: {Store.t(), binary}
+  def setup!({:file, anchor_state}) do
+    Logger.info("[Store Setup] Setting up store from genesis file.")
+
+    anchor_block = %{
+      SszEx.default(SignedBeaconBlock)
+      | message: %{SszEx.default(BeaconBlock) | state_root: Ssz.hash_tree_root!(anchor_state)}
+    }
```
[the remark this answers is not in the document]
MegaRedHand: Update: we're sure

mpaulucci (approved):
It's a bit hard to review this, but LGTM

PR #947, opened 2024-04-02
feat: add default methods for all schemas/ssz types

### LINKED ISSUE #395: Add default values for each SSZ type

Arkenan:
**Motivation**

We need default types for the ssz library, specially for the beacon block, as it's required to initialize the beacon node from a genesis file.

**Description**

Using the [ssz specs](https://github.com/ethereum/consensus-specs/blob/v1.3.0/ssz/simple-serialize.md#default-values) as a reference for default types.

Closes #395


lib/utils/bit_list.ex
```diff
@@ -9,10 +9,16 @@ defmodule LambdaEthereumConsensus.Utils.BitList do
...
-  @spec default(non_neg_integer) :: t
-  def default(size), do: <<0::size(size)>>
+  @spec zero(non_neg_integer) :: t
+  def zero(size), do: <<0::size(size)>>
+
+  @doc """
+  Creates a default (empty/size 0) bit list.
+  """
+  @spec default() :: t
+  def default(), do: <<>>
```
mpaulucci, line 21: The size is 0 instead of `size` because it's a list and not a vector right?

lib/ssz_ex.ex
```diff
@@ -428,6 +428,21 @@ defmodule LambdaEthereumConsensus.SszEx do
...
+  def default({:int, _}), do: 0
+  def default(:bool), do: false
+  def default({:bytes, size}), do: <<0::size(size * 8)>>
+  def default({:list, _, _}), do: []
+  def default({:vector, inner_type, size}), do: default(inner_type) |> List.duplicate(size)
+  def default({:bitlist, _}), do: BitList.default()
+  def default({:bitvector, size}), do: BitVector.new(size)
+
+  def default(module) when is_atom(module) do
+    module.schema() |> Enum.map(fn {_attr, schema} -> default(schema) end)
```
mpaulucci, line 443: maybe we could add a test in ssz static that creates a `default` container and encodes it to ensure it doesn't error?

Or maybe it's a bit overkill and we can do it in ssz generic instead

MegaRedHand: We can add it as a check in `LambdaEthereumConsensus.Container.__after_compile__/2`, just to see if it fails anywhere.

MegaRedHand (approved):
LGTM

PR #956, opened 2024-04-04
refactor: use byte_vector and byte_list instead of bytes, vector-bytes and list-bytes

### LINKED ISSUE #951: Simplify `SszEx` schemas: `:bytes`
We currently use `{:bytes, N}` and `{:byte_list, N}` to encode SSZ byte-vectors and byte-lists in Elixir binaries, but we also allow `{:vector, :bytes, N}` and `{:list, :bytes, N}` for the same purpose. To simplify it, we should replace:
- `{:bytes, N}` and `{:vector, :bytes, N}` -> `{:byte_vector, N}`
- `{:list, :bytes, N}` -> `{:byte_list, N}`

avilagaston9:
Replaces:
- `{:bytes, N}` and `{:vector, :bytes, N}` -> `{:byte_vector, N}`
- `{:list, :bytes, N}` -> `{:byte_list, N}`

Closes #951


lib/ssz_ex.ex
```diff
@@ -17,10 +17,10 @@ defmodule LambdaEthereumConsensus.SszEx do
       Composite type of ordered homogeneous elements with an exact number of elements (`length`).
       `inner_type` is the schema of the inner elements. Depending on it, the vector could be
       variable-size or fixed-size.
-    - `{:byte_list, max_length}` or `{:list, :bytes, max_length}`
+    - `{:byte_list, max_length}`
       Same as `{:list, {:int, 8}, length}` (i.e. a list of bytes), but
       encodes-from/decodes-into an Elixir binary.
-    - `{:bytes, length}` or `{:vector, :bytes, length}`
+    - `{:bytes_vector, length}`
```
MegaRedHand, line 23: Typo:

```suggestion
    - `{:byte_vector, length}`
```

lib/ssz_ex.ex
```diff
@@ -205,9 +203,10 @@ defmodule LambdaEthereumConsensus.SszEx do
   @spec hash_tree_root(non_neg_integer, {:int, non_neg_integer}) :: {:ok, Types.root()}
   def hash_tree_root(value, {:int, size}), do: {:ok, pack(value, {:int, size})}
 
-  @spec hash_tree_root(binary, {:bytes, non_neg_integer}) :: {:ok, Types.root()}
-  def hash_tree_root(value, {:bytes, size}) do
-    packed_chunks = pack(value, {:bytes, size})
+  @spec hash_tree_root(binary, {:byte_list | :byte_vector, non_neg_integer}) ::
+          {:ok, Types.root()}
+  def hash_tree_root(value, {type, _size}) when type in [:byte_list, :byte_vector] do
```
MegaRedHand, line 208: The original code uses `:bytes` only. Wouldn't this need to include a length mix-in for the `:byte_list` case?

lib/types/type_aliases.ex
```diff
@@ -42,5 +44,5 @@ defmodule TypeAliases do
   def blob_sidecars_by_root_request,
     do: {:list, Types.BlobIdentifier, ChainSpec.get("MAX_REQUEST_BLOB_SIDECARS")}
 
-  def error_message, do: {:bytes, 256}
+  def error_message, do: {:byte_vector, 256}
```
MegaRedHand, line 47: This one's actually a `:byte_list`. I mistakenly used `:bytes` because I thought they were the same 😅

MegaRedHand (commented):
Left some comments. Otherwise, 🚀 🚀 🚀

MegaRedHand (approved):
👍

Issue #911, opened 2024-03-22
Calculate `eth1_data` for a block to be proposed


mpaulucci:
Necessary reading: https://eth2book.info/capella/part2/deposits-withdrawals/deposit-processing/
Spec: https://github.com/lambdaclass/lambda_ethereum_consensus/blob/main/docs/specs/validator.md#eth1-data

This required communicating with the Execution Client to fetch the deposit root to cast an `eth1_vote`.

I suggest looking at other client implementations for ideas. This task is non-trivial and one thing that has to be taken into account is how to handle this while the execution client is not synced with the tip of the chain.

Potencially can be divided into further tasks.

MegaRedHand:
Blocked by #949 and #950

PR #960, opened 2024-04-04
feat: compute eth1 data from recent blocks

MegaRedHand:
Related to #911


lib/lambda_ethereum_consensus/execution/eth1_chain.ex
```diff
@@ -0,0 +1,128 @@
+defmodule LambdaEthereumConsensus.Execution.Eth1Chain do
```
mpaulucci, line 1: Maybe we could call it ExecutionChain (to get out of the legacy Eth1 nomenclature) and perform other duties like checking engine capabilities and making sure that we're on the proper chain (out of scope for this PR)

mpaulucci (approved):

lib/lambda_ethereum_consensus/execution/eth1_chain.ex
MegaRedHand: Sure!

PR #961, opened 2024-04-08
fix: hash tree root

### LINKED ISSUE #852: [SSZ]: enable `hash_tree_root` for `ssz_static` spec tests containers

avilagaston9:
This PR will enable hash_tree_root for ssz_static spec tests containers.

Closes #852


lib/ssz_ex.ex
```diff
@@ -238,28 +236,10 @@ defmodule LambdaEthereumConsensus.SszEx do
...
-        leaf_count = chunks |> get_chunks_len() |> next_pow_of_two()
-        root = chunks |> merkleize_chunks_with_virtual_padding(leaf_count)
-        {:ok, root}
-
-      {:error, reason} ->
-        {:error, reason}
-    end
+  def hash_tree_root(vec(_) = list, {:list, _any, _non_neg_integer} = schema) do
+    list
+    |> Aja.Vector.to_list()
```
mpaulucci, line 241: Maybe we can do this as a follow up, but we shouldn't do these kind of things here. We could call this `encode` function (see for example: https://github.com/lambdaclass/lambda_ethereum_consensus/blob/main/lib/types/beacon_chain/beacon_state.ex#L152) before doing the hash tree root in case of a container

lib/types/beacon_chain/execution_payload.ex
```diff
@@ -77,7 +77,7 @@ defmodule Types.ExecutionPayload do
       {:base_fee_per_gas, TypeAliases.uint256()},
       {:block_hash, TypeAliases.hash32()},
       {:transactions, TypeAliases.transactions()},
-      {:withdrawals, {:list, Types.Withdrawal, ChainSpec.get("MAX_TRANSACTIONS_PER_PAYLOAD")}},
+      {:withdrawals, {:list, Types.Withdrawal, ChainSpec.get("MAX_WITHDRAWALS_PER_PAYLOAD")}},
```
mpaulucci, line 80: 👀

mpaulucci (approved):
great work!

PR #963, opened 2024-04-08
feat: fetch deposit logs for eth1 data validation

### LINKED ISSUE #911: Calculate `eth1_data` for a block to be proposed

MegaRedHand:
Closes #911


lib/beacon_api/controllers/v2/beacon_controller.ex
```diff
@@ -36,7 +36,7 @@ defmodule BeaconApi.V2.BeaconController do
 
   def get_block(conn, %{block_id: "0x" <> hex_block_id}) do
     with {:ok, block_root} <- Base.decode16(hex_block_id, case: :mixed),
-         block <- Blocks.get_signed_block(block_root) do
+         %{} = block <- Blocks.get_signed_block(block_root) do
```
mpaulucci, line 39: ?

mpaulucci (approved):
great work!

lib/beacon_api/controllers/v2/beacon_controller.ex
MegaRedHand: We found this with @avilagaston9. The result can be either a map or `nil`, but we aren't matching on the map. This can result in weird OK responses.

PR #964, opened 2024-04-08
feat: add logic to build an execution payload when proposing blocks.

mpaulucci:
**Motivation**
Being able to build an execution block

**Description**
Naive implementation of calling the execution layer to get an execution payload

### DIFF
--- a/lib/lambda_ethereum_consensus/execution/execution_client.ex
+++ b/lib/lambda_ethereum_consensus/execution/execution_client.ex
@@ -3,6 +3,7 @@ defmodule LambdaEthereumConsensus.Execution.ExecutionClient do
   Execution Layer Engine API methods
   """
   alias LambdaEthereumConsensus.Execution.EngineApi
+  alias LambdaEthereumConsensus.Execution.RPC
   alias LambdaEthereumConsensus.SszEx
   alias Types.DepositData
   alias Types.ExecutionPayload
@@ -10,6 +11,18 @@ defmodule LambdaEthereumConsensus.Execution.ExecutionClient do
   require Logger
 
   @type execution_status :: :optimistic | :valid | :invalid | :unknown
+
+  @spec get_payload(any()) :: {:error, any()} | {:ok, any()}
+  def get_payload(payload_id) do
+    case EngineApi.get_payload(payload_id) do
+      {:ok, %{"execution_payload" => raw_payload}} ->
+        {:ok, parse_raw_payload(raw_payload)}
+
+      {:error, reason} ->
+        Logger.warning("Error when calling get payload: #{inspect(reason)}")
+        {:error, reason}
+    end
+  end
 
   @doc """
   Verifies the validity of the data contained in the new payload and notifies the Execution client of a new payload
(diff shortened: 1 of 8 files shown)


lib/lambda_ethereum_consensus/validator/validator.ex
```diff
@@ -408,35 +411,69 @@ defmodule LambdaEthereumConsensus.Validator do
...
-    {:ok, ssz_encoded} = Ssz.to_ssz(signed_block)
-    {:ok, encoded_msg} = :snappyer.compress(ssz_encoded)
-    fork_context = BeaconChain.get_fork_digest() |> Base.encode16(case: :lower)
+    with {:ok, payload_id} <-
+           ExecutionClient.notify_forkchoice_updated(
+             forkchoice_state,
+             payload_attributes
+           ),
+         # TODO: find a better way
+         :ok <- Process.sleep(3000),
```
MegaRedHand, line 442: We should retry this a few times:

```elixir
Stream.repeatedly(fn -> ExecutionClient. get_payload(payload_id) end)
|> Stream.zip(Stream.timer(timeout_in_ms) |> Stream.cycle())
|> Stream.map(fn {v, 0} -> v end) 
|> Stream.take(number_of_retries)
|> Enum.find(&match?({:ok, _}, &1})
```

lib/lambda_ethereum_consensus/validator/validator.ex
```diff
@@ -408,35 +411,69 @@ defmodule LambdaEthereumConsensus.Validator do
...
 
-    # TODO: we might want to send the block to ForkChoice
+  defp propose(%{root: head_root, validator: %{index: index, privkey: privkey}}, proposed_slot) do
+    head_state = BlockStates.get_state!(head_root)
 
-    case Libp2pPort.publish("/eth2/#{fork_context}/beacon_block/ssz_snappy", encoded_msg) do
-      :ok -> Logger.info("[Validator] Proposed block for slot #{proposed_slot}")
-      _ -> Logger.error("[Validator] Failed to publish block for slot #{proposed_slot}")
+    with {:ok, execution_payload} <-
+           build_execution_block(head_state |> process_slots(proposed_slot), head_root) do
```
MegaRedHand, line 453: nit
```suggestion
           head_state |> process_slots(proposed_slot) |> build_execution_block(head_root) do
```

lib/lambda_ethereum_consensus/execution/execution_client.ex
```diff
@@ -11,6 +12,18 @@ defmodule LambdaEthereumConsensus.Execution.ExecutionClient do
...
   @type execution_status :: :optimistic | :valid | :invalid | :unknown
 
+  @spec get_payload(any()) :: {:error, any()} | {:ok, any()}
+  def get_payload(payload_id) do
+    case EngineApi.get_payload(payload_id) do
+      {:ok, %{"execution_payload" => raw_payload}} ->
+        {:ok, parse_raw_payload(raw_payload)}
+
+      {:error, reason} ->
+        Logger.warning("Error when calling get payload: #{inspect(reason)}")
```
MegaRedHand, line 22: This should be "error" level

MegaRedHand (approved):
Left some nits. LGTM otherwise

lib/lambda_ethereum_consensus/validator/validator.ex
MegaRedHand: But isn't that time variable? What happens if we call this without enough sleep?

mpaulucci: There will be fewer/no transactions but it will succeed

mpaulucci: When we make this code more sophisticated, we need to balance a time that should be long enough to let the execution pack as many transactions as possible (more fees for us) while giving enough time to propagate the block and have it included

mpaulucci: Added this comment in code.

PR #967, opened 2024-04-09
refactor: move block building code to proposer.

mpaulucci:
**Motivation**
Consolidating all block proposing code into the Proposer module

### DIFF
--- a/lib/lambda_ethereum_consensus/validator/build_block_request.ex
+++ b/lib/lambda_ethereum_consensus/validator/build_block_request.ex
@@ -14,9 +14,7 @@ defmodule LambdaEthereumConsensus.Validator.BuildBlockRequest do
     attester_slashings: [],
     attestations: [],
     voluntary_exits: [],
-    bls_to_execution_changes: [],
-    eth1_data: nil,
-    execution_payload: nil
+    bls_to_execution_changes: []
   ]
 
   @enforce_keys enforced_keys
--- a/lib/lambda_ethereum_consensus/validator/build_block_request.ex
+++ b/lib/lambda_ethereum_consensus/validator/build_block_request.ex
@@ -1,4 +1,4 @@
-defmodule LambdaEthereumConsensus.Validator.BlockRequest do
+defmodule LambdaEthereumConsensus.Validator.BuildBlockRequest do
   @moduledoc """
   Struct that stores and validates data for block construction.
   Most of the data is already validated when computing the state
--- a/lib/lambda_ethereum_consensus/validator/proposer.ex
+++ b/lib/lambda_ethereum_consensus/validator/proposer.ex
@@ -175,8 +175,22 @@ defmodule LambdaEthereumConsensus.Validator.Proposer do
 
   defp fetch_eth1_data(slot, head_state) do
     case ExecutionChain.get_eth1_vote(slot) do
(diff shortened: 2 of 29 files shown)


lib/lambda_ethereum_consensus/validator/proposer.ex
```diff
@@ -2,17 +2,42 @@ defmodule LambdaEthereumConsensus.Validator.Proposer do
...
+           pre_state
+           |> StateTransition.process_slots(proposed_slot)
+           |> build_execution_block(parent_root) do
+      block_request =
+        request
+        |> Map.merge(fetch_operations_for_block())
+        |> Map.put(:eth1_data, fetch_eth1_data(proposed_slot, pre_state))
+        |> Map.put(:execution_payload, execution_payload)
+
+      construct_block(pre_state, block_request, request.privkey)
```
MegaRedHand, line 37: Since we already have the `privkey` in the `block_request`, we should drop it from the args.

MegaRedHand (approved):
Good idea! Left some minor comments. LGTM otherwise.

lib/lambda_ethereum_consensus/validator/build_block_request.ex
```diff
@@ -14,9 +14,7 @@ defmodule LambdaEthereumConsensus.Validator.BlockRequest do
     attester_slashings: [],
     attestations: [],
     voluntary_exits: [],
-    bls_to_execution_changes: [],
-    eth1_data: nil,
-    execution_payload: nil
```
MegaRedHand: I'm not sure about this change. It might become a parameter explosion in the future.

mpaulucci: Im not convinced about this argument, I would like to separate the arguments passed from the caller (which are customizable) and the ones generated internally.

test/unit/validator/block_builder_test.exs
```diff
@@ -30,21 +34,23 @@ defmodule Unit.Validator.ProposerTests do
...
-    {:ok, signed_block} = Proposer.construct_block(pre_state, block_request, privkey)
+    {:ok, signed_block} =
+      private(
+        BlockBuilder.construct_block(
+          pre_state,
+          block_request,
+          spec_block.message.body.execution_payload,
+          spec_block.message.body.eth1_data
+        )
+      )
```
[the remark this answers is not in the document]
mpaulucci: We could do that, but why are you against the current approach? In general I think we should avoid making functions that are inherently private to be public just because of unit tests (we can solve it with patch)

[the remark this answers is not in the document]
MegaRedHand: It's just that testing a private function _smells_, and I think that this function could be used in a _standalone_ fashion for building customized blocks (e.g. with external payload builders). But yeah, I'm probably getting ahead of myself, and we should leave this for a later refactor.

lib/lambda_ethereum_consensus/validator/build_block_request.ex
MegaRedHand: I'm not thinking of them as "internal", that's why (related to https://github.com/lambdaclass/lambda_ethereum_consensus/pull/967#discussion_r1559865377). Let's merge this PR and discuss this later if/when it becomes a problem.

PR #972, opened 2024-04-11
fix: attesting on empty slots with the wrong state

MegaRedHand:
This fixes a bug we were having when attesting on empty slots:

```
no match of right hand side value: {:error, "slot is older than state"}
```

Since we process duties each time a block arrives, we don't notice the empty slot and try to attest with the state from the next slot, which triggers the error.


mpaulucci (approved):
nice!

PR #976, opened 2024-04-11
refactor: take Aja.Vectors out of SszEx module

avilagaston9:
This PR makes SszEx module agnostic of the list implementation being used (removing Aja.Vectors). Only Enumerable protocol is required.


MegaRedHand (approved):
LGTM

PR #977, opened 2024-04-11
perf: add a benchmark comparing hash_tree_root implementations

avilagaston9:
This PR adds a benchmark comparing hash_tree_root implementations between rustler and elixir library.
Also adds the same refactor made in `merkleize_chunks_with_virtual_padding()` to `merkleize_chunks()`.  `merkleize_chunks()` is currently not being used.


bench/ssz_ex_bench.exs
```diff
@@ -0,0 +1,36 @@
+alias LambdaEthereumConsensus.SszEx
+alias LambdaEthereumConsensus.Store.StateDb
+alias Types.BeaconState
+
+# To run this benchmarks, checkpoint-sync is needed.
```
MegaRedHand, line 5: We should also mention the `--mode db` necessary to run this

lib/ssz_ex.ex
```diff
@@ -331,39 +331,25 @@ defmodule LambdaEthereumConsensus.SszEx do
     root |> hash_nodes(serialized_len)
   end
 
+  # Currently, we are not using this one.
```
MegaRedHand, line 334: Let's make it a todo:
```suggestion
  # TODO: we are not using this
```

bench/ssz_ex_bench.exs
mpaulucci, line 5: I would remove this comment, since we don't need checkpoint-sync, we just need to have a state

lib/ssz_ex.ex
```diff
@@ -331,39 +331,25 @@ defmodule LambdaEthereumConsensus.SszEx do
...
-      new_buffer =
-        1..node_count
-        |> Enum.filter(fn x -> rem(x, 2) == 0 end)
+      power = leaf_count |> compute_pow()
+      height = power + 1
+      first_layer = chunks |> convert_to_next_pow_of_two(leaf_count)
+
+      final_layer =
+        1..(height - 1)
         |> Enum.reverse()
```
MegaRedHand, line 347: Since this is a range, you can just invert the range:
```suggestion
        (height - 1)..1
```

MegaRedHand (approved):
Left some comments. Otherwise, LGTM

PR #981, opened 2024-04-12
fix: use intermediate state when building a block.

mpaulucci:
**Description**
From the validator specs:

```
Note: In this section, state is the state of the slot for the block proposal without the block yet applied. That is, state is the previous_state processed through any empty slots up to the assigned slot using process_slots(previous_state, slot).
```
We call that state `mid_state` here


MegaRedHand (approved):
LGTM

PR #983, opened 2024-04-12
refactor: split SszEx into different submodules

avilagaston9:
**Motivation**

This PR refactors the SszEx module by splitting it into separate submodules. 
- Encode
- Decode
- Merkleization
- Hash
- Utils


bench/ssz.exs
```diff
@@ -1,4 +1,5 @@
 alias LambdaEthereumConsensus.SszEx
```
mpaulucci, line 1: I would remove this benchmark, we care about higher level ones.

lib/lambda_ethereum_consensus/ssz_ex/encode.ex
```diff
@@ -0,0 +1,187 @@
...
+         [{value, schema} | acc_variable_values]}
+      else
+        {[{value, schema} | acc_fixed_size_values],
+         acc_fixed_length + Utils.get_fixed_size(schema), acc_variable_values}
+      end
+    end)
+  end
+
+  defp encode_schemas(tuple_values) do
+    Enum.map(tuple_values, fn {value, schema} -> encode(value, schema) end)
```
mpaulucci, line 153: It would be nice that when we encode/decode or hash schemas and one field fails, then the field is mentioned in the error message.

lib/lambda_ethereum_consensus/ssz_ex/decode.ex
```diff
@@ -0,0 +1,379 @@
...
+          first_offset,
+          []
+        )
+        |> Enum.reverse()
+        |> Utils.flatten_results()
+      end
+    end
+  end
+
+  defp decode_variable_list_elements(
```
mpaulucci, line 186: is this handling the case were `num_elements` is `0`?

lib/lambda_ethereum_consensus/ssz_ex/decode.ex
```diff
@@ -0,0 +1,379 @@
...
+    with {:ok, fixed_parts, _offsets, items_index} <-
+           decode_fixed_section(binary, schemas, fixed_length),
+         :ok <- check_byte_len(items_index, byte_size(binary)) do
+      {:ok, struct!(module, fixed_parts)}
+    end
+  end
+
+  defp check_first_offset([{offset, _} | _rest], items_index, _binary_size) do
+    cond do
+      offset < items_index -> {:error, "OffsetIntoFixedPortion (#{offset})"}
```
mpaulucci, line 253: this seems inconsistent with other error messages. I would change the error message to something more readable

mpaulucci (approved):

lib/lambda_ethereum_consensus/ssz_ex/hash.ex
```diff
@@ -26,4 +25,11 @@ defmodule LambdaEthereumConsensus.Utils.ZeroHashes do
     <<_::binary-size(offset), hash::binary-size(@bytes_per_chunk), _::binary>> = zero_hashes
     hash
   end
+
+  @compile {:inline, hash: 1}
+  @spec hash(iodata()) :: binary()
+  def hash(data), do: :crypto.hash(:sha256, data)
+
+  @spec hash_nodes(binary(), binary()) :: binary()
+  def hash_nodes(left, right), do: :crypto.hash(:sha256, left <> right)
```
MegaRedHand, line 34: Nit:
```suggestion
  def hash_nodes(left, right), do: hash(left <> right)
```

lib/lambda_ethereum_consensus/ssz_ex/decode.ex
avilagaston9: It's done in line [#149](https://github.com/lambdaclass/lambda_ethereum_consensus/pull/983/files#diff-85ee24efe82c8f3df9daa43de90926eb4e761f11d46545f0fcce389e03108d3eR149).

PR #996, opened 2024-04-15
refactor: move SszEx library out of LambdaEthereumConsensus

avilagaston9:
This PR moves the SSZ library out of the LambdaEthereumConsensus module.


lib/types/deposit_tree.ex
```diff
@@ -3,8 +3,7 @@ defmodule Types.DepositTree do
   Pruned Merkle tree, for use in block production.
   Implementation adapted from [EIP-4881](https://eips.ethereum.org/EIPS/eip-4881).
   """
-  alias LambdaEthereumConsensus.SszEx
-  alias LambdaEthereumConsensus.SszEx.Merkleization
+  alias SszEx.Merkleization
```
MegaRedHand, line 6: Can't we use `SszEx` here?

PR #1005, opened 2024-04-16
feat: compute blob commitment inclusion proofs

MegaRedHand:
This PR implements the computation of the commitment inclusion proofs for the blob sidecars

### DIFF
--- a/lib/lambda_ethereum_consensus/validator/block_builder.ex
+++ b/lib/lambda_ethereum_consensus/validator/block_builder.ex
@@ -276,7 +276,9 @@ defmodule LambdaEthereumConsensus.Validator.BlockBuilder do
 
     commitment_tree_proofs =
       0..(commitment_number - 1)
-      |> Enum.map(&compute_merkle_proof(commitment_leaves, &1, commitment_tree_height))
+      |> Enum.map(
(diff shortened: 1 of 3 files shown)


lib/lambda_ethereum_consensus/validator/block_builder.ex
```diff
@@ -240,24 +241,80 @@ defmodule LambdaEthereumConsensus.Validator.BlockBuilder do
...
+
+  defp compute_merkle_proof(leaves, index, height) do
+    compute_merkle_proof(leaves, index, 0, height, [])
+  end
+
+  defp compute_merkle_proof([_root], _, max_height, max_height, proof) do
+    Enum.reverse(proof)
+  end
+
+  defp compute_merkle_proof(leaves, index, height, max_height, proof) do
```
mpaulucci, line 307: can we move this to a utility? I think there were spec tests specifically for merkle

MegaRedHand: We can move this somewhere else, but the spec-tests for Merkle proofs expect a generic implementation while this one is specific for blob commitments, so we can't use the test vectors.

MegaRedHand: I moved it to `SszEx.Merkleization`

mpaulucci (approved):

PR #1009, opened 2024-04-18
feat: add Dockerfile

### LINKED ISSUE #1000: Add Dockerfile
For easier packaging and integration with some tools (e.g. kurtosis) we need to dockerize our client.

MegaRedHand:
Closes #1000

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,18 @@
+FROM elixir:1.16.2
+
+RUN mkdir /app
+WORKDIR /app
+
+ENV MIX_ENV=prod
+
+RUN mix local.hex --force
+
+# copy release to app container
+COPY . .
+
+RUN chown -R nobody: /app
+USER nobody
+
+RUN make deps
+
+CMD ["iex", "-S", "mix", "run", "--", "--checkpoint-sync-url", "https://sepolia.checkpoint-sync.ethpandaops.io/", "--network", "sepolia", "--metrics", "--validator-file", "validator_sepolia.txt"]
--- a/mix.exs
+++ b/mix.exs
@@ -5,7 +5,7 @@ defmodule LambdaEthereumConsensus.MixProject do
     [
       app: :lambda_ethereum_consensus,
       version: "0.1.0",
-      elixir: "~> 1.15",
+      elixir: "~> 1.16",
       start_permanent: Mix.env() == :prod,
       deps: deps(),
       dialyzer: dialyzer(),
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,17 @@
+_build/
+.elixir_ls/
+.github/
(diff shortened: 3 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+
+# copy release to app container
+COPY . .
+
+RUN chown -R nobody: /app
+USER nobody
+
+RUN make deps
+
+CMD ["iex", "-S", "mix", "run", "--", "--checkpoint-sync-url", "https://sepolia.checkpoint-sync.ethpandaops.io/", "--network", "sepolia", "--metrics", "--validator-file", "validator_sepolia.txt"]
```
mpaulucci, line 18: should we use parameters or env variables here?

MegaRedHand: We might want to add support for env vars instead of CLI flags.

mix.exs
```diff
@@ -39,7 +39,7 @@ defmodule LambdaEthereumConsensus.MixProject do
       {:jason, "~> 1.4"},
       {:scrypt_elixir, "~> 0.1.1", hex: :scrypt_elixir_copy},
       {:joken, "~> 2.6"},
-      {:rustler, "~> 0.32"},
+      {:rustler, "~> 0.32", runtime: false},
```
mpaulucci, line 42: what are the implications of this change?

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+ENV PATH="${PATH}:/root/.cargo/bin"
+
+COPY . .
+COPY --from=libp2p_builder /libp2p_port/libp2p_port /app/priv/native/libp2p_port
+
+RUN mix deps.get
+RUN mix compile
+
+ENTRYPOINT [ "iex", "-S", "mix", "run", "--"]
```
mpaulucci, line 37: are the options from the caller passed here?

MegaRedHand: Yes. I forgot to add documentation, but the idea is you call `docker run consensus --checkpoint-sync <URL> etc`

mpaulucci: nice!

mpaulucci (approved):
Lets go! 🚀

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+
+RUN mkdir /app
+WORKDIR /app
+
+ENV MIX_ENV=prod
+
+RUN mix local.hex --force
+
+# Install dependencies
+RUN apt update
```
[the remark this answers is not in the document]
MegaRedHand: `apt-get` is the backend used by `apt`. I couldn't find any reason to use one over the other, so I just changed it to `apt-get`.

PR #1022, opened 2024-04-23
refactor: beacon block gossip to make it more flexible.

mpaulucci:
**Motivation**
We want to be able to manually enable when to start listening to the blocks topic, which we should do after the initial sync.

**Description**
Refactored beacon block topic subscriber to be it's own genserver.


MegaRedHand (approved):
LGTM

PR #1025, opened 2024-04-23
fix: switch order between metadata and beacon node processes

Arkenan:
**Description**

Fixes a race condition where metadata is requested immediately after p2p is initialized, but metadata genserver is not yet ready.


mpaulucci (approved):

MegaRedHand (approved):
Good catch!

PR #1042, opened 2024-04-25
feat: add keystore file and password file to runtime

### LINKED ISSUE #1027: Integrate Keystore with CLI
Context: #1013

Needs some thinking. Maybe add `--validator-keystore <keystore-path>` and `--validator-secret <secret-path>` flags.

Arkenan:
First naive implementation where keystore.json is in one file and the password in a different file, similar to some local testnet setups.

Closes #1027

### DIFF
--- a/config/runtime.exs
+++ b/config/runtime.exs
@@ -15,7 +15,6 @@ switches = [
   testnet_dir: :string,
   metrics: :boolean,
   metrics_port: :integer,
-  validator_file: :string,
   log_file: :string,
   beacon_api: :boolean,
   beacon_api_port: :integer,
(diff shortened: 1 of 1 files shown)


config/runtime.exs
```diff
@@ -140,18 +144,8 @@ config :lambda_ethereum_consensus, BeaconApi.Endpoint,
     layout: false
   ]
 
-# Validator
-#
-# `validator_file` should be a file with two non-empty lines, the first being
-# the public key and the second the private key, both hex-encoded
-# TODO: move to ERC-2335 keystores
-if validator_file do
```
MegaRedHand, line 148: We should also remove the `validator_file` switch

PR #1045, opened 2024-04-25
fix: add hexpm mirror

MegaRedHand:
We're having [this reoccurring error](https://github.com/lambdaclass/lambda_ethereum_consensus/actions/runs/8838463470/job/24269689352?pr=1044). This PR fixes that and removes the unneeded setup-beam step on building the native libraries.


mpaulucci (approved):
nice

PR #1051, opened 2024-04-26
fix: make sure to flush the logger before halting.

mpaulucci:
**Motivation**
The application was crashing but I wasn't seen any log before. This fixes it


MegaRedHand (approved):
Nice catch!

PR #1052, opened 2024-04-26
refactor: move subscriptions directly to operations_collector

### LINKED ISSUE #540: Wait for `ForkChoice` initialization before topic subscribing and request handling
I've seen this in lighthouse when doing checkpoint sync.

The client should NOT subscribe to gossip topics or process incoming requests until other parts of the application are initialized.

There should be a way to trigger the subscription of topics that is not tied to the genserver initialization.

### LINKED ISSUE #1048: Move subscription of operations to OperationsCollector
We want https://github.com/lambdaclass/lambda_ethereum_consensus/blob/main/lib/lambda_ethereum_consensus/p2p/gossip/operations_collector.ex to subscribe to the appropriate topics and listen to `{:gossipsub` directly.

We should be able to remove these files afterwards:
https://github.com/lambdaclass/lambda_ethereum_consensus/blob/main/lib/lambda_ethereum_consensus/p2p/gossip/gossipsub.ex
https://github.com/lambdaclass/lambda_ethereum_consensus/blob/main/lib/lambda_ethereum_consensus/p2p/gossip/consumer.ex
https://github.com/lambdaclass/lambda_ethereum_consensus/blob/main/lib/lambda_ethereum_consensus/p2p/gossip/handler.ex

avilagaston9:
This PR moves the subscriptions form **gossipsub** directly to **operations_collector**, removing the intermediary.
This way, we can trigger subscriptions after the sync is completed.

Closes #1048
Closes #540


mpaulucci (approved):
LGTM

PR #1053, opened 2024-04-26
perf: improve Dockerfile caching

MegaRedHand:
This PR improves docker build caching by pre-compiling Rust NIFs and mix dependencies, before copying the whole repo.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,19 +33,37 @@ RUN apt-get update && apt-get install -y cmake protobuf-compiler
 # Install protobuf for elixir
 RUN mix escript.install --force hex protobuf
 
+# Download openapi spec
+COPY Makefile .oapi_version /consensus/
+RUN make download-beacon-node-oapi
+
 # Install rust
 RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
 
 ENV PATH="${PATH}:/root/.cargo/bin:/root/.mix/escripts"
+
+# Precompile rust crates. Rustler stores targets under _build
+COPY ./native /consensus/native
+RUN cd /consensus/native/bls_nif && cargo build --release \
+    --target-dir="/consensus/_build/prod/lib/lambda_ethereum_consensus/native/bls_nif"
+RUN cd /consensus/native/kzg_nif && cargo build --release \
+    --target-dir="/consensus/_build/prod/lib/lambda_ethereum_consensus/native/kzg_nif"
+RUN cd /consensus/native/snappy_nif && cargo build --release \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -33,19 +33,37 @@ RUN apt-get update && apt-get install -y cmake protobuf-compiler
...
+RUN cd /consensus/native/kzg_nif && cargo build --release \
+    --target-dir="/consensus/_build/prod/lib/lambda_ethereum_consensus/native/kzg_nif"
+RUN cd /consensus/native/snappy_nif && cargo build --release \
+    --target-dir="/consensus/_build/prod/lib/lambda_ethereum_consensus/native/snappy_nif"
+RUN cd /consensus/native/ssz_nif && cargo build --release \
+    --target-dir="/consensus/_build/prod/lib/lambda_ethereum_consensus/native/ssz_nif"
+
+# Precompile elixir dependencies
+COPY mix.exs mix.lock .fork_version ./
+COPY ./config/config.exs /consensus/config/config.exs
```
mpaulucci, line 58: didn't know this file is read when compiling dependencies, but it makes sense

Dockerfile
```diff
@@ -33,19 +33,37 @@ RUN apt-get update && apt-get install -y cmake protobuf-compiler
...
+
 # Install rust
 RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
 
 ENV PATH="${PATH}:/root/.cargo/bin:/root/.mix/escripts"
 
+# Precompile rust crates. Rustler stores targets under _build
+COPY ./native /consensus/native
+RUN cd /consensus/native/bls_nif && cargo build --release \
+    --target-dir="/consensus/_build/prod/lib/lambda_ethereum_consensus/native/bls_nif"
```
mpaulucci, line 48: I think technically we should do `mix release` at the end and then this path would change. Wonder if we have another way the doesn't involver hardcoding this path.

I've been doing a small poc of doing `mix release` and then executing the binary directly instead of `iex` but the problem is that it doesn't recognize the cli for a reason.

This is what chatGPT said on the matter xD.

```
Direct CLI Arguments with start
The ./bin/app start command for a release typically does not support passing custom CLI arguments directly to the application's code because it starts the application in the background without an attached shell. This mode is meant for production where direct terminal interaction is not common. Instead, configurations are often managed through environment variables or external configuration files, as previously mentioned.
```

MegaRedHand: I also tried that and apparently, only env vars reach the depths of the `runtime.exs` script.

mpaulucci (approved):

Issue #849, opened 2024-03-06
Move BLS and KZG specific types to their specific modules


MegaRedHand:
We currently have some types that are used in the `Bls` and `Kzg` modules that should be moved to those modules, for example:

https://github.com/lambdaclass/lambda_ethereum_consensus/blob/9445b459c3ab539b4d6e7337e3c98e24a2586853/lib/kzg.ex#L41-L45

https://github.com/lambdaclass/lambda_ethereum_consensus/blob/9445b459c3ab539b4d6e7337e3c98e24a2586853/lib/bls.ex#L23

PR #1056, opened 2024-04-27
Move bls and kzg specific types to their specific modules

guha-rahul:
**Motivation**

**Description**

solves #849 
Closes #issue_number


guha-rahul:
This is super weird. I changed my branch.

PR #1057, opened 2024-04-27
refactor: move BLS and KZG specific types to their specific modules

### LINKED ISSUE #849: Move BLS and KZG specific types to their specific modules

guha-rahul:
**Motivation**

**Description**

Closes #849


lib/bls.ex
```diff
@@ -5,6 +5,10 @@ defmodule Bls do
   use Rustler, otp_app: :lambda_ethereum_consensus, crate: "bls_nif"
 
   @type privkey() :: <<_::256>>
+  @type bytes48 :: <<_::384>>
+  @type bytes96 :: <<_::768>>
+  @type bls_pubkey :: bytes48
+  @type bls_signature :: bytes96
```
MegaRedHand, line 11: We should remove the `bls_` prefix. Also, let's inline the `bytesN` types.

lib/kzg.ex
```diff
@@ -4,6 +4,11 @@ defmodule Kzg do
   """
   use Rustler, otp_app: :lambda_ethereum_consensus, crate: "kzg_nif"
 
+  # Max size: BYTES_PER_FIELD_ELEMENT * FIELD_ELEMENTS_PER_BLOB
+  @type blob :: binary
+  @type kzg_commitment :: bytes48
+  @type kzg_proof :: bytes48
```
MegaRedHand, line 10: We should remove the `kzg_` prefix, and inline the `bytesN` types

lib/types/types.ex
```diff
@@ -44,8 +44,4 @@ defmodule Types do
   @type withdrawal_index :: uint64
   @type payload_id :: bytes8
   @type blob_index :: uint64
-  # Max size: BYTES_PER_FIELD_ELEMENT * FIELD_ELEMENTS_PER_BLOB
-  @type blob :: binary
```
MegaRedHand, line 48: Let's keep these in here, to keep changes to a minimum.

lib/kzg.ex
```diff
@@ -4,6 +4,11 @@ defmodule Kzg do
...
+
   @spec blob_to_kzg_commitment(Types.blob()) :: {:ok, Types.kzg_commitment()} | {:error, binary()}
```
MegaRedHand, line 12: These typespecs should be updated with the new types:


```suggestion
  @spec blob_to_kzg_commitment(blob()) :: {:ok, commitment()} | {:error, binary()}
```

(same for `Bls` functions)

MegaRedHand (changes requested):
Thanks for your contribution!

lib/types/types.ex
guha-rahul: So i will be importing only the blob type from types.ex.

lib/kzg.ex
guha-rahul: updated all the types except blob()

guha-rahul:
@MegaRedHand hey i forgot to the update the typespaces in other files , but even after replacing them , I am getting the `dialyzer` error.

lib/types/types.ex
```diff
@@ -46,6 +46,4 @@ defmodule Types do
   @type blob_index :: uint64
   # Max size: BYTES_PER_FIELD_ELEMENT * FIELD_ELEMENTS_PER_BLOB
   @type blob :: binary
-  @type kzg_commitment :: bytes48
-  @type kzg_proof :: bytes48
```
MegaRedHand, line 50: The dialyzer errors are due to these two guys dissapearing. Let's keep them, but as a reference to `Kzg` types (i.e. `@type kzg_proof :: Kzg.proof()`). Same for Bls.

MegaRedHand (approved):
Great work!

guha-rahul:
Thank you for helping out!

PR #1065, opened 2024-05-03
fix: protobuf warnings in container

MegaRedHand:
When running the docker container, it emits warnings because we're redefining the protobuf-generated structs. This happens due to the output path being different from the repo, along with the `.dockerignore` not correctly ignoring generated files when copying.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -62,7 +62,7 @@ RUN mix deps.compile
 COPY . .
 COPY --from=libp2p_builder /libp2p_port/libp2p_port /consensus/priv/native/libp2p_port
 
-RUN protoc --elixir_out=./lib proto/libp2p.proto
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -62,7 +62,7 @@ RUN mix deps.compile
 COPY . .
 COPY --from=libp2p_builder /libp2p_port/libp2p_port /consensus/priv/native/libp2p_port
 
-RUN protoc --elixir_out=./lib proto/libp2p.proto
```
mpaulucci, line 65: are you sure about this change? in the ci it's `protoc --elixir_out=./lib proto/libp2p.proto`

MegaRedHand: Locally we build with the makefile that uses `.` only. We should update the CI too, just to have them all match.

mpaulucci (approved):

Issue #1028, opened 2024-04-23
Derive Keystore public key from secret key


MegaRedHand:
Context: #1013

Currently we're parsing it from the file, but the value could be wrong. Also, deriving it from the secret enables us to add an additional check by comparing both the parsed and derived values.

guha-rahul:
Can i take it?

MegaRedHand:
Sure! Go ahead! Let me know if you need some pointers.

guha-rahul:
> Sure! Go ahead! Let me know if you need some pointers.
Hey sorry for the late reply but dump as many pointers as you like .

MegaRedHand:
* The keystore is in `lib/keystore.ex`.
* We currently parse the public key here:
https://github.com/lambdaclass/lambda_ethereum_consensus/blob/24acec6627229953886c0d6203b9298ebd0a6392/lib/keystore.ex#L26-L27
* The idea is to derive the key from the secret key before.
* You might need to add a function to `lib/bls.ex` for that (keep in mind it's a NIF, so actual code is in Rust and lives inside `native/bls_nif`)

PR #1067, opened 2024-05-10
feat: derive Keystore public key from secret key

guha-rahul:
**Motivation**

**Description**

addresses #1028 

Having an error  in deserialization. deserialization accepts u8 but private key is of type Binary. In other functions it works fine.
Wondering where I went wrong.


lib/bls.ex
```diff
@@ -77,4 +77,15 @@ defmodule Bls do
       {:error, _} -> false
     end
   end
+
+  @doc """
+  Converts private to public key
+  """
+  @spec derive_pubkey?(privkey()) :: pubkey()
+  def derive_pubkey?(private_key) do
```
MegaRedHand, line 85: The name of this function doesn't match the one on Rust's side. NIFs (modules like these) work by modifying the code inside functions to make them call native functions. For that, you need to have a `derive_pubkey` function on this side too.

MegaRedHand (commented):
Left some comments. Let me know if you're having problems!

guha-rahul:
hey @MegaRedHand sorry for the late pr. But i have fixed the issue , let me know if you need any changes.

lib/keystore.ex
```diff
@@ -25,6 +25,9 @@ defmodule Keystore do
     privkey = decrypt!(decoded_json["crypto"], password)
     # TODO: derive from privkey and validate with this pubkey
     pubkey = Map.fetch!(decoded_json, "pubkey") |> parse_binary!()
+    if Bls.derive_pubkey(privkey) == pubkey do
```
MegaRedHand, line 28: This check is inverted: we want to raise when the derived pubkey is different from the specified one.

lib/bls.ex
```diff
@@ -77,4 +81,15 @@ defmodule Bls do
       {:error, _} -> false
     end
   end
+
+  @doc """
+  Converts private to public key
+  """
+  @spec derive_pubkey?(privkey()) :: {:ok, pubkey()} | {:error, any()}
+  def derive_pubkey?(private_key) do
```
MegaRedHand, line 89: The `?` at the end is because of an Elixir naming convention for functions that return booleans. We don't return booleans here, so we should remove the `?`.

lib/bls.ex
```diff
@@ -77,4 +81,15 @@ defmodule Bls do
...
+    case Bls.derive_pubkey(private_key) do
+      {:ok, pubkey} -> pubkey
+      {:error, _} -> false
+    end
```
MegaRedHand, line 93: We should do something similar to [`Ssz.hash_tree_root!`](https://github.com/lambdaclass/lambda_ethereum_consensus/blob/main/lib%2Fssz.ex#L51-L55) here: match against `{:ok, pk}`, and return `pk`. That way, if the Rust code returns an error, we just raise it.

MegaRedHand (commented):
Left you some more comments. Good work so far!

guha-rahul:
Thank you and add the necessary changes. Not sure why the 2nd commit happened.

lib/bls.ex
```diff
@@ -54,10 +54,6 @@ defmodule Bls do
     :erlang.nif_error(:nif_not_loaded)
   end
 
-  @spec derive_pubkey(privkey()) :: {:ok, pubkey()} | {:error, any()}
-  def derive_pubkey(_private_key) do
-    :erlang.nif_error(:nif_not_loaded)
```
MegaRedHand: Let's keep this one instead of the other. That way if the NIF is not loaded (and the implementation is not overwritten), calling the function raises a clear error.

lib/keystore.ex
```diff
@@ -23,11 +23,13 @@ defmodule Keystore do
     validate_empty_path!(decoded_json["path"])
 
     privkey = decrypt!(decoded_json["crypto"], password)
-    # TODO: derive from privkey and validate with this pubkey
+
     pubkey = Map.fetch!(decoded_json, "pubkey") |> parse_binary!()
-    if Bls.derive_pubkey(privkey) == pubkey do
+
+    if Bls.derive_pubkey(privkey) != {:ok, pubkey} do
```
MegaRedHand, line 29: Let's match against the `:ok` outside the `if`, so we differentiate between pubkey deriving errors and "pubkeys differ" errors:

```suggestion
    {:ok, derived_pubkey} = Bls.derive_pubkey(privkey)

    if derived_pubkey != pubkey do
```

MegaRedHand (commented):
Left you some more comments.

guha-rahul:
added the fix.

MegaRedHand (approved):
Great work! Please mark the PR as "ready for review" so we can merge it.

guha-rahul:
Thank you @MegaRedHand , for helping me out.

PR #1071, opened 2024-05-14
feat: remove states older than the last finalized checkpoint

### LINKED ISSUE #1063: Remove beacon states older than the last finalized block
We should remove beacon states older than the finalized block to keep the db size in check.

This is a naive approach, we want might to re-think this in the future if we need access to older states.

avilagaston9:
This PR makes the StateDb to remove BeaconStates older than the last finalized checkpoint. The process is triggered when the ForkChoice updates its Store.

Closes #1063


lib/lambda_ethereum_consensus/store/state_db.ex
```diff
@@ -30,6 +30,45 @@ defmodule LambdaEthereumConsensus.Store.StateDb do
...
+    with {:ok, it} <- Db.iterate(),
+         {:ok, @stateslot_prefix <> _slot, _value} <-
+           Exleveldb.iterator_move(it, last_finalized_key),
+         {:ok, slots_to_remove} <- get_slots_to_remove([], it),
+         :ok <- Exleveldb.iterator_close(it) do
+      slots_to_remove |> Enum.map(&remove_by_slot/1)
+    end
+  end
+
+  defp get_slots_to_remove(slots_to_remove, iterator) do
```
mpaulucci, line 46: the naming is a bit confusing. On one hand you are asking for slots to remove but you're also providing it as argument

lib/lambda_ethereum_consensus/store/state_db.ex
```diff
@@ -30,6 +30,45 @@ defmodule LambdaEthereumConsensus.Store.StateDb do
     Db.put(slothash_key_block, block_root)
   end
 
+  def remove_old_states(last_finalized_epoch) do
```
mpaulucci, line 33: IMO this function is too high level for this module.

mpaulucci:
good work!

lib/lambda_ethereum_consensus/fork_choice/handlers.ex
```diff
@@ -264,11 +264,19 @@ defmodule LambdaEthereumConsensus.ForkChoice.Handlers do
     )
     |> if_then_update(
       finalized_checkpoint.epoch > store.finalized_checkpoint.epoch,
-      # Update finalized checkpoint
-      &%Store{&1 | finalized_checkpoint: finalized_checkpoint}
+      fn store -> update_finalized_checkpoint(store, finalized_checkpoint) end
```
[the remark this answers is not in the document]
MegaRedHand: I agree. We can move this to `ForkChoice` easily.

mpaulucci (approved):
great work!

PR #1080, opened 2024-05-17
feat: add support for multiple validators.

mpaulucci:
**Motivation**
We want to be able to support multiple validators for a single node

**Description**
The way this works is to have a `keystore_dir ` and `keystore_password_file` each of which has one file per validator. The files look like this according to kurtosis:

`validator-keys/teku-keys/0xb05cafec5912f22dbd6f15677f25f13d93ecd5ec6f957fddd7cf27d73521b34aaaf6a219f77b21128d18321c2c8d679b.json`
`validator-keys/teku-secrets/0xb05cafec5912f22dbd6f15677f25f13d93ecd5ec6f957fddd7cf27d73521b34aaaf6a219f77b21128d18321c2c8d679b.txt`

Also fixed some validator bugs.


config/runtime.exs
```diff
@@ -153,17 +153,10 @@ config :lambda_ethereum_consensus, BeaconApi.Endpoint,
     layout: false
   ]
 
-if is_binary(keystore) and is_binary(keystore_pass) do
-  {pubkey, privkey} = Keystore.decode_from_files!(keystore, keystore_pass)
+config :lambda_ethereum_consensus, LambdaEthereumConsensus.Validator.Supervisor,
+  keystore_dir: keystore_dir,
+  keystore_pass_dir: keystore_pass_dir
```
MegaRedHand, line 158: Parsing should be done here to fail fast. We can add a TODO and open an issue if it's too much work.

lib/lambda_ethereum_consensus/validator/validator.ex
```diff
@@ -35,29 +39,39 @@ defmodule LambdaEthereumConsensus.Validator do
...
-    validator =
-      case {Keyword.get(config, :pubkey), Keyword.get(config, :privkey)} do
-        {nil, nil} -> nil
-        {pubkey, privkey} -> %{index: nil, privkey: privkey, pubkey: pubkey}
-      end
+  @type state :: %{
+          slot: Types.slot(),
+          root: Types.root(),
+          duties: %{
+            attester: list(:not_computed),
```
MegaRedHand, line 46: This is `list(:not_computed | %{...})`

lib/lambda_ethereum_consensus/validator/validator.ex
```diff
@@ -230,7 +231,7 @@ defmodule LambdaEthereumConsensus.Validator do
         old -> old |> Enum.reverse() |> Enum.take(1)
       end
 
-    %{duties | attester: attester_duties, proposer: old_duty ++ proposer_duties}
+    %{duties | attester: attester_duties, proposer: Enum.dedup(old_duty ++ proposer_duties)}
```
MegaRedHand, line 234: This reintroduces a bug related to last-slot-in-epoch proposals. We should instead discard "future" duties before computing `old_duty`. Something like:

```elixir
slot = Misc.compute_start_slot_for_epoch(epoch)
case duties.proposer do
  :not_computed -> []
  old -> old |> Enum.filter(&(&1 < slot)) |> Enum.reverse() |> Enum.take(1)
end
```

lib/lambda_ethereum_consensus/beacon/beacon_chain.ex
```diff
@@ -244,9 +244,11 @@ defmodule LambdaEthereumConsensus.Beacon.BeaconChain do
   defp notify_subscribers(logical_time) do
     log_new_slot(logical_time)
 
-    Enum.each([Validator, Gossip.BeaconBlock], fn subscriber ->
+    Enum.each([Gossip.BeaconBlock], fn subscriber ->
       GenServer.cast(subscriber, {:on_tick, logical_time})
     end)
+
+    ValidatorManager.notify_tick(logical_time)
```
mpaulucci, line 251: breaking the pattern here... ValidatorManager is not a genserver so it can't receive cast messages

lib/lambda_ethereum_consensus/validator/block_builder.ex
```diff
@@ -202,8 +202,16 @@ defmodule LambdaEthereumConsensus.Validator.BlockBuilder do
   end
 
   defp get_finalized_block_hash(state) do
-    finalized_block = Blocks.get_block!(state.finalized_checkpoint.root)
-    finalized_hash = finalized_block.body.execution_payload.block_hash
+    finalized_root = state.finalized_checkpoint.root
+
+    finalized_hash =
+      if finalized_root == <<0::256>> do
+        <<0::256>>
```
mpaulucci, line 209: dealing with the case where there is not finalized block yet

https://github.com/ethereum/execution-apis/blob/main/src/engine/paris.md#forkchoicestatev1

`Note: safeBlockHash and finalizedBlockHash fields are allowed to have 0x0000000000000000000000000000000000000000000000000000000000000000 value unless transition block is finalized.`

lib/lambda_ethereum_consensus/validator/validator.ex
```diff
@@ -35,29 +42,39 @@ defmodule LambdaEthereumConsensus.Validator do
...
+          payload_builder: {Types.slot(), Types.root(), BlockBuilder.payload_id()} | nil
+        }
 
+  @impl true
+  @spec init({Types.slot(), Types.root(), {Bls.pubkey(), Bls.privkey()}}) ::
+          {:ok, state, {:continue, any}}
+  def init({head_slot, head_root, {pubkey, privkey}}) do
     state = %{
-      slot: slot,
+      slot: head_slot,
```
Arkenan, line 60: Probably something to think about in a separate issue/PR, but what happens if a validator process dies and is revived after the node has been running for a while? If I'm thinking this correctly, the initial head_slot and head_root will probably be highly outdated. Is that something that may cause issues here?

My intuition tells me that pubkey and privkey are effectively init args, while head_slot and head_root should probably be obtained from a single source of truth instead of being passed as init args.

lib/lambda_ethereum_consensus/validator/validator_manager.ex
```diff
@@ -0,0 +1,80 @@
...
+
+    keystore_dir = Keyword.get(config, :keystore_dir)
+    keystore_pass_dir = Keyword.get(config, :keystore_pass_dir)
+
+    if keystore_dir == nil or keystore_pass_dir == nil do
+      Logger.warning(
+        "[Validator Manager] No keystore_dir or keystore_pass_dir provided. Validator will not start."
+      )
+
+      :ignore
```
Arkenan, line 25: Do we need to re-validate this here?

lib/lambda_ethereum_consensus/validator/validator_manager.ex
```diff
@@ -0,0 +1,80 @@
...
+      rescue
+        error ->
+          Logger.error(
+            "[Validator Manager] Failed to decode keystore file: #{keystore_file}. Pass file: #{keystore_pass_file} Error: #{inspect(error)}"
+          )
+
+          nil
+      end
+    end)
+    |> Enum.filter(&is_tuple/1)
```
Arkenan, line 91: Optional: maybe `Enum.reject(&is_nil)` is clearer?

Arkenan (approved):
LGTM, added some comments but are either optional or to be tackled on separate PRs.

lib/lambda_ethereum_consensus/validator/validator_manager.ex
mpaulucci: I agree!

lib/lambda_ethereum_consensus/validator/validator.ex
mpaulucci: In general we should have a ticket to revisit restarts of genservers.

mpaulucci: https://github.com/lambdaclass/lambda_ethereum_consensus/issues/1097

PR #1086, opened 2024-05-21
fix: trigger pruning on tick

avilagaston9:
State pruning was not being triggered because checkpoint updates were occurring within the `on_tick` function, while only updates within the `on_block` function had been considered.

### DIFF
--- a/lib/lambda_ethereum_consensus/store/state_db.ex
+++ b/lib/lambda_ethereum_consensus/store/state_db.ex
@@ -40,7 +40,7 @@ defmodule LambdaEthereumConsensus.Store.StateDb do
            Exleveldb.iterator_move(it, last_finalized_key),
          {:ok, slots_to_remove} <- get_slots_to_remove(it),
          :ok <- Exleveldb.iterator_close(it) do
-      slots_to_remove |> Enum.map(&remove_by_slot/1)
+      slots_to_remove |> Enum.each(&remove_by_slot/1)
       Logger.info("[StateDb] Pruning finished. #{length(slots_to_remove)} slots removed.")
     end
   end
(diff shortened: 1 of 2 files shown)


lib/lambda_ethereum_consensus/store/state_db.ex
```diff
@@ -31,29 +32,30 @@ defmodule LambdaEthereumConsensus.Store.StateDb do
...
            Exleveldb.iterator_move(it, last_finalized_key),
          {:ok, slots_to_remove} <- get_slots_to_remove(it),
          :ok <- Exleveldb.iterator_close(it) do
-      slots_to_remove |> Enum.map(&remove_by_slot/1)
+      slots_to_remove |> Enum.each(&remove_by_slot/1)
+      Logger.info("[StateDb] Pruning finished. #{length(slots_to_remove)} slots removed.")
     end
   end
 
   defp get_slots_to_remove(slots_to_remove \\ [], iterator) do
```
mpaulucci, line 48: can you add typespec here? I am confused if slots are integers or binaries

lib/lambda_ethereum_consensus/store/state_db.ex
```diff
@@ -30,30 +31,34 @@ defmodule LambdaEthereumConsensus.Store.StateDb do
...
            Exleveldb.iterator_move(it, last_finalized_key),
          {:ok, slots_to_remove} <- get_slots_to_remove(it),
          :ok <- Exleveldb.iterator_close(it) do
-      slots_to_remove |> Enum.map(&remove_by_slot/1)
+      slots_to_remove |> Enum.each(&remove_by_slot/1)
+      Logger.info("[StateDb] Pruning finished. #{length(slots_to_remove)} slots removed.")
     end
   end
 
+  @spec get_slots_to_remove(list(), :eleveldb.itr_ref()) :: {:ok, list(non_neg_integer())}
```
mpaulucci, line 49: list of what?

lib/lambda_ethereum_consensus/store/state_db.ex
avilagaston9: Maybe I can create an issue to refactor this for clarity.

mpaulucci (approved):

PR #1102, opened 2024-05-22
feat: add blobs pruning

### LINKED ISSUE #1103: Serve blob_sidecars for the latest `MIN_EPOCHS_FOR_BLOB_SIDECARS_REQUESTS` epochs
We should be able to serve blob_sidecars up until MIN_EPOCHS_FOR_BLOB_SIDECARS_REQUESTS epochs

See: https://github.com/ethereum/consensus-specs/blob/dev/specs/deneb/validator.md#sidecar

As part of this ticket, we should remove blobs older than MIN_EPOCHS_FOR_BLOB_SIDECARS_REQUESTS

avilagaston9:
Every time a new checkpoint is reached, spawns a supervised Task that removes blobs older than MIN_EPOCHS_FOR_BLOB_SIDECARS_REQUESTS.

Closes #1103


lib/lambda_ethereum_consensus/store/blob_db.ex
```diff
@@ -55,8 +72,55 @@ defmodule LambdaEthereumConsensus.Store.BlobDb do
...
+          {:ok, list(binary())}
+  defp get_block_root_keys_to_remove(keys_to_remove \\ [], iterator) do
+    case Exleveldb.iterator_move(iterator, :prev) do
+      {:ok, <<@block_root_prefix, _rest::binary>> = block_root_key, _root} ->
+        [block_root_key | keys_to_remove] |> get_block_root_keys_to_remove(iterator)
+
+      _ ->
+        {:ok, keys_to_remove}
+    end
+  end
```
Arkenan, line 101: Optional: maybe we can use `Enum.reduce_while` to get the key and delete it in the same iteration?

lib/lambda_ethereum_consensus/store/blob_db.ex
```diff
@@ -55,8 +72,55 @@ defmodule LambdaEthereumConsensus.Store.BlobDb do
...
+  def prune_blobs_older_than(slot) do
+    Logger.info("[BlobDb] Pruning started.", slot: slot)
+    last_finalized_key = slot |> block_root_key(0)
+
+    with {:ok, it} <- Db.iterate(),
+         {:ok, @block_root_prefix <> _, _value} <-
+           Exleveldb.iterator_move(it, last_finalized_key),
+         {:ok, keys_to_remove} <- get_block_root_keys_to_remove(it),
+         :ok <- Exleveldb.iterator_close(it) do
+      keys_to_remove |> Enum.each(&remove_blob_by_block_root_key/1)
```
Arkenan, line 85: We shouldn't do an enum.each and ignore the result of each removal. We should in the bare minimum log failures. If not, we can take care of it with a reduce.

Arkenan (changes requested):

lib/lambda_ethereum_consensus/store/blob_db.ex
Arkenan: Hmmm, let's not get stuck on this one too much, it's not even a performance improvement. We may think about it later on.

avilagaston9: Let me know if you have any additional comments on this PR or if we should approve it!

Arkenan (approved):
LGTM

PR #1106, opened 2024-05-24
fix: bug when proposing on slot 1

### LINKED ISSUE #1087: Localnet: Failing to build first block
When using a localnet with Kurtosis, if we are chosen to build the first block we fail.

Error message:
`ERROR 14:11:03.050 [Validator] Failed to start building payload for slot 1. Reason: No payload id, status is %{"latestValidHash" => nil, "status" => "INVALID", "validationError" => nil} `

avilagaston9:
This PR fix a bug when building the block for slot 1.
The genesis block does not store the payload_block_hash and we were trying to access to a "null" value.

image

Closes #1087

### DIFF
--- a/lib/lambda_ethereum_consensus/validator/block_builder.ex
+++ b/lib/lambda_ethereum_consensus/validator/block_builder.ex
@@ -104,9 +104,10 @@ defmodule LambdaEthereumConsensus.Validator.BlockBuilder do
 
     head_payload_hash =
       if proposed_slot == 1 do
-        {:ok, %{block_hash: block_hash}} = ExecutionClient.get_block_metadata(0)
-        Logger.info("[Store Setup] Receive head payload hash: #{inspect(block_hash)}")
-        block_hash
+        # The genesis block, does not have the payload_block_hash.
+        with {:ok, %{block_hash: block_hash}} <- ExecutionClient.get_block_metadata(0) do
+          block_hash
+        end
       else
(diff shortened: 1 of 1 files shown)


lib/lambda_ethereum_consensus/validator/block_builder.ex
```diff
@@ -101,7 +101,16 @@ defmodule LambdaEthereumConsensus.Validator.BlockBuilder do
...
+
+    head_payload_hash =
+      if proposed_slot == 1 do
+        # If the head is the genesis block, fetches the payload_block_hash.
+        with {:ok, %{block_hash: block_hash}} <- ExecutionClient.get_block_metadata(0) do
+          block_hash
+        end
+      else
+        head_block.body.execution_payload.block_hash
+      end
```
Arkenan, line 113: `with` is not used correctly here, as in an error case `head_payload_hash` will contain an error tuple, which won't be checked anywhere. We should probably do something like this:

```elixir

head_payload_hash =
  if proposed_slot == 1 do
    ExecutionClient.get_block_metadata(0)
  else
    {:ok, block_hash.body.execution_payload}
  end

with {:ok, %{block_hash: block_hash}} <- head_payload_hash
  # The rest of the already existing with
  do
end
```

Arkenan (changes requested):

Arkenan (approved):

PR #1107, opened 2024-05-24
fix: endpoint not available on GET /health

### LINKED ISSUE #1094: Endpoint persistent term not available when calling /health the first couple of times
Might be a configuration issue, but when first starting the node, the following error usually appears a couple of times:

```
ERROR 09:49:52.347 #PID<0.700.0> running BeaconApi.Endpoint (connection #PID<0.699.0>, stream id 1) terminated
Server: 172.16.0.16:4000 (http)
Request: GET /eth/v1/node/health
** (exit) an exception was raised:
    ** (RuntimeError) could not find persistent term for endpoint BeaconApi.Endpoint. Make sure your endpoint is started and note you cannot access endpoint functions at compile-time
        (lambda_ethereum_consensus 0.1.0) deps/phoenix/lib/phoenix/endpoint.ex:544: BeaconApi.Endpoint.persistent!/0
        (lambda_ethereum_consensus 0.1.0) deps/phoenix/lib/phoenix/endpoint.ex:584: BeaconApi.Endpoint.script_name/0
        (lambda_ethereum_consensus 0.1.0) lib/beacon_api/endpoint.ex:1: BeaconApi.Endpoint.call/2
        (plug_cowboy 2.7.1) lib/plug/cowboy/handler.ex:11: Plug.Cowboy.Handler.init/2
        (cowboy 2.12.0) /consensus/deps/cowboy/src/cowboy_handler.erl:37: :cowboy_handler.execute/2
        (cowboy 2.12.0) /consensus/deps/cowboy/src/cowboy_stream_h.erl:306: :cowboy_stream_h.execute/3
        (cowboy 2.12.0) /consensus/deps/cowboy/src/cowboy_stream_h.erl:295: :cowboy_stream_h.request_process/3
        (stdlib 5.2.3) proc_lib.erl:241: :proc_lib.init_p_do_apply/3 
```

It looks like the Endpoint is not available yet, which might be because it's the last thing to be started in the app supervisor.

avilagaston9:
**Motivation**

We had an error when calling GET /health the first couple of times.

**Description**

Just change the order of how the app supervisor children are initialized.

Closes #1094

### DIFF
--- a/lib/lambda_ethereum_consensus/application.ex
+++ b/lib/lambda_ethereum_consensus/application.ex
@@ -44,9 +44,9 @@ defmodule LambdaEthereumConsensus.Application do
   defp get_children(:full) do
     get_children(:db) ++
       [
+        BeaconApi.Endpoint,
         LambdaEthereumConsensus.P2P.Metadata,
-        LambdaEthereumConsensus.Beacon.BeaconNode,
-        BeaconApi.Endpoint
+        LambdaEthereumConsensus.Beacon.BeaconNode
       ]
   end
 


Arkenan (approved):
I'm not sure this is the right way, as we should probably bring up the endpoint after the node is available, but there's probably a non-trivial race condition there, so I'm ok with this for now.

PR #1111, opened 2024-05-27
refactor: libp2pPort gossip as callbacks instead of pids

### LINKED ISSUE #1108: [Refactor] Libp2pPort gossip as callbacks instead of pids
Currently, Libp2p gossip is routed through processes. That means that each message carries a pid, which is used by the Libp2pPort genserver (port owner) to redirect the messages to the corresponding handlers. That forces us to have a 1-handler 1-process equality, and we imiplement each handler as a genserver to make this easier.

To decouple the concurrency design from the Libp2pPort genserver, we can:

1. Create a `GossipHandler` behavior with a single `handle_gossip_message` callback.
2. Create a `topic => handler` behavior where the handler is a module that implements that behavior.
3. When a gossip message arrives, instead of sending a message to a pid, call the callback for the module corresponding to the topic.
4. Each `GenServer` should then implement the callback, and in the implementation, the call to the genserver sending the message should be implemented.

This doesn't change any behavior in the node. It only makes it easier to then change genservers to tasks or decide other concurrency models in the handlers instead of requiring a 1-to-1 relationship between handlers and processes.

avilagaston9:
**Motivation**

In our current implementation, we redirect gossip messages using the PIDs of the consumers. This forces us to have one process per consumer. 

**Description**
- Change the dispatch algorithm to use module names instead.
- Make the consumers implement a `handle_gossip_message` behavior.

Closes #1108


lib/libp2p_port.ex
```diff
@@ -187,25 +187,16 @@ defmodule LambdaEthereumConsensus.Libp2pPort do
   Subscribes to the given topic. After this, messages published to the topic
   will be received by `self()`.
   """
-  @spec subscribe_to_topic(GenServer.server(), String.t()) :: :ok | {:error, String.t()}
-  def subscribe_to_topic(pid \\ __MODULE__, topic_name) do
+  @spec subscribe_to_topic(GenServer.server(), String.t(), atom()) :: :ok | {:error, String.t()}
```
MegaRedHand, line 190:
```suggestion
  @spec subscribe_to_topic(GenServer.server(), String.t(), module()) :: :ok | {:error, String.t()}
```

lib/libp2p_port.ex
```diff
@@ -187,25 +187,16 @@ defmodule LambdaEthereumConsensus.Libp2pPort do
...
+  def subscribe_to_topic(pid \\ __MODULE__, topic_name, module) do
     :telemetry.execute([:port, :message], %{}, %{
       function: "subscribe_to_topic",
       direction: "elixir->"
     })
 
-    call_command(pid, {:subscribe, %SubscribeToTopic{name: topic_name}})
-  end
+    GenServer.cast(__MODULE__, {:new_subscriptor, String.to_atom(topic_name), module})
```
MegaRedHand, line 197: This should send the message to `pid` instead of `__MODULE__`. With that we can re-add the removed tests.

test/unit/libp2p_port_test.exs
```diff
@@ -86,37 +87,31 @@ defmodule Unit.Libp2pPortTest do
...
-
-      # Receive the message
-      assert {^topic, message_id, ^message} = Libp2pPort.receive_gossip()
-
-      Libp2pPort.validate_message(message_id, :accept)
-
-      # Send acknowledgement
-      send(pid, :received)
+      # Publish message
+      :ok = Libp2pPort.publish(:publisher, topic, message)
```
Arkenan, line 103: We don't need the spawn link anymore. This publish can be called right after the subscribe call.

test/unit/libp2p_port_test.exs
```diff
@@ -86,37 +87,31 @@ defmodule Unit.Libp2pPortTest do
     start_port(:publisher)
     start_port(:gossiper, listen_addr: gossiper_addr)
 
+    # The message sent is the pid
```
Arkenan, line 90: Let's explain why we do this (we use the message in this module's handler to decode the pid and send a notification)

Arkenan (changes requested):

Arkenan (approved):

PR #1115, opened 2024-05-29
refactor: remove GenServer behavior from gossip blobsidecar

### LINKED ISSUE #1114: [Refactor] Remove all GenServer behavior from BlobSideCar gossip
With the previous refactor done (#1108), there is no need for the BlobSideCar to be a GenServer anymore.

To complete this refactor, all GenServer behavior should be removed.

avilagaston9:
This PR removes all the GenServer behavior from the gossip BlobSideCar.

Closes #1114


lib/lambda_ethereum_consensus/p2p/gossip/blob_sidecar.ex
```diff
@@ -72,7 +24,38 @@ defmodule LambdaEthereumConsensus.P2P.Gossip.BlobSideCar do
         Logger.warning("[Gossip] Blob rejected, reason: #{inspect(reason)}")
         Libp2pPort.validate_message(msg_id, :reject)
     end
+  end
+
+  @spec join_topics() :: :ok
+  def join_topics() do
+    topics()
+    |> Enum.each(fn topic_name -> Libp2pPort.join_topic(self(), topic_name) end)
```
Arkenan, line 32: We should not use `each` if it can return an error and leave the node in an unusable state. We should be able to return an error here, for example, with a reduce, or with a map and then counting the errors.

Arkenan (changes requested):
1 comment about error handling. The rest is fine.

lib/lambda_ethereum_consensus/p2p/gossip/blob_sidecar.ex
```diff
@@ -72,7 +24,32 @@ defmodule LambdaEthereumConsensus.P2P.Gossip.BlobSideCar do
...
     end
+  end
 
-    {:noreply, topics}
+  @spec subscribe_to_topics() :: :ok | {:error, String.t()}
+  def subscribe_to_topics() do
+    topics()
+    |> Enum.each(fn topic ->
+      Libp2pPort.subscribe_to_topic(topic, __MODULE__)
+      |> case do
```
Arkenan, line 34: Let's avoid the pipe for a single operation.

```elixir
case Libp2pPort.subscribe_to_topic(topic, __MODULE__) do
 :ok -> ...
 {:error, reason} -> ...
end
```

Arkenan (approved):
LGTM with small comment

PR #1125, opened 2024-05-30
refactor: persist forkchoice state

### LINKED ISSUE #1116: [Refactor] Persist the fork choice store.
The motivation behind this is to be able to react to new blocks in fork choice by

1. Reading the store.
2. Applying state transition.
3. Saving the store.

### LINKED ISSUE #1117: [Refactor Fork choice] Measure how long it takes to interact with the store db
We need to see how long it takes to persist and how long it takes to read the store. We should be able to see this in graphana.

avilagaston9:
**Motivation**

In order to reduce complexity, we need to reduce concurrency as much as possible.

**Description**

This PR removes the use of the `GenServer` state, storing it into the db, taking the first step towards removing the `GenServer` calls.

Additionally, it adds some metrics on store persisting and fetching:
image

Closes #1116 
Closes #1117

### DIFF
--- a/lib/lambda_ethereum_consensus/fork_choice/fork_choice.ex
+++ b/lib/lambda_ethereum_consensus/fork_choice/fork_choice.ex
@@ -230,12 +230,17 @@ defmodule LambdaEthereumConsensus.ForkChoice do
   end
 
   defp persist_fork_choice_store(store) do
-    StoreDb.persist_fork_choice_store(store)
-    Logger.debug("[Fork choice] Store persisted")
+    :telemetry.span([:fork_choice, :persist], %{}, fn ->
+      {StoreDb.persist_fork_choice_store(store), %{}}
+    end)
   end
 
   defp fetch_fork_choice_store!() do
-    {:ok, store} = StoreDb.fetch_fork_choice_store()
+    {:ok, store} =
+      :telemetry.span([:fork_choice, :fetch], %{}, fn ->
+        {StoreDb.fetch_fork_choice_store(), %{}}
+      end)
+
     store
   end
 end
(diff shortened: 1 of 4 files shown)


lib/lambda_ethereum_consensus/fork_choice/fork_choice.ex
```diff
@@ -93,6 +95,7 @@ defmodule LambdaEthereumConsensus.ForkChoice do
         prune_old_states(last_finalized_checkpoint.epoch, new_finalized_checkpoint.epoch)
 
         GenServer.cast(from, {:block_processed, block_root, true})
+        persist_fork_choice_store(new_store)
```
Arkenan, line 98: I would persist before calling the cast. As a general rule it's useful that the state is stored and consistent before others are made aware of it. As a matter of fact, if we persist it before launching the recompute_head async task, we can remove the persist call in recompute-head.

lib/lambda_ethereum_consensus/fork_choice/fork_choice.ex
```diff
@@ -219,4 +228,19 @@ defmodule LambdaEthereumConsensus.ForkChoice do
       Logger.debug("[Fork choice] Store persisted")
     end)
   end
+
+  defp persist_fork_choice_store(store) do
+    :telemetry.span([:fork_choice, :persist], %{}, fn ->
```
Arkenan, line 233: Let's have the telemetry call on the db methods directly, so they are called no matter who is storing/fetching the store.

Arkenan (approved):
LGTM with some small comments.

lib/lambda_ethereum_consensus/fork_choice/fork_choice.ex
avilagaston9: Done!

lib/lambda_ethereum_consensus/fork_choice/fork_choice.ex
avilagaston9: That's true. Done!

PR #1126, opened 2024-05-30
refactor: remove forkchoice casts

### LINKED ISSUE #1118: [Refactor Fork Choice] Delete casts
Casts should be replaced by function calls from pending blocks.

avilagaston9:
This PR removes all `GenServer` casts from `ForkChoice`.

Casts were replaced by function calls from `PendingBlocks`.

Also, makes `recompute_head` synchronous adding grafana metrics:

image

Closes #1118


lib/lambda_ethereum_consensus/fork_choice/fork_choice.ex
```diff
@@ -90,25 +55,23 @@ defmodule LambdaEthereumConsensus.ForkChoice do
         :telemetry.execute([:sync, :on_block], %{slot: slot})
         Logger.info("[Fork choice] Added new block", slot: slot, root: block_root)
 
-        Task.async(__MODULE__, :recompute_head, [new_store])
+        Task.start_link(__MODULE__, :recompute_head, [new_store])
```
Arkenan, line 58: I'd just remove this level of concurrency for now. Let's just measure how long it takes and show it on grafana, but do it synchronously. 

On a design note: let's never use Task.start_link if we're not using it on a static supervisor (e.g. setting up processes for the first time). For this cases, we want a [dynamic supervisor](https://hexdocs.pm/elixir/1.12/Task.html#module-dynamically-supervised-tasks). In such a case, we would add a `Task.Supervisor` called `RecomputeHeadSupervisor` inside the beacon node supervisor, and we'd call 

```elixir
Task.Supervisor.start_child(
  RecomputeHeadSupervisor, fn -> recompute_head(new_store) end
)
``` 

when we want to recompute the head.

lib/lambda_ethereum_consensus/fork_choice/fork_choice.ex
```diff
@@ -67,11 +33,10 @@ defmodule LambdaEthereumConsensus.ForkChoice do
     :telemetry.execute([:sync, :on_block], %{slot: head_slot})
 
     persist_store(store)
-    {:ok, store}
   end
 
-  @impl GenServer
-  def handle_cast({:on_block, %BlockInfo{} = block_info, from}, _store) do
+  @spec on_block(BlockInfo.t(), pid()) :: :ok | :error
+  def on_block(%BlockInfo{} = block_info, from \\ self()) do
```
Arkenan, line 39: If this is no longer a GenServer cast, it's no longer necessary to cast back to the caller using `from`. Let's just return the result instead and process it directly in PendingBlocks.

lib/lambda_ethereum_consensus/fork_choice/fork_choice.ex
```diff
@@ -120,11 +83,10 @@ defmodule LambdaEthereumConsensus.ForkChoice do
       end
 
     persist_store(state)
-    {:noreply, state}
   end
 
-  @impl GenServer
-  def handle_cast({:attester_slashing, attester_slashing}, _state) do
+  @spec notify_attester_slashing(Types.AttesterSlashing.t()) :: :ok
+  def notify_attester_slashing(attester_slashing) do
```
Arkenan, line 89: This is not a notification, as this is not a process. Let's call this `on_attester_slashing`. Also, instead of `state` we can call the variable `store`.

Also, let's only persist the store if it actually changes (the handler workded), not in the error case.

lib/lambda_ethereum_consensus/fork_choice/fork_choice.ex
```diff
@@ -139,42 +101,35 @@ defmodule LambdaEthereumConsensus.ForkChoice do
...
 
-  def prune_old_states(last_finalized_epoch, new_finalized_epoch) do
+  defp prune_old_states(last_finalized_epoch, new_finalized_epoch) do
     if last_finalized_epoch < new_finalized_epoch do
       new_finalized_slot =
         new_finalized_epoch * ChainSpec.get("SLOTS_PER_EPOCH")
 
-      Task.async(StateDb, :prune_states_older_than, [new_finalized_slot])
-      Task.async(BlockDb, :prune_blocks_older_than, [new_finalized_slot])
+      Task.start_link(StateDb, :prune_states_older_than, [new_finalized_slot])
```
Arkenan, line 126: Again, let's not call start link unless we're statically supervising that task. Let's have a dynamic supervisor for these.

lib/lambda_ethereum_consensus/beacon/pending_blocks.ex
```diff
@@ -69,6 +74,12 @@ defmodule LambdaEthereumConsensus.Beacon.PendingBlocks do
     |> then(&{:noreply, &1})
   end
 
+  @impl true
+  def handle_cast({:on_tick, time}, state) do
+    ForkChoice.on_tick(time)
```
Arkenan, line 79: Why do we need this to pass through here? It doesn't use any of pending blocks primitives or state. Is it not possible to call it directly?

Arkenan (changes requested):
Thanks for the PR, Gaston! I added some comments.

lib/lambda_ethereum_consensus/fork_choice/fork_choice.ex
avilagaston9: Done!

lib/lambda_ethereum_consensus/fork_choice/fork_choice.ex
avilagaston9: Done!

lib/lambda_ethereum_consensus/beacon/pending_blocks.ex
avilagaston9: This is the only way I found to avoid race conditions. Since `ForkChoice` is no longer a GenServer and we are storing its state in the db, the `on_tick` call from `BeaconChain` may modify the state while  an `on_block` call from `PendingBlocks` is being executed.

lib/lambda_ethereum_consensus/fork_choice/fork_choice.ex
avilagaston9: That's right. Done!

lib/lambda_ethereum_consensus/fork_choice/fork_choice.ex
avilagaston9: You are right. Done!

lib/lambda_ethereum_consensus/beacon/pending_blocks.ex
```diff
@@ -199,12 +186,18 @@ defmodule LambdaEthereumConsensus.Beacon.PendingBlocks do
...
         true ->
-          ForkChoice.on_block(block_info)
-          state |> Map.put(block_root, {signed_block, :processing})
+          process_block(state, block_info)
       end
     end)
   end
 
+  defp process_block(state, block_info) do
+    case ForkChoice.on_block(block_info) do
```
Arkenan, line 195: We should run process blocks afterwards, as there may be blocks that now can be processed.

lib/lambda_ethereum_consensus/fork_choice/fork_choice.ex
```diff
@@ -55,18 +55,19 @@ defmodule LambdaEthereumConsensus.ForkChoice do
...
 
         prune_old_states(last_finalized_checkpoint.epoch, new_finalized_checkpoint.epoch)
 
         persist_store(new_store)
-        GenServer.cast(from, {:block_processed, block_root, true})
 
       {:error, reason} ->
         Logger.error("[Fork choice] Failed to add block: #{reason}", slot: slot, root: block_root)
-        GenServer.cast(from, {:block_processed, block_root, false})
+        :error
```
Arkenan, line 70: We should return the error with its reason and let the caller handle it.

lib/lambda_ethereum_consensus/beacon/pending_blocks.ex
Arkenan: Alright, I'll allow it for this PR. PendingBlocks will no longer be there in a few PRs though, so we'll have to figure that one out.

Arkenan (approved):
LGTM with two small comments.

PR #1127, opened 2024-05-31
feat: persist block with processing status

### LINKED ISSUE #1119: [Pending blocks refactor] Persist blocks with their processing state.
The idea behind this is to remove the pending blocks genserver state.

Arkenan:
**Description**

- Creates a BlockInfo struct that holds a signed block, a block root, and the processing status. This helps making storing operations easier, but also avoiding to recalculate the root if the block info is passed around.
- Every block store operation now uses block_info.

Next steps: use this to replace the PendingBlocks state.

Closes #1119

### DIFF
--- a/lib/lambda_ethereum_consensus/store/block_db.ex
+++ b/lib/lambda_ethereum_consensus/store/block_db.ex
@@ -43,9 +43,18 @@ defmodule LambdaEthereumConsensus.Store.BlockDb do
     end
   end
 
-  def store_block(_root, block_info), do: store_block(block_info)
+  defguard is_status(atom)
+           when atom in [
+                  :pending,
+                  :invalid,
+                  :processing,
+                  :download,
+                  :download_blobs,
+                  :unknown,
+                  :transitioned
+                ]
 
-  def store_block(%BlockInfo{} = block_info) do
+  def store_block_info(%BlockInfo{} = block_info) do
     {:ok, encoded_signed_block} = Ssz.to_ssz(block_info.signed_block)
 
     key = block_key(block_info.root)
@@ -58,11 +67,13 @@ defmodule LambdaEthereumConsensus.Store.BlockDb do
     Db.put(slothash_key, block_info.root)
   end
 
-  @spec get_block(Types.root()) ::
+  @spec get_block_info(Types.root()) ::
           {:ok, BlockInfo.t()} | {:error, String.t()} | :not_found
-  def get_block(block_root) do
-    with {:ok, data} <- Db.get(block_key(block_root)) do
-      decode_block_info(block_root, data)
+  def get_block_info(block_root) do
+    with {:ok, data} <- Db.get(block_key(block_root)),
+         {:ok, {encoded_signed_block, status}} <- :erlang.binary_to_term(data) |> validate_term(),
+         {:ok, signed_block} <- Ssz.from_ssz(encoded_signed_block, SignedBeaconBlock) do
+      {:ok, %BlockInfo{root: block_root, signed_block: signed_block, status: status}}
     end
   end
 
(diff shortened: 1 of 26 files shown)


lib/lambda_ethereum_consensus/store/block_db.ex
```diff
@@ -76,6 +104,16 @@ defmodule LambdaEthereumConsensus.Store.BlockDb do
     Logger.info("[BlockDb] Pruning finished. #{Enum.count(slots_to_remove)} blocks removed.")
   end
 
+  defp decode_block_info(root, data) do
+    with {:d, {encoded_signed_block, status}} <- {:d, :erlang.binary_to_term(data)},
```
avilagaston9, line 108: Sorry, could you explain me the purpose of the :d tag here? I don't catch it.

lib/lambda_ethereum_consensus/store/block_db.ex
```diff
@@ -10,31 +10,59 @@ defmodule LambdaEthereumConsensus.Store.BlockDb do
...
+
+    @spec from_block(SignedBeaconBlock.t(), Types.root(), block_status()) :: t()
+    def from_block(signed_block, root, status) do
+      %__MODULE__{root: root, signed_block: signed_block, status: status}
+    end
   end
 
-  def store_block(%SignedBeaconBlock{} = signed_block, block_root) do
-    {:ok, encoded_signed_block} = Ssz.to_ssz(signed_block)
+  def store_block(_root, block_info), do: store_block(block_info)
```
avilagaston9, line 46: I think we are currently not using this one.

lib/lambda_ethereum_consensus/store/blocks.ex
```diff
@@ -32,29 +32,32 @@ defmodule LambdaEthereumConsensus.Store.Blocks do
     }
   end
 
-  @spec store_block(Types.root(), SignedBeaconBlock.t()) :: :ok
-  def store_block(block_root, signed_block), do: LRUCache.put(@table, block_root, signed_block)
+  @spec store_block(BlockInfo.t()) :: :ok
+  def store_block(block_info) do
+    # The key is included in the block, no need to pass it.
+    LRUCache.put(@table, block_info.root, block_info)
```
avilagaston9, line 38: As the comment says, can't we just pass `block_info`?

lib/lambda_ethereum_consensus/store/block_db.ex
```diff
@@ -10,31 +10,59 @@ defmodule LambdaEthereumConsensus.Store.BlockDb do
...
+
+  def store_block(%BlockInfo{} = block_info) do
```
avilagaston9, line 48: What do you think about renaming this into `store_block_info`? Calling `get_block` is not the same as calling `get_block_info` in `StoreBlocks`, maybe we can continue with that rule.

lib/lambda_ethereum_consensus/store/block_db.ex
Arkenan: You're right! Deleted.

lib/lambda_ethereum_consensus/store/block_db.ex
Arkenan: Good idea! Done.

lib/lambda_ethereum_consensus/store/block_db.ex
Arkenan: Long story short, to identify the line, but it seems like it's a controversial technique. Forum on the topic: https://elixirforum.com/t/proper-way-of-handling-simple-booleans-in-with-statement/45556/20

I'll change it to something more readable.

PR #1129, opened 2024-06-03
refactor: replace PendingBlocks state with db calls

### LINKED ISSUE #1120: [Pending blocks refactor] Replace state with db
We also need metrics on how long it takes to save and to recover a block in this context.

Arkenan:
Changes:

- Removed state from PendingBlocks. Replaced with calls to DB.
- Adding some high level primitives in Blocks, such as change_status (for a single block) or get_blocks_by_status.

Closes #1120


avilagaston9 (approved):
LGTM! 🚀

PR #1136, opened 2024-06-06
refactor: persist gossip attestation state

### LINKED ISSUE #1135: [Gossip Attestation Refactor] Replace state with db
We also need metrics on how long it takes to save and to recover the state.

avilagaston9:
- Persists `Gossip.Attestation` state.
- Creates a `SubnetInfo` for storing subnet attestations.
- Adds metrics for the db calls.

Closes #1135


lib/types/subnet_info.ex
```diff
@@ -0,0 +1,38 @@
...
+  @type t :: %__MODULE__{
+          data: Types.AttestationData.t(),
+          attestations: list(Types.Attestation.t())
+        }
+
+  @spec new_subnet_with_attestation(Types.Attestation.t()) :: Types.SubnetInfo.t()
+  def new_subnet_with_attestation(%Types.Attestation{data: data} = attestation),
+    do: %__MODULE__{data: data, attestations: [attestation]}
+
+  @spec aggregate_attestation(t(), Types.Attestation.t()) :: t()
```
Arkenan, line 19: This doesn't aggregate (e.g. doesn't aggregate signatures). Let's call it something like `add_attestation`. Let's also document the function to explain what the function does (adds new attestations to the struct, but only if their data matches the base one).

lib/types/subnet_info.ex
```diff
@@ -0,0 +1,38 @@
...
+
+  defstruct [:data, :attestations]
+
+  @type t :: %__MODULE__{
+          data: Types.AttestationData.t(),
+          attestations: list(Types.Attestation.t())
+        }
+
+  @spec new_subnet_with_attestation(Types.Attestation.t()) :: Types.SubnetInfo.t()
+  def new_subnet_with_attestation(%Types.Attestation{data: data} = attestation),
```
Arkenan, line 16: Let's document this function and explain that the attestation will be used to filter other attestations added, and that it's typically the attestation that a validator built before starting to collect attestations.

lib/lambda_ethereum_consensus/telemetry.ex
```diff
@@ -142,6 +142,18 @@ defmodule LambdaEthereumConsensus.Telemetry do
       ),
       last_value("fork_choice.fetch.exception.duration",
         unit: {:native, :millisecond}
+      ),
+      last_value("subnet.persist.stop.duration",
```
Arkenan, line 146: Not important for this PR, but I think there's a way in which we can have only one metric for this and then have metadata saying for which table this is. In this way we can unify this for all table stores/gets and then filter in grafana instead.

lib/lambda_ethereum_consensus/p2p/gossip/attestation.ex
```diff
@@ -98,41 +101,45 @@ defmodule LambdaEthereumConsensus.P2P.Gossip.Attestation do
...
-  def handle_call({:stop_collecting, subnet_id}, _from, state) do
-    if Map.has_key?(state.attnets, subnet_id) do
-      {collected, atts} = Map.pop(state.attestations, subnet_id, [])
-      new_state = %{state | attnets: Map.delete(state.attnets, subnet_id), attestations: atts}
-      {:reply, {:ok, collected}, new_state}
-    else
-      {:reply, {:error, "subnet not joined"}, state}
+  def handle_call({:collect, subnet_id, attestation}, _from, _state) do
+    subnet_info = SubnetInfo.new_subnet_with_attestation(attestation)
+    persist_subnet_info(subnet_info, subnet_id)
```
Arkenan, line 110: nit: Let's put the subnet_id first. As a convention, we usually have our calls be `store(key, value)` instead of `store(value, key)`.

lib/lambda_ethereum_consensus/p2p/gossip/attestation.ex
```diff
@@ -9,16 +9,19 @@ defmodule LambdaEthereumConsensus.P2P.Gossip.Attestation do
   alias LambdaEthereumConsensus.P2P
   alias LambdaEthereumConsensus.P2P.Gossip.Handler
   alias LambdaEthereumConsensus.StateTransition.Misc
+  alias LambdaEthereumConsensus.Store.Db
```
Arkenan, line 12: Let's document the module and explain that this module handles the attestations from specific subnets, used by validators fulfilling aggregation duties. This is useful to not confuse it with the operations collector that handles the global attestations topic.

test/unit/attestation.exs
```diff
@@ -0,0 +1,112 @@
...
+    start_link_supervised!({Db, dir: tmp_dir})
+    start_link_supervised!(Attestation)
+    patch(Libp2pPort, :subscribe_to_topic, fn _, _ -> :ok end)
+    patch(Libp2pPort, :join_topic, fn -> :ok end)
+    patch(Libp2pPort, :leave_topic, fn -> :ok end)
+    patch(Libp2pPort, :update_enr, fn _, _, _ -> :ok end)
+    patch(Libp2pPort, :validate_message, fn _, _ -> :ok end)
+    patch(BeaconChain, :get_fork_digest, fn -> "9999" end)
+    patch(BeaconChain, :get_fork_version, fn -> "9999" end)
+    patch(Metadata, :get_metadata, fn -> %{attnets: nil, syncnets: nil} end)
```
Arkenan, line 29: If we're gonna mock everything to only test the db behavior, we might as well move everything related to the db to a separate module (e.g. SubnetInfo) and then test that module. We're not really testing the Attestation module here.

lib/types/subnet_info.ex
```diff
@@ -0,0 +1,38 @@
...
+    do: %__MODULE__{data: data, attestations: [attestation]}
+
+  @spec aggregate_attestation(t(), Types.Attestation.t()) :: t()
+  def aggregate_attestation(subnet_info, attestation) do
+    if subnet_info.data == attestation.data do
+      %__MODULE__{subnet_info | attestations: [attestation | subnet_info.attestations]}
```
Arkenan, line 22: Maybe this is for a different PR (let's start an issue if that's the case), but if we only save attestations with the same data, do we really need to save all attestation structs in a list? We could actually store only the signature and aggregation bits for each one. Either that or we aggregate in-place and hold only one attestation instead of a list.

Arkenan (changes requested):
Thanks for the PR Gastón! Added some comments.

lib/lambda_ethereum_consensus/telemetry.ex
avilagaston9: I'll create an issue for this!

lib/types/subnet_info.ex
avilagaston9: Created #1142!

Arkenan (approved):
LGTM, awesome PR

PR #1144, opened 2024-06-12
refactor: store OperationsCollector state

### LINKED ISSUE #1143: [OperationsCollectorRefactor] Store Genserver State
We need to store the `GenServer` state into the db, in order to remove the `GenServer` behavior in a future refactor.

avilagaston9:
- Stores the `GenServer` state into the db.
- Fixes a bug in `handle_call(:get)`.
- Adds some test.

Closes #1143


test/unit/operations_collector.exs
Arkenan, line 1: Can we add a test where we push two of the same operation (e.g. attestations) and get the operation? We would assert that both of them are present.

Arkenan, line 1: Let's also add a test where we have two different operations and we assert that we can get them separately without issues.

lib/lambda_ethereum_consensus/p2p/gossip/operations_collector.ex
```diff
@@ -202,68 +219,94 @@ defmodule LambdaEthereumConsensus.P2P.Gossip.OperationsCollector do
...
-        attester_slashing: attester_slashings,
-        proposer_slashing: proposer_slashings,
-        voluntary_exit: voluntary_exits,
-        attestation: attestations,
-        slot: slot
-    }
+    fetch_operation!(:attestation)
+    |> Stream.reject(&MapSet.member?(added_attestations, &1))
+    |> Enum.reject(&old_attestation?(&1, slot))
+    |> then(&store_operation(:attestation, &1))
```
Arkenan, line 262: Aren't we missing an update for the slot here? Or is it not necessary? If it's not necessary we should not save it at all. If it is, we should add a test that reflects its importance.

Arkenan (changes requested):
Great PR, asked a couple of questions and requested some changes.

lib/lambda_ethereum_consensus/p2p/gossip/operations_collector.ex
avilagaston9: Actually yes, nice catch! I will correct it and add some tests for that.

test/unit/operations_collector.exs
avilagaston9: Good idea!

avilagaston9: Good idea too!

Arkenan (approved):
LGTM

PR #1148, opened 2024-06-14
refactor: remove GenServer behavior from Gossip.Attestation

### LINKED ISSUE #1147: [Gossip Attestation] Remove GenServer behavior
Replace all `GenServer.casts` by function calls.

avilagaston9:
- Removes all `GenServer` behavior from `Gossip.Attestation`.
- Bonus: Fixes a bug when decoding `aggregation_bits`. We were assuming that the `aggregation_bits` were `binary`:
image

Closes #1147


lib/utils/bit_field.ex
```diff
@@ -89,8 +89,9 @@ defmodule LambdaEthereumConsensus.Utils.BitField do
   @spec bitwise_or(t, t) :: t
   def bitwise_or(left, right) when bit_size(left) == bit_size(right) do
     size = bit_size(left)
-    left_int = :binary.decode_unsigned(left)
-    right_int = :binary.decode_unsigned(right)
+    padding = 8 - rem(size, 8)
+    left_int = :binary.decode_unsigned(<<0::size(padding), left::bitstring>>)
+    right_int = :binary.decode_unsigned(<<0::size(padding), right::bitstring>>)
```
Arkenan, line 94: Great catch! I think the error wasn't the fact that they are bitstrings, but using the `:binary` methods entirely. I think instead of adding padding, we can process them as bitstrings directly. For instance, something like:

```elixir
<<left_int::size(size)>> = left
<<right_int::size(size)>> = right
```

After this you should be able to perform the bitwise or without issues. You can also use the following notation for bitwise or if you prefer so:

```elixir
left_int ||| right_int
```

test/unit/bit_field_test.exs
Arkenan, line 1: Amazing tests. Let's also add an example with more than 8 bits (e.g. 13)

Arkenan (approved):
LGTM with small comments.

PR #1153, opened 2024-06-18
refactor: remove GenServer behavior from OperationsCollector

### LINKED ISSUE #1152: [OperationsCollector] Remove GenServer behavior
Replace all `GenServer.casts` and `GenServer.calls` by function calls.

avilagaston9:
This PR removes all `GenServer` behavior from `OperationsCollector`.

Closes #1152


lib/lambda_ethereum_consensus/p2p/gossip/operations_collector.ex
```diff
@@ -41,35 +40,41 @@ defmodule LambdaEthereumConsensus.P2P.Gossip.OperationsCollector do
...
+        {:error, reason} ->
+          raise "[OperationsCollector] Subscription failed: '#{reason}'"
+      end
+    end)
   end
 
   @spec get_bls_to_execution_changes(non_neg_integer()) :: list(SignedBLSToExecutionChange.t())
   def get_bls_to_execution_changes(count) do
-    GenServer.call(__MODULE__, {:get, :bls_to_execution_change, count})
+    get_operation({:get, :bls_to_execution_change, count})
```
Arkenan, line 57: Let's remove the `:get` atom. We used it before to diferentiate the call from others, but now the function is called directly, so there's no need. We can have it be `get_operation(operation, count)`.

lib/lambda_ethereum_consensus/p2p/gossip/operations_collector.ex
```diff
@@ -41,35 +40,41 @@ defmodule LambdaEthereumConsensus.P2P.Gossip.OperationsCollector do
     "bls_to_execution_change"
   ]
 
-  def start_link(opts), do: GenServer.start_link(__MODULE__, opts, name: __MODULE__)
-
   def start() do
```
Arkenan, line 43: Let's call this subscribe_to_topics, unless we're doing the `start` thing in all the rest. The name should describe what it does, specially if there are side-effects involved. Also, any reason to `raise` instead of returning an error?

lib/lambda_ethereum_consensus/p2p/gossip/operations_collector.ex
```diff
@@ -82,40 +87,17 @@ defmodule LambdaEthereumConsensus.P2P.Gossip.OperationsCollector do
       attestations: block.body.attestations
     }
 
-    GenServer.cast(__MODULE__, {:new_block, block.slot, operations})
+    filter_messages(block.slot, operations)
```
Arkenan, line 90: If this is the only place where we call filter_messages, let's move the contents here directly, and we can avoid constructing the "operations" dictionary. We can use the block directly.

lib/lambda_ethereum_consensus/p2p/gossip/operations_collector.ex
```diff
@@ -82,40 +87,17 @@ defmodule LambdaEthereumConsensus.P2P.Gossip.OperationsCollector do
...
   end
 
-  @impl true
-  def handle_gossip_message(topic, msg_id, message) do
-    GenServer.cast(__MODULE__, {:gossipsub, {topic, msg_id, message}})
-  end
-
-  @impl GenServer
-  def init(_init_arg) do
+  def init() do
```
Arkenan, line 93: Let's add a comment explaining that this both joins the topics and sets the operations as empty in the DB. Remember, side effects must be clear for the caller in the comments.

Arkenan (changes requested):
Thanks for the PR! Added some comments.

lib/lambda_ethereum_consensus/p2p/gossip/operations_collector.ex
avilagaston9: Totally!

lib/lambda_ethereum_consensus/p2p/gossip/operations_collector.ex
avilagaston9: I thought of using raise because if it fails, we would want to restart. But maybe we can handle that in the caller module. I have created an issue for handling subscription errors so I'll return an error message here and leave that for a different PR.

Arkenan (approved):
LGTM

PR #1162, opened 2024-06-19
refactor: new_peer_handler as a function call

### LINKED ISSUE #1161: [PeerBook Refactor] new_peer_handler as a function call
In order to remove GenServer behavior from `Peerbook` we need to stop using PIDs to redirect `NewPeer` notifications.
We should invoke a function call directly from `LibP2PPort` instead.

avilagaston9:
- Removes `:new_peer_handler` from `LibP2PPort` state.
- When a `NewPeer` notification arrives, invokes `Peerbook.new_peer_handler()` directly.

Closes #1161


Arkenan (approved):
LGTM

PR #1171, opened 2024-06-25
refactor: remove GenServer behavior from Peerbook

### LINKED ISSUE #1170: [Peerbook Refactor] Remove GenServer behavior
To complete this issue, we should replace all `GenServer` casts and calls with function calls.

avilagaston9:
- Replaces all `GenServer` casts and calls with function calls.
- Uses `KvSchema` behavior for db calls.

Closes #1170


Arkenan (approved):
LGTM

PR #1175, opened 2024-06-27
fix: get_blocks_with_status

### LINKED ISSUE #1174: get_blocks_with_status sometimes returns elements that are not of the correct status
When executing 

```elixir
Blocks.get_blocks_with_status(:pending)
```

It's returning a list of blocks with this statuses (used `Enum.frequencies_by` to get this result):

```elixir
statuses: %{invalid: 10, pending: 91, transitioned: 1}
```

This is making the same block be processed over and over again.

Arkenan:
Changes `PendingBlocks` to update the root of the block from `:download` to `:download_blobs` once downloaded.
Closes #1174


avilagaston9 (approved):
LGTM!

PR #1185, opened 2024-07-01
refactor: remove execution_chain GenServer

### LINKED ISSUE #1193: Remove ExecutionChain genserver
We need to replace all `GenServer.casts` with function calls by storing the `ExecutionChain`'s state in the db.

avilagaston9:
- Stores the `ExecutionChain`'s state in the db.
- Replaces all `GenServer.casts` with function calls.

closes #1193


lib/lambda_ethereum_consensus/execution/execution_chain.ex
```diff
@@ -266,4 +269,15 @@ defmodule LambdaEthereumConsensus.Execution.ExecutionChain do
 
   defp slots_per_eth1_voting_period(),
     do: ChainSpec.get("EPOCHS_PER_ETH1_VOTING_PERIOD") * ChainSpec.get("SLOTS_PER_EPOCH")
+
+  defp persist_execution_state(state) do
+    put("", state)
+  end
```
Arkenan, line 275: nit, maybe a one-liner here like the get?

lib/lambda_ethereum_consensus/execution/execution_chain.ex
```diff
@@ -266,4 +269,15 @@ defmodule LambdaEthereumConsensus.Execution.ExecutionChain do
...
+
+  defp fetch_execution_state(), do: get("")
+
+  defp fetch_execution_state!() do
```
Arkenan, line 279: Can we add a spec to the get and persist functions so it gives a clear indication of what shape the "execution_state" has?

Arkenan (approved):
LGTM with small comments

lib/lambda_ethereum_consensus/execution/execution_chain.ex
avilagaston9: Done!

lib/lambda_ethereum_consensus/execution/execution_chain.ex
avilagaston9: Good idea!

PR #1195, opened 2024-07-03
refactor: remove IncomingRequests GenServer

### LINKED ISSUE #1192: Remove the Request receiver genserver
There's currently three modules to handle requests:

1. `LambdaEthereumConsensus.P2P.IncomingRequests.Receiver`: GenServer. This is the "handler" that libp2p saves as a pid to send requests to.
2. `LambdaEthereumConsensus.P2P.IncomingRequests`: Supervisor. Has the genserver and dynamically supervises tasks.
3. `LambdaEthereumConsensus.P2P.IncomingRequests.Handler`: Actually handles the request.

We need to:

- [ ] Remove the GenServer. The tasks should be started directly from Libp2pPort, for instance with `IncomingRequests.handle(name, message_id, message)` or something like that. There's no point in starting them from the genserver.
- [ ] Remove the tasks and the supervisor. This has an inferior priority, but if we're removing all concurrency primitives, we can also remove the `IncomingRequests` module and leave only the handler (renamed as `IncomingRequests`).

avilagaston9:
- Handles requests directly from Libp2pPort (sequentially).
- Removes `IncomingRequests` Supervisor and GenServer.
- Adds metrics to the request handlers.
- Updates the ping/pong test.
- Removes the handlers' field from the Go codebase.

Closes #1192


lib/lambda_ethereum_consensus/p2p/incoming_requests_handler.ex
```diff
@@ -1,35 +1,55 @@
...
+
+    case result do
+      {:error, error} -> {:error, "[#{name}] Request error: #{inspect(error)}"}
+      result -> result
     end
   end
 
   @spec handle_req(String.t(), String.t(), binary()) ::
-          :ok | {:error, String.t()}
+          :ok | {:ok, any()} | {:error, String.t()}
```
Arkenan, line 45: I'm not sure it's possible to return just `:ok` here? You either return an `{:ok, {message_id, payload}}` tuple that you wrote manually in each one or the with fails, I assume with an error tuple. I'd remove that from the spec.

lib/libp2p_port.ex
```diff
@@ -299,6 +293,13 @@ defmodule LambdaEthereumConsensus.Libp2pPort do
     OperationsCollector.init()
   end
 
+  @spec enable_request_handlers(port()) :: :ok | {:error, String.t()}
+  defp enable_request_handlers(port) do
+    IncomingRequestsHandler.protocol_ids()
+    |> Stream.map(fn protocol_id -> set_handler(protocol_id, port) end)
+    |> Enum.each(fn true -> nil end)
```
Arkenan, line 300: Why not an Enum.each directly?

```suggestion
    IncomingRequestsHandler.protocol_ids()
    |> Enum.each(fn protocol_id -> set_handler(protocol_id, port) end)
```

Alternatively, if the goal is to make it fail explicitely we can 

```elixir
    IncomingRequestsHandler.protocol_ids()
    |> Enum.each(fn protocol_id -> true = set_handler(protocol_id, port) end)
```

lib/libp2p_port.ex
```diff
@@ -412,15 +415,24 @@ defmodule LambdaEthereumConsensus.Libp2pPort do
...
-         state
+         %{port: port} = state
        ) do
     :telemetry.execute([:port, :message], %{}, %{function: "request", direction: "->elixir"})
-    handler_pid = :erlang.binary_to_term(handler)
-    send(handler_pid, {:request, {protocol_id, request_id, message}})
+
+    case IncomingRequestsHandler.handle(protocol_id, request_id, message) do
+      :ok ->
+        :ok
```
Arkenan, line 427: Is this case not impossible? If so we should remove it.

lib/libp2p_port.ex
```diff
@@ -117,14 +119,16 @@ defmodule LambdaEthereumConsensus.Libp2pPort do
...
   @doc """
-  Sets a Req/Resp handler for the given protocol ID. After this call,
-  peer requests are sent to the current process' mailbox. To handle them,
-  use `handle_request/0`.
+  Sets libp2pport as the Req/Resp handler for the given protocol ID.
   """
-  @spec set_handler(GenServer.server(), String.t()) :: :ok | {:error, String.t()}
-  def set_handler(pid \\ __MODULE__, protocol_id) do
+  @spec set_handler(String.t(), port()) :: boolean()
+  def set_handler(protocol_id, port) do
```
Arkenan, line 125: Maybe we can make this private too?

Arkenan (approved):
LGTM with some comments

lib/libp2p_port.ex
```diff
@@ -166,27 +170,17 @@ defmodule LambdaEthereumConsensus.Libp2pPort do
...
-  end
-
   @doc """
   Sends a response for the request with the given message ID.
   """
-  @spec send_response(GenServer.server(), String.t(), binary()) ::
-          :ok | {:error, String.t()}
-  def send_response(pid \\ __MODULE__, request_id, response) do
+  @spec send_response({String.t(), binary()}, port()) :: boolean()
+  def send_response({request_id, response}, port) do
```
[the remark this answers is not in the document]
avilagaston9: Sure!

lib/libp2p_port.ex
avilagaston9: Done!

PR #1202, opened 2024-07-05
chore(kurtosis): kurtosis setup, convenience tasks, and docs

rodrigo-o:
**Motivation**

Being able to easily test validators behavior through Kurtosis and add documentation about the process of setting up and starting it.

**Description**
- [x] Add convenient make tasks for setting up and running kurtosis
- [x] Add an initial document section about how to use kurtosis
- [x] Set up sensible defaults to avoid the node being shutdown due to OOM 
- [x] Modify and test the Dockerfile to allow a simple node connection to the running instance
  - [x] Maintain backwards compatibility without really changing the way it was previously built, just added IEX_ARGS as a build variable
- [x] Add make tasks for easy connections
- [x] Configure and test metrics working as expected for the node

Isues: #1197 & #1198 

**TO DO**

This task spawn a couple of issues outside of the current scope:
- To be able to appear in the Kurtosis predefined dashboards for the Beacon clients we need to adhere to the metrics specified for consensus clients, which should be the following: https://github.com/ethereum/beacon-metrics/blob/master/metrics.md
- From our custom Dashboard, only the `Node` one is working right now, both `BEAM` & `ERTS memory allocators` appear to be empty. Given that I didn't see them working outside of kurtosis neither, it appears to be a previous issue so I'll leave it for a follow-up.


Arkenan (approved):
LGTM

Issue #1207, opened 2024-07-11
Make DB modules to use KvSchema


Arkenan:
These modules can be modified to use the KvSchema module and unify how telemetry is collected + free methods.

- [ ] Blob DB
- [ ] Block DB
- [ ] Block states
- [ ] State DB
- [ ] Metadata
- [ ] Operations collector
- [ ] SubnetInfo
- [ ] Block by slot (now part of block db)
- [ ] Block status by root.

PR #1217, opened 2024-07-12
refactor: make BlockBySlot its own KvSchema

### LINKED ISSUE #1190: Make checkpoint sync asynchronous
Currently there's this to download missing blocks:

```elixir
Task.async_stream(
        fn chunk -> fetch_blocks_by_slot(chunk.from, chunk.count) end,
        max_concurrency: 4,
        timeout: 20_000,
        on_timeout: :kill_task
      )
```

The problem is that sometimes there are many missing slots and the whole downloading task takes more than 20 seconds. Each of this fetch is synchronous. That means that it will be retried with similar chances of success. Ideally, we want these to async, so we can just have them retry if they fail, and if we set timeouts we may do so individually and not collectively, as the time of that collective task is dependent on the number of missing blocks.

The only tricky part here is that the mechanism to start subscriptions after every block is downloaded will not be trivial as it needs to track which chunks are still missing. That state might need to be handled either at port level or db level.

An alternative is to only download one chunk at a time, which might be slower, but steadier than all of them at once with a hardcoded timeout.

Arkenan:
Changes:

- Added `fold_keys` function + option to include the starting key in KvSchema.
- Added `BlockBySlot` module, implementing `KvSchema`. Includes tests.
- Implement an `all_present?` function in `BlockBySlot` to check if a range of slots has all blocks available.
- Made the BlockDb module use BlockBySlot instead of implementing all. Reduced code via KvSchema gets, puts, folds, removes.
- Added tests for BlockDb behaviors that require block by slot like pruning.

Part of #1207

### DIFF
--- a/lib/beacon_api/controllers/v1/beacon_controller.ex
+++ b/lib/beacon_api/controllers/v1/beacon_controller.ex
@@ -5,7 +5,7 @@ defmodule BeaconApi.V1.BeaconController do
   alias BeaconApi.ErrorController
   alias BeaconApi.Helpers
   alias BeaconApi.Utils
-  alias LambdaEthereumConsensus.Store.BlockDb
+  alias LambdaEthereumConsensus.Store.BlockBySlot
   alias LambdaEthereumConsensus.Store.Blocks
   alias LambdaEthereumConsensus.Store.StoreDb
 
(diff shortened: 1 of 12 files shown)


avilagaston9 (approved):
LGTM!

PR #1219, opened 2024-07-15
feat: make checkpoint sync asynchronous

### LINKED ISSUE #1190: Make checkpoint sync asynchronous

Arkenan:
Changes:

- Removes the sync_blocks task. Makes it a module.
- The chunk requests are now all performed concurrently.
- The amount of retries is incremented to 50.
- After a chunk is downloaded or fails, a message is sent to libp2p port genserver to handle it.
- libp2p port genserver now has a counter of remaining blocks until the optimistic sync is finished.
- libp2p now handles the subscriptions after checkpoint sync, directly from within the genserver.
- Small typo fix: subscriptor -> subscriber.
- Call `add_block` directly when a block is downloaded instead of going through Libp2pPort (the handler is ran through libp2p port anyway.

Closes #1190


lib/libp2p_port.ex
```diff
@@ -370,11 +421,55 @@ defmodule LambdaEthereumConsensus.Libp2pPort do
...
+        LambdaEthereumConsensus.P2P.Gossip.OperationsCollector
+      ]
+      |> Enum.flat_map(&topics_for_module/1)
+      |> Enum.reduce(new_state, fn {module, topic}, state ->
+        command = %Command{c: {:subscribe, %SubscribeToTopic{name: topic}}}
+        send_data(state.port, Command.encode(command))
+        add_subscriber(state, topic, module)
+      end)
+      |> then(fn state -> {:noreply, state} end)
+    end
```
avilagaston9, line 454: I think this function is doing too many things. Maybe we can move this to a new `subscribe_to_gossip_topics` function?

avilagaston9 (approved):
LGTM! With just a nit comment.

PR #1227, opened 2024-07-17
chore(warnings): fixed some error and warning messages seen throughout tests

rodrigo-o:
**Motivation**

This small PR address or comment some error/warnings encountered during the Node execution

**Description**

None of this were critical, but they were easy to solve so compiled them in this PR, this are the ones tackled:

_libGLU.so.1_
```
~c"Failed to load NIF library /usr/local/lib/erlang/lib/wx-2.4.1/priv/erl_gl: 'libGLU.so.1: cannot open shared object file: No such file or directory'"}}
````
This was due to the addition of `:wx` as an extra application by default which needed additional libraries installed in docker but couldn't be used directly. I added an `EXTRA_APPLICATIONS` env variable to deal with this. By default `:wx` is not part of the default `extra_applications`, it's just added if `EXTRA_APPLICATIONS=WX` is set. This is also specified in the only section of the README that talks about `:observer` with a small note.

*Alternative solution:* We could just add `:wx` for `dev`, not `prod`, but i'm not sure that's exactly what we want, this was a quick fix for the issue and still allows for someone testing with observer and wx to just set it up for any env without modifying the `mix.exs`. For a more complex alternative (maybe for the future) like SSH tunnels with the docker container here is an [elixir forum thread](https://elixirforum.com/t/running-local-observer-against-a-remote-docker/4788/7) and an [example repo](https://github.com/chazsconi/connect-remote-elixir-docker)

_eth/v1/node/version API_
```
2024-07-16 13:26:11 DEBUG 16:26:11.852 Processing with BeaconApi.ErrorController.not_found/2
2024-07-16 13:26:11   Parameters: %{"path" => ["eth", "v1", "node", "version"]}
````

This was related to the simple [version RPC method](https://www.quicknode.com/docs/ethereum/eth-v1-node-version). That was the only one appearing in the logs, so I though of tackle it. But unfortunately it was because that failure stopped kurtosis for continuing reaching the API. once added it started asking for the next endpoint: [`eth/v1/node/syncing`](https://www.quicknode.com/docs/ethereum/eth-v1-node-syncing). It'll probably will continue needing more unimplemented methods so I stopped on just this one.

_JSONArgs on Docker ENTRYPOINT_
```
 - JSONArgsRecommended: JSON arguments recommended for ENTRYPOINT to prevent unintended behavior related to OS signals (line 123)
```

The issue here is that OS signals like SIGTERM and SIGKILL can't be received using shell-like syntax. This could impact graceful shutdowns for example. Unfortunately this can't be solved without losing the ability to set a COOKIE through params (a change added alongside the spawn of a node for connecting to the running instance in docker). Right now this is not a big issue, but in the future we can move to [releases ](https://hexdocs.pm/elixir/config-and-releases.html#releases)


Arkenan (approved):
LGTM, thanks for the changes!

PR #1231, opened 2024-07-18
fix: retrieve a state by slot

### LINKED ISSUE #1228: Getting a state by state slot is not working
This is most likely because of a problem in storing, more than in getting. The keys in the state_db module seem off.

Suggested:

- Move the three kv schemas to different modules: StateByBlockRoot, BlockRootByStateRoot and StateRootBySlot.
- Leave a single "States" module that calls the schema module. Use `fold` for pruning.

Definition of done:
- Each of the KvSchema modules should be individually tested.
- The States module should have integration tests as well (e.g. getting a state by root, pruning).

avilagaston9:
**Motivation**

Retrieving a state by state slot wasn't working because we were using different indexes for storing and retrieving.

**Description**
- Fixes the state retrieval issue.
- Creates StateInfoByRoot, StateRootByBlockRoot, and BlockRootBySlot KV schema modules.
- Makes StateDb use BlockRootBySlot.fold_keys() for state pruning.
- Adds tests for each new KV schema.
- Adds tests for StateDb.

Closes #1228


test/unit/store/state_db_test.exs
```diff
@@ -0,0 +1,135 @@
...
+  # - `slot` incremented by 1.
+  # - A random `block_root`.
+  # - A random `state_root`.
+  defp modify_state_info(state_info) do
+    new_beacon_state = state_info.beacon_state |> Map.put(:slot, state_info.beacon_state.slot + 1)
+    {:ok, new_state_info} = StateInfo.from_beacon_state(new_beacon_state)
+
+    new_state_info
+    |> Map.put(:block_root, Random.root())
+    |> Map.put(:root, Random.root())
```
Arkenan, line 40: Is the second `Map.put` necessary? Isn't the root recalculated when calling `StateInfo.from_beacon_state/1`?

test/unit/store/state_db_test.exs
```diff
@@ -0,0 +1,135 @@
+defmodule Unit.Store.StateDb do
```
Arkenan, line 1: We should add a test for `get_latest_state` (at least with 0, 1, and many states).

Arkenan (approved):
LGTM

test/unit/store/state_db_test.exs
avilagaston9: You are right! Thank you!

test/unit/store/state_db_test.exs
avilagaston9: Done!

PR #1232, opened 2024-07-19
feat: fork choice metrics

### LINKED ISSUE #1211: Add metrics for state transition times.
We need to measure how long does state transition take independently of the full block processing, as the latter may process more than one block.

Arkenan:
Metrics added in a hierarchy:
- handler: block processing, attestation processing, attester slashing.
- transition section: block vs epoch processing.
- operations: randao, attestation batch, withdrawals, etc.

Also added grafana panel with pie chart for execution times for each section. Might eventually move to span profiles if we introduce pyroscope.

Closes #1211


avilagaston9 (approved):
LGTM!

PR #1239, opened 2024-07-23
perf: change checkpoint states db ets

Arkenan:
Changes:

- Makes the CheckpointStates db an ets. 
- We now only request each unique checkpoint state once per attestation batch instead of once per attestation.
- The block processing "benchmark" is now fixed and uses the `ForkChoice.process_block` function directly. 
- The block processing benchmark now has an option to run flama (flamegraph generator) instead of benchee was added.

Some results:

- Attestation processing went from about 4 minutes to around 5 seconds.
- Each ets state fetch takes about 0.1 to 0.3 seconds, as the state is a very big object. We may benefit from deleting the ets table all together and having the checkpoint states in a local variable.


lib/lambda_ethereum_consensus/store/checkpoint_states_db.ex
```diff
@@ -41,18 +45,20 @@ defmodule LambdaEthereumConsensus.Store.CheckpointStates do
...
-
-      :not_found ->
-        with {:ok, state} <- compute_target_checkpoint_state(checkpoint.epoch, checkpoint.root) do
-          put(checkpoint, state)
-          {:ok, state}
-        end
+    case LRUCache.get(
+           @table,
+           checkpoint,
+           fn checkpoint -> compute_target_checkpoint_state(checkpoint.epoch, checkpoint.root) end
```
avilagaston9, line 51: In this line, we need to do something like this:
```elixir
with {:ok, value} <- compute_target_checkpoint_state(checkpoint.epoch, checkpoint.root) do
    value
end
```
Or to change the return type of `compute_target_checkpint_state`, otherwise we will have doubly indented results: 
```elixir
{:ok, {:ok, value}}
````

avilagaston9 (approved):
LGTM! You just need to add `CheckpointStates` to [this line](https://github.com/lambdaclass/lambda_ethereum_consensus/blob/8f0019ca2757c6771308b1824353df05af10ca14/test/spec/tasks/generate_spec_tests.ex#L79) and check the comment below for the spec tests!

lib/lambda_ethereum_consensus/store/checkpoint_states_db.ex
Arkenan: Changed the whole code, as it's now an ets.

PR #1240, opened 2024-07-26
fix: add a check for pending blocks at node startup

### LINKED ISSUE #1235: If recovered history is old, node fails to start fork choice
Sometimes the db records are found and considered "recent history", so the node doesn't do checkpoint sync. However, the anchor block and state may be too old, and blocks may be discarded when going to fork choice handlers. That in turn makes no block be transitioned, so we can't perform state transition.

Arkenan:
Changes:

- Makes `process_blocks` public in `PendingBlocks` 
- Calls `process_blocks` once at app startup in `Libp2pPort` at `handle_continue`, so pending blocks from prior executions are processed if considered recent.

Closes #1235


avilagaston9 (approved):
LGTM!

PR #1242, opened 2024-07-29
refactor: merge metrics with PromEx

### LINKED ISSUE #903: Merge `prometheus_ex` and `telemetry_metrics_prometheus` clients
See #902

avilagaston9:
**Motivation**

We were using two Prometheus clients ([prometheus_ex](https://github.com/deadtrickster/prometheus.ex) and [telemetry_metrics_prometheus](https://github.com/beam-telemetry/telemetry_metrics_prometheus)).

**Description**

Merge our BEAM and custom metrics using [PromEx](https://github.com/akoutmos/prom_ex):

- Configures the `PromEx` module to start a standalone HTTP server to serve the aggregated metrics .
- Removes `prometheus_ex`, `telemetry_metrics_prometheus` and their dependencies.
- Sets the `PromEx.Beam` plugin, which captures metrics regarding the Erlang Virtual Machine.
- Exports the beam.js pre-built dashboard and uses it.
- Defines a custom plugin with all the metrics that we were using for the `node` dashboard.
- Removes the old BEAM dashboards since they are replaced by the pre-built one.
- Modifies the `--metrics` and `--metrics-port` flags to have the same logic as the beacon-api flags.

As a result, it is no longer necessary to start the `BeaconApi` to see the BEAM metrics. They are all aggregated in the PromEx server.

Closes #903


rodrigo-o (approved):
LGTM! 

A small comment for kurtosis, given that we need to setup the dashboards with `make kurtosis.setup.grafana` for it to be taken, the current version in main of the make task just add the new dashboards without removing the old ones, we could make this change to take that into account:

```make
kurtosis.setup.grafana:
	[ -d  $(KURTOSIS_GRAFANA_DASHBOARDS_DIR)/lambdaconsensus ] && \
		rm $(KURTOSIS_GRAFANA_DASHBOARDS_DIR)/lambdaconsensus/* || \
		mkdir $(KURTOSIS_GRAFANA_DASHBOARDS_DIR)/lambdaconsensus
	cp -r $(GRAFANA_DASHBOARDS_DIR)/* $(KURTOSIS_GRAFANA_DASHBOARDS_DIR)/lambdaconsensus
```

Issue #1168, opened 2024-06-25
[Tracking] Removal of GenServers


Arkenan:
This is a broad PR to track the removal of all GenServers before re-designing the architecture.

- [x] ForkChoice
- [x] Attestation
- [x] BeaconBlock
- [x] BlobSideCar
- [x] OperationsCollector
- [x] Metadata
- [x] Peerbook
- [x] BeaconChain: can be replaced by fork choice store db calls.
- [x] PendingBlocks. The more complex replacement is the periodic scheduled tasks and the fact that downloading awaits for an async process.
   - [x] #1122 
   - [x] #1121
- [x] IncomingRequestsReceiver: creates handler tasks. We need to remove the set_handler + the tasks should be started from libp2p port directly using a function in the handler module: #1192 
- [x] ExecutionChain: deposit tree is tricky. The rest is trivial. #1193 
- [x] Validator: maybe replaceable by modules that just react on slots. Calculating duties might be shared instead of repeated. Independent failure might be more complex if this is a needed feature. #1178 
- [x] ValidatorManager: Validators were removed naively touching the least amount of logic or state management. Removing the ValidatorManager may involve a bigger refactor. #1230

rodrigo-o:
After removing the Validator GenServers, it was necessary to add one GenServer (instead of the previous N) for the ValidatorManager

Arkenan:
Completed after #1244 🚀

PR #1244, opened 2024-07-30
refactor: validator manager genserver removal

### LINKED ISSUE #1230: Remove ValidatorManager GenServer
This was temporarily added to make the removal of the Validators GenServers as straightforward as possible. This allowed us to see (at least in kurtosis) that the execution of multiple Validators as they are now (naively executed sequentially, with duplicated work and state as they were before as GenServers) doesn't appear to significantly delay execution steps (but this should be properly measured).

rodrigo-o:
**Motivation**

This is part of the ongoing effort of removing all GenServers to redefine the distributed design of the node.

**Description**

This PR did the following:
- Remove the Clock completely
- Remove the ValidatorManager Genserver and make it just a collection of setup functions
  - To minimize changes in the execution of validators, this PR does not deal with the well-needed state management refactor, it just moves the current way of managing state from ValidatorManager to LibP2P
  - To avoid storing the state in the DB (which may be decided in the validator state management refactor) we added a quick workarround for LibP2P to notify blocks thorugh :handle_info, this will probably be removed in the future.
  - `LambdaEthereumConsensus.Validator.Setup` was the previous `LambdaEthereumConsensus.Validator.ValidatorManager`,  it's not a Genserver anymore and the notify functions were moved, but the setup code (functions still there) are the same, it hasn't changed but the rename shows everything as new.
- Some fixes for the Validators execution
  - Recalculate duties just at the end of an epoch instead of at any new block
  - workaround implemented to deal with the duplicated attestations generated by the syncroneous execution (when removing casts from the ValidatorManager)
  
Resolves #1230 and is the last task of #1168


rodrigo-o:
**HOW TO TEST**

Undortunately right now testing validators is a manual endevour, so to make sure this works as expected, [after following the initial setup of kurtosis explained in the README](https://github.com/lambdaclass/lambda_ethereum_consensus?tab=readme-ov-file#testing-environment-with-kurtosis) a simple `make kurtosis.clean-start` should start kurtosis with the default config. 

Kurtosis comes with the Dora explorer, check the ports in Docker and got to `http://localhost:{dora port}/slots`, this should show the following:

image

If our node is green at proposals, the chain progress, and we can see attestation there, everything is fine. There are still a couple of things to check (like Sync Aggregate being 0, or duplicate aggregates), but that was the current behavior before the GenServer removal and will be tackled in a follow-up PRs.

lib/libp2p_port.ex
```diff
@@ -688,4 +729,85 @@ defmodule LambdaEthereumConsensus.Libp2pPort do
...
+
+  defp on_tick(time, %{genesis_time: genesis_time, slot_data: slot_data} = state) do
+    # TODO: we probably want to remove this (ForkChoice.on_tick) from here, but we keep it
+    # here to have this serialized with respect to the other fork choice store modifications.
+
+    ForkChoice.on_tick(time)
+
+    new_slot_data = compute_slot(genesis_time, time)
+
+    updated_state = maybe_tick_validators(slot_data != new_slot_data, new_slot_data, state)
```
Arkenan, line 743: nit: I may be a boomer here, but I feel that a simple if might be easier to follow, like so:

```elixir
if slot_data == new_slot_data do
  state
else
  updated_validators = notify_validators(state.validators, {:on_tick, new_slot_data})
  %{state | slot_data: new_slot_data, validators: updated_validators}
end
```

lib/libp2p_port.ex
```diff
@@ -688,4 +729,85 @@ defmodule LambdaEthereumConsensus.Libp2pPort do
...
+      "[Validator Manager] #{inspect(msg)} notified to all Validators after #{end_time - start_time} ms"
+    )
+
+    updated_validators
+  end
+
+  defp notify_validator({pubkey, validator}, {:on_tick, slot_data}),
+    do: {pubkey, Validator.handle_tick(slot_data, validator)}
+
+  defp notify_validator({pubkey, validator}, {:new_block, slot, head_root}),
```
Arkenan, line 777: Maybe we can have all of this stuff in a Validator module?

Arkenan (approved):
LGTM with some style comments. Thanks for taking this non trivial PR!

lib/lambda_ethereum_consensus/validator/setup.ex
```diff
@@ -0,0 +1,82 @@
...
+    This function expects two files for each validator:
+      - <keystore_dir>/<public_key>.json
+      - <keystore_pass_dir>/<public_key>.txt
+  """
+  @spec decode_validator_keys(binary(), binary()) ::
+          list({Bls.pubkey(), Bls.privkey()})
+  def decode_validator_keys(keystore_dir, keystore_pass_dir)
+      when is_binary(keystore_dir) and is_binary(keystore_pass_dir) do
+    File.ls!(keystore_dir)
+    |> Enum.map(fn filename ->
```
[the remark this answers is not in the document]
rodrigo-o: Yes totally agree, this hasn't changed really, it's marked as change because of the rename and github doesn't track that so good on the diffs. In git though, it shows this was the same as the previous iteration. 

That said I totally agree with you and I even already refactored this in the PR I'm working on ^^, I had this 2 already extracted:

```elixir
  defp decode_key({keystore_file, keystore_pass_file}) do
    # TODO: remove `try` and handle errors properly
    try do
      Keystore.decode_from_files!(keystore_file, keystore_pass_file)
    rescue
      error ->
        Logger.error(
          "[Validator] Failed to decode keystore file: #{keystore_file}. Pass file: #{keystore_pass_file} Error: #{inspect(error)}"
        )

        nil
    end
  end

  defp paths_from_filename(keystore_dir, keystore_pass_dir, filename, ".json") do
    basename = Path.basename(filename, ".json")

    keystore_file = Path.join(keystore_dir, "#{basename}.json")
    keystore_pass_file = Path.join(keystore_pass_dir, "#{basename}.txt")

    {keystore_file, keystore_pass_file}
  end

  defp paths_from_filename(_keystore_dir, _keystore_pass_dir, basename, _ext) do
    Logger.warning("[Validator] Skipping file: #{basename}. Not a json keystore file.")
    nil
  end
```

lib/lambda_ethereum_consensus/validator/validator.ex
```diff
@@ -300,12 +297,16 @@ defmodule LambdaEthereumConsensus.Validator do
...
+
     aggregation_bits =
-      attestations
+      unique_attestations
       |> Stream.map(&Map.fetch!(&1, :aggregation_bits))
       |> Enum.reduce(&BitField.bitwise_or/2)
 
-    {:ok, signature} = attestations |> Enum.map(&Map.fetch!(&1, :signature)) |> Bls.aggregate()
+    {:ok, signature} =
+      unique_attestations |> Enum.map(&Map.fetch!(&1, :signature)) |> Bls.aggregate()
```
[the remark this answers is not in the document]
rodrigo-o: Yes totally! I tried to make the smallest amount of changes to the logic in this one, I saw a couple of opportunities to improve code clarity but didn't wanted to go that rabbit hole yet, I'm in it now though xD, so I'll take this one into account for the current PR!

lib/libp2p_port.ex
rodrigo-o: Sure, I saw some `maybe_` prefixed functions and feel somewhat odd when variables are equated to an if expression xD `updated_state = if ...` but thats just me xD, already moved to the good ol' `if` ^^

lib/libp2p_port.ex
```diff
@@ -468,6 +502,13 @@ defmodule LambdaEthereumConsensus.Libp2pPort do
     {:noreply, new_state}
   end
 
+  @impl GenServer
+  def handle_info({:new_block, slot, head_root}, %{validators: validators} = state) do
```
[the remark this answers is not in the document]
rodrigo-o: I changed the name to avoid confusions; my bad, good call!

lib/libp2p_port.ex
```diff
@@ -104,9 +111,12 @@ defmodule LambdaEthereumConsensus.Libp2pPort do
...
 
-  @spec on_tick(Types.uint64()) :: :ok
-  def on_tick(time) do
-    GenServer.cast(__MODULE__, {:on_tick, time})
+  @spec notify_new_block(Types.slot(), Types.root()) :: :ok
+  def notify_new_block(slot, head_root) do
+    # TODO: This is quick workarround to notify the libp2p port about new blocks from within
+    # the ForkChoice.recompute_head/1 without moving the validators to the store this
+    # allows to deferr that move until we simplify the state and remove duplicates.
+    send(self(), {:new_block, slot, head_root})
```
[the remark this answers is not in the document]
rodrigo-o: I even had an initial version with an if regarding that (checking if `whereis(__MODULE__)` was or not `self()`), but I thought it was too much, I added that note, missed it in the comment when removed the if.

lib/libp2p_port.ex
rodrigo-o: I was moving this to a new module that represents the state of the pool of validators, but agree that we could just move it to the Validator module in the meantime (thats on me for renaming the ValidatorManger xD), I'll move them!

rodrigo-o: Moved it again to `Validator.Setup`, to remove it from libp2p, it will be better handled in the follow-up PR.

PR #1245, opened 2024-07-31
perf: compute all committees at once at the beginning of an epoch.

Arkenan:
Changes:

- Add `compute_all_committees` function that shuffles the full list of indices and then split the individual committees, which is more efficient than calculating each committee.
- Add documentation for everything related to committees.
- (unrelated) Refactor a date function for filenames in benchmarks and profile jobs.


lib/lambda_ethereum_consensus/utils.ex
```diff
@@ -47,4 +47,27 @@ defmodule LambdaEthereumConsensus.Utils do
...
+  def chunk_by_sizes(enum, [], [], 0, all_chunks), do: [enum | Enum.reverse(all_chunks)]
+
+  def chunk_by_sizes(enum, [size | rem_sizes] = sizes, chunk, chunk_size, all_chunks) do
+    if chunk_size == size do
+      chunk_by_sizes(enum, rem_sizes, [], 0, [Enum.reverse(chunk) | all_chunks])
+    else
+      [elem | rem_enum] = enum
+      chunk_by_sizes(rem_enum, sizes, [elem | chunk], chunk_size + 1, all_chunks)
+    end
+  end
```
rodrigo-o, line 72: I may be missing some case, just tested it quickly, but i think we can simplify this to something like:
```elixir
  def chunk_by_sizes(enum, sizes) do
    Enum.reduce(sizes, {enum, []}, fn size, {remaining, acc} ->
      {chunk, rest} = Enum.split(remaining, size)
      {rest, [chunk | acc]}
    end)
    |> elem(1)
    |> Enum.reverse()
  end
```

I don't think we'll gain anything performance-wise, but its easier to read I guess.

rodrigo-o (approved):
LGTM! Kudos to the comments, they add a lot of clarity to the way everything is calculated, specially regarding the different type of indices, and the role of every argument. Added a small comment regarding a function simplification but it's just a matter of preference!

lib/lambda_ethereum_consensus/utils.ex
Arkenan: It's definitely easier to read! It's also similar how I implemented it originally, and it had some performance problems, which is why I did it manually. I'll merge and review it later on though! Maybe I did something wrong in the first try.

PR #1246, opened 2024-07-31
feat: implement KeyManager API

### LINKED ISSUE #1110: Implement KeyManager API to be able to integrate to EthDocker
https://ethereum.github.io/keymanager-APIs/#/

### LINKED ISSUE #1256: Implement Local Key Manager set of endpoints
https://ethereum.github.io/keymanager-APIs/#/Local%20Key%20Manager

avilagaston9:
**Motivation**

To be able to integrate with EthDocker.

**Description**

Implements the Local Key Manager set of endpoints: https://ethereum.github.io/keymanager-APIs/#/Local%20Key%20Manager.
- Creates a new Phoenix endpoint with `5000` as the default port.
- Adds Open Api Spex at `GET api/openapi`.
- Creates a `Keystore` struct to hold the keystore data.
- Adds the keystore data to the `Validator` struct.

Now we can list, add and delete Validators from `Libp2pPort`.

**Tracked Issues**

To stay in sync with the team, I've limited this PR to the changes above, which add the basic structure of the Keystore. I've created the following issues for tasks that are outside the scope of this PR.
- #1260
- #1261 
- #1262 
- #1263 
- #1265 
- #1267 
- #1268 
- #1269 

**Observations**
- The [ethereum-package](https://github.com/lambdaclass/ethereum-package) was modified to enable the `validator-port`. If you have already cloned it, a pull may be needed.
- To enable the `validator-port` with Kurtosis, the `keymanager-enabled` parameter must be set in the `network_params.yaml` as specified in their [docs](https://github.com/lambdaclass/ethereum-package).

Closes #1256


lib/libp2p_port.ex
```diff
@@ -530,6 +539,49 @@ defmodule LambdaEthereumConsensus.Libp2pPort do
     {:noreply, state}
   end
 
+  @impl GenServer
+  def handle_call(:get_keystores, _from, %{validators: []} = state),
+    do: {:reply, [], state}
```
Arkenan, line 544: Why is this clause necessary? If the validators list is empty, the enum.map will work, and return an empty list too.

lib/libp2p_port.ex
```diff
@@ -530,6 +539,49 @@ defmodule LambdaEthereumConsensus.Libp2pPort do
...
+
+  @impl GenServer
+  def handle_call(:get_keystores, _from, %{validators: validators} = state),
+    do: {:reply, Enum.map(validators, fn {_pk, validator} -> validator.keystore end), state}
+
+  @impl GenServer
+  def handle_call({:delete_validator, pk}, _from, %{validators: validators} = state) do
+    case Map.fetch(validators, pk) do
```
Arkenan, line 552: Let's use pubkey instead of pk to distinguish these from private keys.

lib/libp2p_port.ex
```diff
@@ -530,6 +539,49 @@ defmodule LambdaEthereumConsensus.Libp2pPort do
...
+      :error ->
+        {:error, "Pubkey #{inspect(pk)} not found."}
+    end
+  end
+
+  @impl GenServer
+  def handle_call({:add_validator, keystore}, _from, %{validators: validators} = state) do
+    # TODO (#1263): handle 0 validators
+    first_validator = validators |> Map.values() |> List.first()
+    validator = Validator.new({first_validator.slot, first_validator.root, keystore})
```
Arkenan, line 567: Nothing to be done in this PR, but I think this shows how weird it is to have these values tied to individual validators. We'll get those values from the store when the other PR is merged.

Arkenan (approved):
LGTM with some comments

lib/libp2p_port.ex
avilagaston9: You are right! It is not necessary.

lib/libp2p_port.ex
rodrigo-o: +1

lib/beacon_api/utils.ex
```diff
@@ -31,6 +31,12 @@ defmodule BeaconApi.Utils do
     "0x" <> Base.encode16(binary, case: :lower)
   end
 
+  def hex_decode("0x" <> binary) do
+    with {:ok, decoded} <- Base.decode16(binary, case: :lower) do
```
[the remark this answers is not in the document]
avilagaston9: Actually, this is used to decode the user input. I have updated its definition to the following:
```elixir
def hex_decode("0x" <> binary), do: Base.decode16(binary, case: :lower)
def hex_decode(binary), do: {:error, "Not valid pubkey: #{inspect(binary)}"}
```
I have also updated the caller function to handle the result properly:
```elixir
        with {:ok, pubkey} <- Utils.hex_decode(pubkey),
             ..., do:
             ...
        else
          {:error, reason} ->
            Logger.error("[Keystore] Error removing key. Reason: #{reason}")
        end
```

lib/lambda_ethereum_consensus/validator/validator.ex
```diff
@@ -31,35 +32,27 @@ defmodule LambdaEthereumConsensus.Validator do
...
   # TODO: Slot and Root are redundant, we should also have the duties separated and calculated
   # just at the begining of every epoch, and then just update them as needed.
   @type state :: %__MODULE__{
           slot: Types.slot(),
           epoch: Types.epoch(),
           root: Types.root(),
           duties: Duties.duties(),
-          validator: validator(),
+          index: non_neg_integer() | nil,
+          keystore: Keystore.t(),
```
rodrigo-o, line 43: The Validator state is weird but this will be a lot better in the coming days. Lot of this is going to change in the refactor I'm working on, I already removed slot, epoch, root and even duties from here, it should be easier by then hopefully.

network_params.yaml
```diff
@@ -2,11 +2,12 @@ participants:
...
+    validator_count: 1
   - el_type: geth
     cl_type: lambda
     cl_image: lambda_ethereum_consensus:latest
     use_separate_vc: false
     count: 1
-    validator_count: 32
-    cl_max_mem: 4096
\ No newline at end of file
+    validator_count: 63
```
rodrigo-o, line 11: np: do we want to have 1 in every lighthouse node and 63 (62 would be sufficient) in ours as the new default config?

rodrigo-o (approved):
LGTM!

lib/lambda_ethereum_consensus/validator/validator.ex
avilagaston9: I agree. I just hope that this doesn't add too much noise to your PR!

PR #1266, opened 2024-08-09
feat: handle keys with no deposits

avilagaston9:
WIP


avilagaston9:
Closed since it'll be much easier to handle once #1274  is merged, if not totally solved.

PR #1274, opened 2024-08-13
refactor: validator state management refactor

### LINKED ISSUE #1255: Share duty calculations and state between validators
Currently, the validator structs hold, individually, data that is shared between all of them. Most notably, epoch, slot and head root.

rodrigo-o:
**Motivation**

After the latest refactors regarding Genserver removal, the state management of validators became more complex than it needed, before it made sense because of everyone beign and independen Genserver, but now most of the state could be shared by a `ValidatorSet`, this PR deals with that.

**Description**

This PR introduces a new `ValidatorSet` module that manage Validators and Duties. It process both new_head notifications as well as ticks throughout the slot lifecycle and select just the validators that need to be involved in every step of the process. The logic is maintained as it were previously implemented, this PR aims to reduce complexity, duplication and enhance clarity but it doesn't go in depth in the current logic, it tries to reuse current functions when available and simplify the ones that are intuitive to do so.

- ValidatorSet creation
- Moving the process of new_heads and ticks from Validators to ValidatorSet
- Simplified the Validator state as much as possible
- Move Duties to work on a per-epoch-basis, i.e. every duty is calculated once per epoch
- ValidatorSet selects only the validators with a duty to accomplish in every slot third, and just them do their work.

_Note: The long commit history is because this branch started out of `validator-manager-genserver-removal` (#1224) before merging it to main._

Resolves #1255 

NEXT STEPS:
- Subnet Info checks (we are not joining anymore to subnets at the duties calculation but everything continue working, this needs to be investigated further)


lib/lambda_ethereum_consensus/validator/validator.ex
```diff
@@ -32,220 +28,93 @@ defmodule LambdaEthereumConsensus.Validator do
...
-    target_root = if slot == start_slot, do: head_root, else: last_root
-
-    # Process the start of the new epoch
-    new_beacon = fetch_target_state(epoch, target_root) |> go_to_slot(start_slot)
-
-    new_duties =
-      Duties.shift_duties(state.duties, epoch, last_epoch)
-      |> Duties.maybe_update_duties(new_beacon, epoch, state.index, state.keystore.privkey)
-
-    move_subnets(state.duties, new_duties)
```
rodrigo-o, line 181: This leaved subnets and joined the new ones. But in the new calculation of duties this isn't done and both publish/receive of attestations for aggregation is working. I'm going to create an issue for further check this.

lib/lambda_ethereum_consensus/validator/validator_set.ex
```diff
@@ -0,0 +1,287 @@
...
+       do: set
+
+  defp compute_duties(set, epoch, slot, head_root) do
+    epochs_to_calculate =
+      [{epoch, slot}, {epoch + 1, Misc.compute_start_slot_at_epoch(epoch + 1)}]
+      |> Enum.reject(&Map.has_key?(set.duties, elem(&1, 0)))
+
+    epochs_to_calculate
+    |> Map.new(&compute_duties_for_epoch!(set, &1, head_root))
+    |> merge_duties_and_prune(epoch, set)
```
rodrigo-o, line 136: This is assuming (as it was done previously) that we can calculate duties ahead of time, this shouldn't be the case because Validators participating in an epoch will be decided at the start of that epoch.

lib/lambda_ethereum_consensus/validator/validator.ex
rodrigo-o: Added the issue: #1279

lib/lambda_ethereum_consensus/validator/validator_set.ex
Arkenan: From the randao chapter in the [eth2book](https://eth2book.info/capella/part2/building_blocks/randomness/#lookahead)

>The RANDAO seed at the end of epoch  N is used to compute validator duties for the whole of epoch N+2. This interval is controlled by [MIN_SEED_LOOKAHEAD](https://eth2book.info/capella/part3/config/preset/#min_seed_lookahead) via the [get_seed()](https://eth2book.info/capella/part3/helper/accessors/#def_get_seed) function. Thus, validators have at least one full epoch to prepare themselves for any duties, but no more than two.

lib/lambda_ethereum_consensus/validator/validator_set.ex
```diff
@@ -0,0 +1,284 @@
...
+        set
+
+      validator_index ->
+        validators
+        |> Map.update!(validator_index, &Validator.start_payload_builder(&1, slot, head_root))
+        |> update_validators(set)
+    end
+  end
+
+  defp propose(%{validators: validators} = set, epoch, slot, head_root) do
```
Arkenan, line 177: nit `maybe_propose`?

lib/lambda_ethereum_consensus/validator/validator_set.ex
```diff
@@ -0,0 +1,284 @@
...
+    %__MODULE__{}
+  end
+
+  defp setup_validators(slot, head_root, keystore_dir, keystore_pass_dir) do
+    validator_keystores = decode_validator_keystores(keystore_dir, keystore_pass_dir)
+    epoch = Misc.compute_epoch_at_slot(slot)
+
+    validators =
+      Map.new(validator_keystores, fn keystore ->
+        validator = Validator.new(keystore, slot, head_root)
```
Arkenan, line 60: Instead of fetching the state internally here, can we fetch the state once in line 57?

lib/lambda_ethereum_consensus/validator/validator_set.ex
```diff
@@ -0,0 +1,284 @@
...
+  end
+
+  @doc """
+  Notify all validators of a new head.
+  """
+  @spec notify_head(t(), Types.slot(), Types.root()) :: t()
+  def notify_head(set, slot, head_root) do
+    Logger.debug("[ValidatorSet] New Head", root: head_root, slot: slot)
+    epoch = Misc.compute_epoch_at_slot(slot)
+
```
Arkenan, line 77: If there are non-initialized validators and we need to do so here, we can fetch the state here too. For a futre PR, not this one, as it's a relatively big change, we might want to include the store in this call, as it possesses the state without need to fetch it.

lib/lambda_ethereum_consensus/validator/validator_set.ex
```diff
@@ -0,0 +1,284 @@
...
+  defp decode_key({keystore_file, keystore_pass_file}) do
+    # TODO: remove `try` and handle errors properly
+    [Keystore.decode_from_files!(keystore_file, keystore_pass_file)]
+  rescue
+    error ->
+      Logger.error(
+        "[Validator] Failed to decode keystore file: #{keystore_file}. Pass file: #{keystore_pass_file} Error: #{inspect(error)}"
+      )
+
+      []
```
Arkenan, line 282: Nit: aren't this functions more related to the Keystore module?

lib/libp2p_port.ex
```diff
@@ -549,46 +550,42 @@ defmodule LambdaEthereumConsensus.Libp2pPort do
...
-      :error ->
+  def handle_call({:delete_validator, pubkey}, _from, %{validator_set: validator_set} = state) do
+    validator_set.validators
+    |> Enum.find(fn {_index, validator} -> validator.keystore.pubkey == pubkey end)
+    |> case do
+      {index, _validator} ->
+        Logger.warning("[Libp2pPort] Deleting validator with index #{inspect(index)}.")
+        updated_validators = Map.delete(validator_set.validators, index)
+        {:reply, :ok, Map.put(state.validator_set, :validators, updated_validators)}
+
```
Arkenan, line 568: Maybe we can have a `ValidatorSet.delete(pubkey)`?

lib/libp2p_port.ex
```diff
@@ -549,46 +550,42 @@ defmodule LambdaEthereumConsensus.Libp2pPort do
...
-     %{
-       state
-       | validators:
-           Map.put(
-             validators,
-             pubkey,
-             validator
-           )
-     }}
+    {:reply, :ok, put_in(state.validator_set, [:validators, validator.index], validator)}
```
Arkenan, line 588: Maybe a `ValidatorSet.add(index, pubkey)?`

Arkenan (approved):
Thanks for the PR rodri, left a bunch of comments. None of them are blocking, feel free to merge as is. Most of them are nits. The only one that I'm doubtful about is the setting of false instead of true.

lib/lambda_ethereum_consensus/validator/duties.ex
```diff
@@ -15,161 +16,125 @@ defmodule LambdaEthereumConsensus.Validator.Duties do
...
+          attester_duties() | proposer_duties()
+        ) :: duties()
+  def update_duties!(duties, kind, epoch, slot, updated),
+    do: put_in(duties, [epoch, kind, slot], updated)
+
+  @spec attested(attester_duty()) :: attester_duty()
+  def attested(duty), do: Map.put(duty, :attested?, true)
+
+  @spec aggregated(attester_duty()) :: attester_duty()
+  def aggregated(duty), do: Map.put(duty, :should_aggregate?, false)
```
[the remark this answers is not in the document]
rodrigo-o: I also saw the same, didn't want to continue changing the duties struct they were already ser with attested? and should_aggregate? probably becase aggregated sounds weird, but given you thought the same as me I'm gonna rename it!

lib/lambda_ethereum_consensus/validator/duties.ex
```diff
@@ -15,161 +16,125 @@ defmodule LambdaEthereumConsensus.Validator.Duties do
...
-
-        {d, _} ->
-          d
-      end)
-
-    [epp | duties]
-  end
+  @spec compute_attesters_for_epoch(BeaconState.t(), Types.epoch(), ValidatorSet.validators()) ::
+          attester_duties_per_slot()
+  def compute_attesters_for_epoch(%BeaconState{} = state, epoch, validators) do
```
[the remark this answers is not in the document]
rodrigo-o: I'm gonna take a look into this!

lib/lambda_ethereum_consensus/validator/validator.ex
```diff
@@ -32,220 +28,93 @@ defmodule LambdaEthereumConsensus.Validator do
...
-    case try_setup_validator(state, head_slot, head_root) do
+    case fetch_validator_index(beacon, state.keystore.pubkey) do
       nil ->
-        # TODO: Previously this was handled by the validator continously trying to setup itself,
-        # but now that they are processed syncronously, we should handle this case different.
-        # Right now it's just omitted and logged.
-        Logger.error("[Validator] Public key not found in the validator set")
+        Logger.warning(
+          "[Validator] Public key #{state.keystore.pubkey} not found in the validator set"
+        )
```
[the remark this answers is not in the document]
rodrigo-o: Agree, I had this in my TODOs when started tackling the PR and forget about it in the end! I'll make sure the change is not big and do it, if it proves to be bigger than a couple of changes I'm gonna create an issue!

lib/lambda_ethereum_consensus/validator/validator_set.ex
rodrigo-o: They are! The Keystore PR was merged in the middle of this one and this was still in the previous `Setup` module instad of the Keystore so I mantained it here, but I agree with you, I created a small issue to move this and also create an `add` and `delete` in the ValidatorSet module. #1281

lib/libp2p_port.ex
rodrigo-o: Sure, I'll added this to #1281 and will tackle it soon!

lib/libp2p_port.ex
rodrigo-o: Sure, #1281

lib/lambda_ethereum_consensus/validator/validator_set.ex
rodrigo-o: I'm sure I had that in place at some point in time and then moved back to do it in the Validator, but yeah, I cant think of a good reason now to not move it out, I'll do it!

lib/lambda_ethereum_consensus/validator/validator_set.ex
```diff
@@ -0,0 +1,284 @@
...
+
+  @type t :: %__MODULE__{
+          head_root: Types.root() | nil,
+          duties: %{
+            Types.epoch() => %{
+              proposers: Duties.proposer_duties(),
+              attesters: Duties.attester_duties()
+            }
+          },
+          validators: validators()
```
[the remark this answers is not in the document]
rodrigo-o: Yes, this isn't handle in the PR. I created an issue for this and linked this and the next comment about this in the notify_head there! #1282

lib/lambda_ethereum_consensus/validator/validator_set.ex
rodrigo-o: Agree, I added a new issue #1282 to handle this.

lib/lambda_ethereum_consensus/validator/validator_set.ex
rodrigo-o: Thanks Tomy! I was wondering about validator's activations and exits when thinking about this, like wouldn't new activations or exits affect already calculated duties with a different set of validators? I might be lacking more knowledge about RANDAO and how the lookahead plays with the calculation of duties for te active set of validators.

lib/lambda_ethereum_consensus/validator/validator_set.ex
rodrigo-o: Sure, should do the same with attest and build payload

lib/lambda_ethereum_consensus/validator/validator.ex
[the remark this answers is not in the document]
rodrigo-o: This will be tackled as part of #1282!

lib/lambda_ethereum_consensus/validator/duties.ex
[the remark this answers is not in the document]
rodrigo-o: I also felt the same, didn't want to continue changing the duties struct they were already set with attested? and should_aggregate?, the case is that should_aggregate? is used both for checking if aggregation is needed (attestation is always needed) and to make sure aggregation happened. We can have two keys, an additional one to check if it was already aggregated, or adding a comment here to explain this. I'll go with the comment for now but I'm open to the other alternative if we believe its necessary.

lib/lambda_ethereum_consensus/validator/validator_set.ex
rodrigo-o: It was my fault, I even had the other new signature in validation without using it, great catch, already solved!

rodrigo-o:
Here is a quick run in kurtosis showing more than one epoch running in Kurtosis with other 64 validators in another clients:

image

lib/lambda_ethereum_consensus/validator/validator_set.ex
rodrigo-o: Done

lib/lambda_ethereum_consensus/validator/duties.ex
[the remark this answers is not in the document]
Arkenan: I actually re-read this and realized that it makes sense that one is true and the other is false, as 

```
attested -> !should_attestate
aggregated -> !should_aggregate
```

lib/lambda_ethereum_consensus/validator/duties.ex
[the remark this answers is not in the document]
Arkenan: Changed my previous comments to `maybe_prefetch_all_committees`, that calculates the committees for a single epoch in a batch IF not already calculated.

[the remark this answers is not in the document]
rodrigo-o: Thanks! I'll add it!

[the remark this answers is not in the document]
rodrigo-o: Done

rodrigo-o:
The latest commit fixes an issue regarding the start of the node with no Validators that I spotted while testing the `maybe_prefetch_committees` addition.

PR #1275, opened 2024-08-13
fix: path generation

avilagaston9:
**Motivation**

The path generation for the Keystore files wasn't working.

**Description**
- Use the encoded version of the `pubkey` to generate the path to the keystore file.
- `@keystore_dir` and `@keystore_pass_dir` were always set to the defaults values since they were being computed at compile time. Replace them with function calls to the `Keystore` module.


rodrigo-o (approved):
LGTM!

PR #1276, opened 2024-08-14
refactor: add ethereum-package as a submodule

### LINKED ISSUE #1270: Add submodule with ethereum-package in the correct branch
Our setup process for the node now includes cloning the ethereum-package for kurtosis manually in a specific branch, which is error prone because the branch is in active development and manual updates may be needed in a devs setup to reflect changes (e.g. go to that separate repo and git pull).

It would be better to have a git submodule pointing the the branch/commit of ethereum-package being used, so if we start using a newer one, that version change would be reflected in version control and users would receive that update automatically in the normal development process.

avilagaston9:
- Adds Lambda's [ethereum-package](https://github.com/lambdaclass/ethereum-package/tree/lecc-integration) fork as a submodule.
- Updates the Makefile to use the submodule.
- Updates README.md

Closes #1270

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "ethereum-package"]
+	path = ethereum-package
+	url = https://github.com/lambdaclass/ethereum-package.git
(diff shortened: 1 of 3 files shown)


rodrigo-o (approved):
LGTM

PR #1277, opened 2024-08-15
refactor: improve checkpoint-sync algorithm

### LINKED ISSUE #1194: Make checkpoint sync recursive and smarter

avilagaston9:
This PR uses the latest `head_slot` as the `start_slot` for requesting blocks when restarting. Previously we were using the `last_finalized_slot` instead and then ignoring those blocks since they had already been processed.


rodrigo-o (approved):
LGTM

Issue #1278, opened 2024-08-16
Use the store in the validator calls


Arkenan:
Some of the date that validators need to calculate duties is already available in the store. This can be integrated as soon as PR #1253 is merged. This also allows us to delete the `compute_target_checkpoint_state` function altogether and only use the one in the store.

Issue #1254, opened 2024-08-07
Duplicated and old aggregate attestations


rodrigo-o:
**Context**

During the work on #1244 duplicate attestations generated Invalid Signature errors at the time of block proposals when trying to generate aggregates (the BLS aggregate signature was computing the same aggregator more than once in the signature generated), this didn't happened before making the `handle_new_block` and `handle_tick` synchronous (i.e. without removing the GenServer and just converting casts to calls in the ValidatorManager the error was reproduced). A quick solution is [already in place](https://github.com/lambdaclass/lambda_ethereum_consensus/blob/d643d2d17c83c7df51bc1070fc81442cdb5046dc/lib/lambda_ethereum_consensus/validator/validator.ex#L300-L301):

```elixir
  # TODO: We need to check why we are producing duplicate attestations, this was generating invalid signatures
  unique_attestations = attestations |> Enum.uniq()
```

**Description**
The purpose of this task is two-fold

1. Understand the root cause of duplicated attestations and fix it
We are receiving duplicates just from ourselves, this error pop-up only when we are publishing attestations, if we just receive them from other nodes we don't have this issue.

2. Understand why we are still adding duplicate aggregate attestations in our block proposals (this was happening before #1244)
This might be related to the previous one, given that we only uniquely filtered them at the aggregation step this behavior might be affecting also aggregates and not just committee attestations. 

Example of a block with duplicated aggregate attestations:

![image](https://github.com/user-attachments/assets/cd509108-fa41-4a0d-bdb6-a3d2c1a20316)

rodrigo-o:
This also happens now with sync_committee_messages as per #1284

rodrigo-o:
Additionally this should tackle the issue that we add attestation for slot-2 to our blocks instead of slot-1 as the other nodes.

PR #1284, opened 2024-08-26
feat: sync committee contribution

### LINKED ISSUE #835: SyncCommiteeContribution Aggregators lookahead
SyncCommitteeMessages are aggregated by sync committee aggregators. A validator that is in a sync committee can check if it's also an aggregator by ussing `get_sync_committee_selection_proof(state, slot, subcommittee_index, privkey) |> is_sync_committee_aggregator(signature)`. This should be looked ahead of time, as it changes every slot.

specs ref: [link](https://github.com/lambdaclass/lambda_ethereum_consensus/blob/main/docs/specs/validator.md#aggregation-selection-1)

### LINKED ISSUE #837: SyncCommitteeContribution construction
A sync aggregator must collect all valid `sync_committee_messages` from the `sync_committee_{subnet_id}` topic during the selected slot that match its own beacon block root, and then turned into a single `SyncCommitteeContribution`. Finally, it needs to be signed to build a `SignedContributionAndProof` struct.

A pseudoelixir code describing this:

```elixir
sync_message =  get_sync_committee_message(state, block_root, validator_index, privkey)

aggregation_bits = 
  sync_committee_messages
  |> Enum.filter(& &1.beacon_block_root == sync_message.beacon_block_root)
  |> Enum.map(& &1.validator_index |> validator_index_to_sync_committee_index())
  |> Enum.reduce(
    BitVector.new(Constants.sync_committee_size), 
    fn idx, bv -> BitVector.set(bv, idx) end)

# Maybe we can add a BitVector.from_index_list(size, index_list) function that does this last step.

contribution = %SyncCommitteeContribution{
  slot: state.slot,
  beacon_block_root: sync_message.beacon_block_root,
  subcommittee_index: subnet_id,
  aggregation_bits: aggregation_bits,
  signature: bls.Aggregate(Enum.map(sync_committee_messages, & &1.signature))
}

cont = get_contribution_and_proof(state, aggregator_index, contribution, privkey)
sig = get_contribution_and_proof_signature(state, cont, privkey)
%SignedContributionAndProof{messsage: cont, signature: sig}
```

Spec ref: [link](https://github.com/lambdaclass/lambda_ethereum_consensus/blob/main/docs/specs/validator.md#construct-sync-committee-contribution)

### LINKED ISSUE #841: SignedContributionAndProof broadcast
After building a `SignedContributionAndProof` in #837, the sync aggregator needs to send it over the `sync_committee_contribution_and_proof` after the second interval of the slot is finished.

refs spec: [link](https://github.com/lambdaclass/lambda_ethereum_consensus/blob/main/docs/specs/validator.md#broadcast-sync-committee-contribution)

rodrigo-o:
**Motivation**

SyncCommitteeMessages are aggregated by sync committee aggregators. This is the next step in the Sync Committee feature addition.

**Description**

This PR adds the calculation of aggregate duties, it's lookahead, and sync contribution production and publish. [[Link to the Validator Spec]](https://github.com/lambdaclass/lambda_ethereum_consensus/blob/main/docs/specs/validator.md#aggregation-selection-1)

Closes #835
Closes #837
Closes #841


rodrigo-o:
Right now we are in a stable position both in the minimal and the main spec, here is an example of a 42 epoch run in minimal:

image
_Here is an example of another node picking up contributions published by ourselves both in the first and last subcommitee_

image
_Here is the whole run still on going_

Makefile
```diff
@@ -52,6 +52,8 @@ KURTOSIS_GRAFANA_DASHBOARDS_DIR ?= $(KURTOSIS_DIR)/static_files/grafana-config/d
 KURTOSIS_COOKIE ?= secret
 # Name of the kurtosis service pointing to the lambdaconsesus node
 KURTOSIS_SERVICE ?= cl-3-lambda-geth
+# Name of the enclave to be used with kurtosis
+KURTOSIS_ENCLAVE ?= lambdanet
```
rodrigo-o, line 56: This change was to be able to run multiple enclaves using the make tasks.

lib/types/subnet_info.ex
```diff
@@ -1,4 +1,4 @@
-defmodule Types.SubnetInfo do
+defmodule Types.AttSubnetInfo do
```
rodrigo-o, line 1: To avoid unnecessary diffs showed here, I didn't yet changed the name of this file. Once the rest of the PR is approved I'll renaming this on as `att_subnet_info`.

lib/lambda_ethereum_consensus/validator/validator_set.ex
```diff
@@ -73,11 +75,12 @@ defmodule LambdaEthereumConsensus.ValidatorSet do
   def notify_head(set, slot, head_root) do
     Logger.debug("[ValidatorSet] New Head", root: head_root, slot: slot)
     epoch = Misc.compute_epoch_at_slot(slot)
+    head_state = fetch_target_state_and_go_to_slot(epoch, slot, head_root)
```
rodrigo-o, line 78: After rethinking, this is unnecessary here. We need it for attestations, but we can fetch it at the attest funtion where we already know if there are validators to attest or not.

lib/lambda_ethereum_consensus/validator/validator_set.ex
```diff
@@ -103,14 +106,18 @@ defmodule LambdaEthereumConsensus.ValidatorSet do
   end
 
   defp process_tick(%{head_root: head_root} = set, epoch, {slot, :second_third}) do
+    head_state = fetch_target_state_and_go_to_slot(epoch, slot, head_root)
```
rodrigo-o, line 109: Same as the previous comment.

lib/lambda_ethereum_consensus/validator/validator_set.ex
```diff
@@ -254,6 +285,37 @@ defmodule LambdaEthereumConsensus.ValidatorSet do
     |> then(&%{set | duties: &1})
   end
 
+  ##########################
+  # Target State
+
+  @spec fetch_target_state_and_go_to_slot(Types.epoch(), Types.slot(), Types.root()) ::
+          Types.BeaconState.t()
+  def fetch_target_state_and_go_to_slot(epoch, slot, root) do
```
rodrigo-o, line 293: This was removed from the validator and placed here.

lib/lambda_ethereum_consensus/validator/validator.ex
```diff
@@ -223,42 +200,144 @@ defmodule LambdaEthereumConsensus.Validator do
...
+      slot: List.first(messages).slot,
+      beacon_block_root: List.first(messages).beacon_block_root,
+      subcommittee_index: subnet_id,
+      aggregation_bits: aggregation_bits,
+      signature: aggregate_sync_committee_signature(messages, indices_in_subcommittee)
+    }
+  end
+
+  defp aggregate_sync_committee_signature(messages, indices_in_subcommittee) do
+    # TODO: as with attestations, we need to check why we recieve duplicate sync messages
```
rodrigo-o, line 311: This is related to #1254

lib/lambda_ethereum_consensus/validator/duties.ex
```diff
@@ -27,26 +26,38 @@ defmodule LambdaEthereumConsensus.Validator.Duties do
...
-          # Given that we send messages in EVERY slot, instead
-          # of tracking them all, we just keep track of the last
-          # slot we broadcasted to avoid double publishs.
-          last_slot_broadcasted: Types.slot(),
+          broadcasted?: boolean(),
+          message_domain: Types.domain(),
+          validator_index: Types.validator_index(),
           subnet_ids: [Types.uint64()],
-          validator_index: Types.validator_index()
+          aggregation: [sync_committee_aggregator_duty()]
```
rodrigo-o, line 41: Duties haven't been refactored, if we agreed with this way to represent them we might want to move attestation duties to follow this same pattern, separating the main attestation from its aggregation duty. This shouldn't impact the code a lot but should be easier to read and clearly separate the two kind of duties.

lib/lambda_ethereum_consensus/validator/duties.ex
```diff
@@ -135,6 +144,109 @@ defmodule LambdaEthereumConsensus.Validator.Duties do
...
+        Map.update(acc, slot, [sync_committee], &[sync_committee | &1])
+    end
+  end
+
+  # Recomputes the sync committee duties for the given epoch without recalculating subnet_ids and
+  # ignoring validators already known to be outside the sync committee.
+  #
+  # Unfortunatelly, extracting the common logic between this function and `compute_sync_committee_duties`
+  # directly impacts readability.
+  defp recompute_sync_committee_duties(
```
rodrigo-o, line 188: I didn't give it a lot of thought to avoid delaying the PR, but accept ideas ^^

lib/lambda_ethereum_consensus/validator/validator_set.ex
rodrigo-o: Done

lib/lambda_ethereum_consensus/validator/validator_set.ex
rodrigo-o: Done

lib/lambda_ethereum_consensus/validator/duties.ex
```diff
@@ -73,21 +84,26 @@ defmodule LambdaEthereumConsensus.Validator.Duties do
...
+              |> recompute_sync_committee_duties(beacon, epoch, validators)
+              |> then(&{&1, misc_data(duties_map, last_epoch)})
 
             {:not_computed, period} ->
               Logger.debug("[Duties] Computing sync committees for period: #{period}.")
 
-              compute_current_sync_committees(beacon, validators)
+              beacon
+              |> compute_sync_committee_duties(epoch, validators)
+              |> then(&{&1, compute_misc_data(beacon, epoch)})
```
Arkenan, line 99: If this is the heavy one that calculates all participants for an epoch we probably shouldn't do it inside this for.

Arkenan (approved):
Hi Rodri, thanks for the PR. Here are some comments. Most are style changes, although there are some performance concerns that we already talked about offline. Feel free to open new issues for those, + TODOs in code referring to them.

lib/lambda_ethereum_consensus/validator/validator_set.ex
```diff
@@ -197,12 +201,36 @@ defmodule LambdaEthereumConsensus.ValidatorSet do
     end
   end
 
+  defp maybe_publish_sync_aggregates(set, slot) do
+    # Sync committee is broadcasted for the next slot, so we take the duties for the correct epoch.
+    epoch = Misc.compute_epoch_at_slot(slot + 1)
+    %{sync_subcommittee_participants: participants} = Duties.misc_data(set.duties, epoch)
```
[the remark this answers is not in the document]
rodrigo-o: Probably bad naming (misc was from miscellaneous), maybe `shared` or `common`? The idea is to have a place for common cached calculations not associated to a particular slots/duty, they could be recalculated at will (every epoch, every period, etc) and may be a way to avoid recalculating costly operations, for example, the map we were discussing regarding validator indices could be placed here as well

test/unit/sync_subnet_info.exs
```diff
@@ -0,0 +1,78 @@
...
+  test "stop collecting with one attestation" do
+    subnet_id = 1
+
+    expected_message = %SyncCommitteeMessage{
+      slot: 5_057_010_135_270_197_978,
+      beacon_block_root:
+        <<31, 38, 101, 174, 248, 168, 116, 226, 15, 39, 218, 148, 42, 8, 80, 80, 241, 149, 162,
+          32, 176, 208, 120, 120, 89, 123, 136, 115, 154, 28, 21, 174>>,
+      validator_index: 0,
+      signature: <<>>
```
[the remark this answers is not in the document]
rodrigo-o: Yes for sure, just wanted to be more explicit with the `expected` but it could be misleading, I'll change it.

lib/lambda_ethereum_consensus/validator/duties.ex
```diff
@@ -27,26 +26,38 @@ defmodule LambdaEthereumConsensus.Validator.Duties do
...
-          last_slot_broadcasted: Types.slot(),
+          broadcasted?: boolean(),
+          message_domain: Types.domain(),
+          validator_index: Types.validator_index(),
           subnet_ids: [Types.uint64()],
-          validator_index: Types.validator_index()
+          aggregation: [sync_committee_aggregator_duty()]
         }
 
+  @type misc_data :: %{sync_subcommittee_participants: %{}}
```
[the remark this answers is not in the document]
rodrigo-o: Probably bad naming (misc was from miscellaneous), maybe shared or common? The idea is to have a place for common cached calculations not associated to a particular slots/duty, they could be recalculated at will (every epoch, every period, etc) and may be a way to avoid recalculating costly operations, for example, the map we were discussing regarding validator indices could be placed here as well

lib/lambda_ethereum_consensus/validator/validator_set.ex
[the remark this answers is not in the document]
rodrigo-o: related to what misc_data means I already answered [here](https://github.com/lambdaclass/lambda_ethereum_consensus/pull/1284#discussion_r1752397670), but aggree that we can have specific function for the specific calculations of the shared data. I'll add it.

lib/lambda_ethereum_consensus/validator/validator_set.ex
```diff
@@ -254,6 +284,37 @@ defmodule LambdaEthereumConsensus.ValidatorSet do
     |> then(&%{set | duties: &1})
   end
 
+  ##########################
+  # Target State
+
+  @spec fetch_target_state_and_go_to_slot(Types.epoch(), Types.slot(), Types.root()) ::
+          Types.BeaconState.t()
+  def fetch_target_state_and_go_to_slot(epoch, slot, root) do
```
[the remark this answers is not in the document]
rodrigo-o: I'll check if there issue is created!

lib/lambda_ethereum_consensus/validator/validator.ex
```diff
@@ -223,42 +200,144 @@ defmodule LambdaEthereumConsensus.Validator do
...
+          Types.slot()
+        ) :: :ok
+  def publish_sync_aggregate(
+        %{index: validator_index, keystore: keystore},
+        %{validator_index: validator_index} = duty,
+        sync_subcommittee_participants,
+        slot
+      ) do
+    for %{subcommittee_index: subnet_id} = aggregation_duty <- duty.aggregation do
+      case Gossip.SyncCommittee.stop_collecting(subnet_id) do
```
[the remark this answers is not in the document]
rodrigo-o: Yes, this is exactly what happens, we discussed this briefly on Friday's sync if I remember correctly, this happens also with attestations and that's why i didn't wanted to tackle it yet and just leave it as it was, here is an issue to handle it correctly: #1286 (BTW this is a p2 because outside of kurtosis we have very low probabilities to have more than one of our aggregators in a particular slot)

lib/types/subnet_info.ex
rodrigo-o: Yeah I know, but it doesn't work, locally git recognizes the rename, it's github who instead do the create/delete, I didn't dig enough but I'm not the only one ^^: 

https://github.com/orgs/community/discussions/43177#discussioncomment-5021208

rodrigo-o: Yeah I know, the issue I saw a couple of days ago (on the previous PR) was that locally git recognized the rename, but the Github UI instead do the create/delete, I didn't dig enough but some other people had the same issue: 

https://github.com/orgs/community/discussions/43177#discussioncomment-5021208

It appears to be related to changes after the rename, or passing a certain changes threshold that Github reinterprets as a create/delete, maybe in this case it's not an issue because I already modified the file before the rename.

lib/lambda_ethereum_consensus/validator/validator_set.ex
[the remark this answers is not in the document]
rodrigo-o: If I'm correct, this issue you created is related to that: #1278

lib/lambda_ethereum_consensus/validator/validator.ex
rodrigo-o: Added to the comment itself

lib/lambda_ethereum_consensus/validator/validator.ex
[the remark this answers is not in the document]
rodrigo-o: Added the TODO comment and the task in it.

lib/lambda_ethereum_consensus/p2p/gossip/sync_committee.ex
```diff
@@ -2,26 +2,126 @@ defmodule LambdaEthereumConsensus.P2P.Gossip.SyncCommittee do
...
+    end
+
+    :ok
+  end
+
+  @spec stop_collecting(non_neg_integer()) ::
+          {:ok, list(Types.Attestation.t())} | {:error, String.t()}
+  def stop_collecting(subnet_id) do
+    # TODO from Attestation: implement some way to unsubscribe without leaving the topic
+    # TODO: This handle individual subnet_id while the other ones handle lists.
```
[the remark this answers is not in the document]
rodrigo-o: Done!

lib/lambda_ethereum_consensus/validator/duties.ex
rodrigo-o: This is indeed inside the epoch for `for {epoch, slot} <- epochs_and_start_slots, reduce: duties_map do`, which normally is just one epoch and two at the beginning when we don't have any lookahead. But it's also inside a case:

```elixir
   case sync_committee_compute_check(epoch, {last_epoch, Map.get(duties_map, last_epoch)}) do
      {:already_computed, sync_committee_duties} ->
        sync_committee_duties
        |> recompute_sync_committee_duties(beacon, epoch, validators)
        |> then(&{&1, misc_data(duties_map, last_epoch)})

      {:not_computed, period} ->
        Logger.debug("[Duties] Computing sync committees for period: #{period}.")

        beacon
        |> compute_sync_committee_duties(epoch, validators)
        |> then(&{&1, compute_misc_data(beacon, epoch)})
    end
```

sync_committee_compute_check/2, makes sure we didn't change epoch since our previous epoch duty calculation, it just hit the  `{:not_computed, period} ->` when we get into a new period. Right now is the only thing that's calculated, but I'll rename the function to let us know its just calculating the participants.

lib/lambda_ethereum_consensus/validator/validator_set.ex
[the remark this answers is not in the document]
rodrigo-o: Also added a comment!

lib/lambda_ethereum_consensus/validator/duties.ex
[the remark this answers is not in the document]
rodrigo-o: I moved on with `shared` but added the particular `sync_subcommittee_participant` functions as discused previously.

lib/lambda_ethereum_consensus/validator/validator_set.ex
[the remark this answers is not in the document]
rodrigo-o: Created the sync_subcommittee_participants function!

lib/lambda_ethereum_consensus/validator/duties.ex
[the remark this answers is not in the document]
rodrigo-o: Also added a comment related to this type to be clear of the intentions.

lib/lambda_ethereum_consensus/validator/duties.ex
rodrigo-o: renamed to `compute_sync_subcommittee_participants` as discussed

lib/types/subnet_info.ex
rodrigo-o: Done!

lib/lambda_ethereum_consensus/validator/utils.ex
```diff
@@ -82,23 +88,72 @@ defmodule LambdaEthereumConsensus.Validator.Utils do
...
+  def sync_subcommittee_participants(state, epoch) do
+    state
+    |> Accessors.get_sync_committee_for_epoch!(epoch)
+    |> Map.get(:pubkeys)
+    |> Enum.chunk_every(Misc.sync_subcommittee_size())
+    |> Enum.with_index()
+    |> Map.new(fn {pubkeys, subcommittee_i} ->
+      pubkeys
+      |> Enum.with_index()
+      |> Enum.group_by(&fetch_validator_index(state, elem(&1, 0)), &elem(&1, 1))
```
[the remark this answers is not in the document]
rodrigo-o: Already created an issue for this: #1290. This happens once per period (i.e. every 4 epochs) fetch_index_validator is only used also at the creation of a new validator so this shouldn't impact any other place. We should be mindful of this and probably measure it in the future and add that map.

PR #1285, opened 2024-09-03
refactor: delete duplicated logic and wrap DBLevel calls

### LINKED ISSUE #1191: Macro the db.ex function wrappers
Most wrappers in the `db.ex` have the following pattern:

```elixir
def some_function(some_args) do
  ref = GenServer.call(@registered_name, :get_ref)
  Exleveldb.some_function(some_args)
end
```

This makes adding each function in leveldb a bit of an unnecessary effort. Ideally we should have a macro that allowed us to do something like this:

```elixir
wrapper function_a(a_args)
wrapper function_b(b_args)
```

That automatically did the boilerplate for a single function. This ticket will be finished when both the macro is coded and also all functions in Exleveldb are wrapped.

leobz:
**Motivation**

Wrap `Exleveldb` and `:eleveldb` calls into `Db` module to centralize logic, and refactor `Db` module to avoid duplicated code and decrease effort by adding leveldb functions.

**Description**

- Refactor `Db` module, extracting repeated logic to `perform/2` function
- Replace `Exleveldb` and `:eleveldb` calls by adding the functions to `Db` module

- TODO: Add all the features of `Exleveldb` with tests
- TODO: Implement macros if necessary

**Closes** #1191


Arkenan (changes requested):
Hi Leo! Thanks a lot for the PR, it's very much appreciated. Welcome to the consensus journey!

Regarding the solution, I don't see much value in the `perform` function. I think it makes code more obscure and error prone (converts a compile time function check into an execution time check). It also does not reduce code length, every function is still there.

However, I do like the `ref()` idea. Maybe we can simplify this PR, remove the `perform` function and just put `ref()` in every call? That would, effectively, reuse the part that is common to all functions in the `Db` module.

Arkenan:
Hi @leobz thanks for the update! I really like the code. Marked it ready for review and run CI. It seems to be doing pretty well, but dialyzer (type analyzer) failed. You may debug that locally by running `make dialyzer` and seeing what's wrong. At a first glance, it seems that the typespec for the `Db.iterator_move` function is too restrictive or doesn't match the contract.

leobz:
Thank you @Arkenan! Sorry for the delay, I wasn't available these weeks, but I'm going to fix it :raised_hands:

Arkenan (approved):
LGTM, thanks for taking this issue and congrats on your first PR on the repo!

PR #1293, opened 2024-09-13
refactor: moved add and remove logic to ValidatorSet and moved out keystore fuctions

### LINKED ISSUE #1281: ValidatorSet and Keystore related refactor
Move Keystore functions from the ValidatorSet to the Keystore and also create and add and delete functione to avoid validatorset logic in LibP2P

rodrigo-o:
**Motivation**

Refactor some logic to be in it's proper place

**Description**

This PR moves Keystore functions from the ValidatorSet to the Keystore and also create and add and delete function in the ValidatorSet to avoid foreign logic in LibP2P. It also re adds the slot to the ValidatorSet struct and update them (slot + root) always to be able to instantiate the new validator in the correct slot & root.

Closes #1281


Arkenan (approved):
LGTM

PR #1316, opened 2024-10-02
fix: added MAXIMUM_GOSSIP_CLOCK_DISPARITY into account to distinguish future slots

### LINKED ISSUE #1315: Blocks being ignored because we transition the slot a couple of milliseconds later compared to peers (Sepolia long run)
This issue was spot on muntiple test, and was close in time to the ocurrence of #1309. After solving it in a test branch, the issue doesn't impacted on #1309 issue, but it's still a bug so this issue is created to track it and solve it. The problem is apparently related to the `MAXIMUM_GOSSIP_CLOCK_DISPARITY` not being taken into account. It's a P1 because it appears to be the cause of other issues, but i wasn't able to pinpoint them yet, given that the solution is pretty straightforward I preferred to tackle it now.

rodrigo-o:
**Motivation**

This PR fixes an issue where some blocks where ignored due to one of two checks:
`ERROR 15:41:48.960 [Fork choice] Failed to add block: block is from the future slot=5970209 root=0xd42..6485`
`WARNING 21:11:11.831 [Gossip] Block ignored, reason: "Block is from the future: block.slot=6000656. Current slot: 6000655.".`

**Description**
In fact, the blocks where being received in time, just a couple of milliseconds before we made the transition, having into account `MAXIMUM_GOSSIP_CLOCK_DISPARITY` in the calculation, allow us to process the block at the time.

Resolves #1315

### DIFF
--- a/lib/lambda_ethereum_consensus/fork_choice/fork_choice.ex
+++ b/lib/lambda_ethereum_consensus/fork_choice/fork_choice.ex
@@ -31,7 +31,7 @@ defmodule LambdaEthereumConsensus.ForkChoice do
 
     store = Handlers.on_tick(store, time)
 
-    :telemetry.execute([:sync, :store], %{slot: Store.get_current_slot(store)})
+    :telemetry.execute([:sync, :store], %{slot: get_current_slot(store)})
     :telemetry.execute([:sync, :on_block], %{slot: head_slot})
 
     Metrics.block_status(head_root, head_slot, :transitioned)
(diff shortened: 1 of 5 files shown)


rodrigo-o:
After various tests, I've spotted no more warnings related to blocks being received a couple of milliseconds before. while still seeing them in the main branch, specially at different times of the day.

Arkenan (approved):
LGTM. I think we should simplify the current_slot logic. There's at least three different ways to calculate it and it's not clear when to use which. Out of scope of this PR though.

rodrigo-o:
> LGTM. I think we should simplify the current_slot logic. There's at least three different ways to calculate it and it's not clear when to use which. Out of scope of this PR though.

Completely agree, now at least the 2 more used functions are in the same module but we still need to unify them, I'll add an issue!

rodrigo-o:
> > LGTM. I think we should simplify the current_slot logic. There's at least three different ways to calculate it and it's not clear when to use which. Out of scope of this PR though.
> 
> Completely agree, now at least the 2 more used functions are in the same module but we still need to unify them, I'll add an issue!

#1322 was created!

Issue #1309, opened 2024-09-26
Root not found in tree error hangs the node (Sepolia long run)


rodrigo-o:
This is something that could happen along the node executions but is easily reproducible on the initial setup. After the initial sync and some time running (a couple of minutes) an initial pruning kicks in and for some reason it generates a slot in the future error, here is an example log:

```
INFO 17:43:12.001 [StateDb] Pruning started. slot=5963552
INFO 17:43:12.001 [BlobDb] Pruning started. slot=5832480
INFO 17:43:12.001 [BlockDb] Pruning started. slot=5963552
INFO 17:43:12.001 [BlobDb] Pruning finished. 0 blobs removed. 
INFO 17:43:12.003 [Libp2p] Slot transition slot=5963616
INFO 17:43:12.009 [BlockDb] Pruning finished. 52 blocks removed. 
INFO 17:43:12.019 [StateDb] Pruning finished. 52 states removed. 
INFO 17:43:12.684 [Gossip] Block received, block.slot: 5963616. 
INFO 17:43:24.002 [Libp2p] Slot transition slot=5963617
INFO 17:43:24.216 [Gossip] Block received, block.slot: 5963617. 
INFO 17:43:29.977 [Fork choice] Adding new block slot=5963613 root=0x3de..fc05
INFO 17:43:30.204 [Fork choice] Block processed. Recomputing head. 
ERROR 17:43:30.204 GenServer LambdaEthereumConsensus.Libp2pPort terminating
** (RuntimeError) Parent 14C35B27D611342466C990ABE85147AFCDA13DBE2D5C05C4B3155309CF94FFFD not found in tree
    (lambda_ethereum_consensus 0.1.0) lib/lambda_ethereum_consensus/fork_choice/simple_tree.ex:84: LambdaEthereumConsensus.ForkChoice.Simple.Tree.get_children!/2
    (lambda_ethereum_consensus 0.1.0) lib/types/store.ex:147: Types.Store.get_children/2
    (lambda_ethereum_consensus 0.1.0) lib/lambda_ethereum_consensus/fork_choice/head.ex:90: LambdaEthereumConsensus.ForkChoice.Head.filter_block_tree/4
    (lambda_ethereum_consensus 0.1.0) lib/lambda_ethereum_consensus/fork_choice/head.ex:85: LambdaEthereumConsensus.ForkChoice.Head.get_filtered_block_tree/1
[...]

PR #1318, opened 2024-10-03
fix: base not found in tree

rodrigo-o:
**Motivation**

After a new finalized checkpoint we are having some issues related to base root not found in our fork choice tree (roots of blocks already recieved and showed as added in the logs)

**Description**
The issue that caused the base not found in the Fork choice tree was related to a bug in the PendingBlocks module. The bug made that when we had already processed a block, if some of its blobs are recieved later then the block is processed again, moved first to :pending (without checking its precvious status) and effectively being again the top of the fork choice. Then future blocks checked that their parents were downloaded (they were even processed) but none of them updated the head. At the moment of finalizing the tree the missing slots (the slot reprocessed +1 until the next epoch) are not part of the tree even if the state trantition moved on amd that caused the:
`Parent 14C35B27D611342466C990ABE85147AFCDA13DBE2D5C05C4B3155309CF94FFFD not found in tree`

To solve the issue a check was added at the time of blob processing, and just the blocks on `:download_blobs` status are taken into account.

**Possible things to investigate further:**
- Why did the fork choice preferred the old block root as the head? shouldn't it prefer their children's? Reprocessing the same block shouldn't happen in the first place but in case of happening why did it re selected an old root as the head when the block was the same as before (not a reorg)

_Note: I already merged both `sync-issues-on-initial-node-run` and `current-slot-with-maximum-disparity` to be in a better position regarding tests._

Resolvess #1309


Arkenan (approved):
LGTM

PR #1324, opened 2024-10-08
feat: assertoor implementation support

### LINKED ISSUE #1323: Integrate Assertoor as part of the CI
Before starting tackling the performance optimizations we need an easy way to check for regressions, assetoor will fulfill that role. We need to have a battery of assertions running as part of our CI. First of all we need to investigate a bit about it and it's integration, later we could split this issue into multiple ones.

Repo: https://github.com/ethpandaops/assertoor

rodrigo-o:
**Motivation**

Assertoor will allow us to have better regression checks when we tackle performance refactors

**Description**

Endpoints that needed fix/implementation for assertoor to work:
- [x] `node/health` - fixed
- [x] `node/syncing` - added (semi-stub)
- [x] `beacon/genesis` - fixed
- [x] `configs/spec` - added (semi-stub)
- [x] `headers/head` - added (semi-stub)
- [x] `states/head/finality_checkpoints`- fixed
- [x]  `events` - This was implemented in #1365

Closes #1323


lib/beacon_api/controllers/error_controller.ex
```diff
@@ -13,6 +16,8 @@ defmodule BeaconApi.ErrorController do
 
   @spec not_found(Plug.Conn.t(), any) :: Plug.Conn.t()
   def not_found(conn, _params) do
+    Logger.error("Not found resource, path: #{conn.request_path}")
```
ElFantasma, line 19:
```suggestion
    Logger.error("Resource not found, path: #{conn.request_path}")
```

lib/lambda_ethereum_consensus/fork_choice/fork_choice.ex
```diff
@@ -125,8 +125,7 @@ defmodule LambdaEthereumConsensus.ForkChoice do
   @doc """
   Get the current chain slot based on the system time.
 
-  There are just 2 uses of this function outside this module:
-   - At the begining of SyncBlocks.run/1 function, to get the head slot
+  There are just 1 use of this function outside this module:
```
ElFantasma, line 128:
```suggestion
  There is just 1 use of this function outside this module:
```

ElFantasma (approved):
LGTM, just a couple of suggestions on English redaction

PR #1365, opened 2025-01-21
feat: sse implementation for events beacon API

rodrigo-o:
**Motivation**

The events endpoint and specifically the `finalized_checkpoint` and `block` topics are needed to integrate with assertoor

**Description**

This will be a quck implementation, adding support for SSE and the current assertoor need from the api `eth/v1/events?topics=finalized_checkpoint, block`

Completes part of #1369


ElFantasma (approved):
LGTM!

Issue #1398, opened 2025-03-18
Generic SSZ encode/decode issues on general 1.5.0-beta.3 spec tests


rodrigo-o:
This appears to be unrelated to the electra update and instead being something now tested in the phase0 general spec tests that fails on a particular parameter ([spectest version 1.5.0-beta.3](https://github.com/ethereum/consensus-spec-tests/tree/v1.5.0-beta.3)):

```
1) test c:general f:phase0 r:ssz_generic h:containers s:valid -> ComplexTestStruct_nil_8 (General.Phase0.SszGenericTest)
     test/generated/general/phase0/ssz_generic.exs:9689
     Assertion with == failed
     code:  assert deserialized == real_deserialized
     left:  %Helpers.SszStaticContainers.ComplexTestStruct{
              A: 9027,
              B: [],
              C: 140,
              D: [],
              E: %Helpers.SszStaticContainers.VarTestStruct{A: 54691, B: [], C: 202},
    [...]
    right: %Helpers.SszStaticContainers.ComplexTestStruct{
              A: 9027,
              B: [],
              C: 140,
              D: [0],
              E: %Helpers.SszStaticContainers.VarTestStruct{A: 54691, B: [], C: 202},
```

This can be easily replicated on the [electra-support branch](https://github.com/lambdaclass/lambda_ethereum_consensus/tree/electra-support) running ` make spec-test-general-*`. 

14 test fail equally on that suite and they all have the same difference: `[]` instead of `[0]`.

rodrigo-o:
Closed in #1400

PR #1400, opened 2025-03-21
feat: electra container and ssz support

### LINKED ISSUE #1405: SSZ errors `OffsetSkipsVariableBytes(NNNN)` on all electra spec-test
After the initial configuration of the electra-support branch, all spec test fail with the same error `OffsetSkipsVariableBytes(NNNN)`, before starting tackling them one by one we need to get rid of the common issues, in this case this is related to the old schema for Containers, we need to update our container implementation with the changes introduced in electra.

rodrigo-o:
**Motivation**

This PR adds all new containers and modify the already existing one with the new [electra spec](https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/beacon-chain.md#containers)

**Description**

To be able to add new types we need to add them in 2 different places, one in the elixir code and another one in the ssz_nif crate. Also, apart from the container implementations we needed to add the constants and preset to be able to configure them.

This PR address an issue that happened on all spec-test outside of the general suite, where everyone failed with `OffsetSkipsVariableBytes` (#1405). Now, all those fail with different issues which enables us to tackle them separately.

Also after finding out that the config/preset in the rust ssz_nif is fixed instead of taking from a yaml (as done in elixir) I created an issue for reworking it, this is something that is too error prone to maintain as it is but probably something to tackle after the electra implementation: #1402

We also fixed a couple of small ssz issues spread around, one was related to an incorrect mapping of the "0x" literal string (#1398) and another one that happened specificaly in the random test suite (#1403). 

**Next Steps**

Before star tackling specific electra task we might want to tackle #1407 & #1408 to make sure all our unit test still run before starting to tackle the logic of the electra additions. We also want to make sure both Docker and kurtosis work as expected #1412 and at least have a simple version of assertoor in place #1413 although it probably is not complete until the actual fork updates are made to the code.

Closes #1405 & #1403 & #1398


lib/lambda_ethereum_consensus/validator/validator.ex
```diff
@@ -175,7 +175,8 @@ defmodule LambdaEthereumConsensus.Validator do
     %Attestation{
       data: attestation_data,
       aggregation_bits: bits,
-      signature: signature
+      signature: signature,
+      committee_bits: BitList.zero(committee_length) |> BitList.set(committee_index)
```
ElFantasma, line 179: Shouldn't this be?
```suggestion
      committee_bits: BitList.zero(committee_length) |> BitList.set(index_in_committee)
```

lib/types/beacon_chain/attestation.ex
```diff
@@ -13,33 +13,43 @@ defmodule Types.Attestation do
...
 
   @type t :: %__MODULE__{
-          # MAX_VALIDATORS_PER_COMMITTEE
+          # [Modified in Electra:EIP7549]
           aggregation_bits: Types.bitlist(),
           data: Types.AttestationData.t(),
-          signature: Types.bls_signature()
+          signature: Types.bls_signature(),
+          # [New in Electra:EIP7549]
+          committee_bits: Types.bitlist()
```
ElFantasma, line 29: According to the [spec](https://eips.ethereum.org/EIPS/eip-7549) the `committee_bits` type should be bitvector, not bitlist.

rodrigo-o: Great catch! this was exactly the issue that generated #1403. It's already fixed!

lib/lambda_ethereum_consensus/validator/validator.ex
rodrigo-o: I tried to give it a quick look but i guess is better for now to just add zeros, add a comment and later correctly implement it when we start tackling this particular EIP. This PR just adds/modifys the containers we should deal with the logic in future ones.

test/fixtures/validator/proposer/beacon_state.ssz_snappy
rodrigo-o, line 1: This file was needed to be replaced by one of the current electra spec-test

test/spec/utils.ex
```diff
@@ -35,7 +35,7 @@ defmodule SpecTestUtils do
   end
 
   def sanitize_yaml({k, v}), do: {String.to_atom(k), sanitize_yaml(v)}
-  def sanitize_yaml("0x"), do: <<0>>
+  def sanitize_yaml("0x"), do: []
```
rodrigo-o, line 38: This was the root cause of #1398, `"0x"` represent the empty byte array, i.e. `[]`, not `[0]` as previously mapped. [Here](https://github.com/ethereum/consensus-specs/blob/dev/ssz/simple-serialize.md#json-mapping) is the specific spec section

test/fixtures/validator/proposer/beacon_state.ssz_snappy
rodrigo-o: This change reduce our unit test failures from 13 to 4

test/fixtures/validator/proposer/empty_signed_beacon_block.ssz_snappy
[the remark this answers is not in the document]
rodrigo-o: Done: #1407

test/fixtures/validator/proposer/beacon_state.ssz_snappy
rodrigo-o: Here is an issue tracking the still failing test: #1408

native/ssz_nif/src/ssz_types/config.rs
```diff
@@ -181,6 +214,17 @@ impl Config for Gnosis {
...
+    // Electra added fields
+    type MaxConsolidationRequestsPerPayload = U2;
+    type MaxDepositRequestsPerPayload = U8192;
+    type MaxWithdrawalRequestsPerPayload = U16;
+    type PendingDepositsLimit = U134217728;
+    type PendingPartialWithdrawalsLimit = U134217728;
+    type PendingConsolidationsLimit = U262144;
+    type MaxAttesterSlashingsElectra = U1;
+    type MaxAttestationsElectra = U8;
+    type MaxValidatorsPerCommitteePerMaxCommitteesPerSlot = U131072; // 2048 * 64, this as the rest is fixed and we need to be really carefull about any change
```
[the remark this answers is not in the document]
ElFantasma: Can't it be `MaxValidatorsPerSlot` or `TotalValidatorsPerSlot` or something more meaningful?

ElFantasma (approved):
It looks good. Left only a comment regarding a type name

native/ssz_nif/src/ssz_types/config.rs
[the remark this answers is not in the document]
rodrigo-o: Agree its awfull it was needed to be created because of the way the configs are managed in the rust nif (In elixir we have them multiplied). I preferred to leave it as bad to make sure it's noted maybe Times instead of the middle Per is better. In any case all this need to be reworked as commented in this issue: #1402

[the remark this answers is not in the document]
rodrigo-o: Done [here](https://github.com/lambdaclass/lambda_ethereum_consensus/pull/1400#:~:text=a%20type%20name-,Changed%20name%20for%20the%20newly%20created%20MaxValidatorPerSlot%20config,-Verified)

PR #1411, opened 2025-03-31
feat: checkpoint-sync-url now can be a list of checkpoint nodes

### LINKED ISSUE #1085: Allow multiple checkpoint sync urls
We should fetch a block from each one, and then compare them, failing if they aren't equal.

LeanSerra:
**Motivation**

We want the ability to download the state from multiple checkpoint nodes when performing a checkpoint sync then compare the downloaded states failing if any of them are different

**Description**

- The flag --checkpoint-sync-url now parses a comma separated list of urls then fetch the state from each node finally compare them if any of them is different to the rest the program exits.

**Note**
- When testing most checkpoint nodes returned error 500 when calling `/eth/v1/beacon/deposit_snapshot` as this endpoint will be deprecated in the Electra fork. For this reason I didn't add the comparison when downloading the snapshot as it was hard to find two nodes that agreed in the result.

**How to try it**

```Shell
rm -rf level_db/mainnet
```

```Shell
iex -S mix run -- --checkpoint-sync-url "https://mainnet-checkpoint-sync.stakely.io, https://sync.invis.tools"  --discovery-port 9009 --mode full
```

Closes #1085


lib/lambda_ethereum_consensus/beacon/store_setup.ex
```diff
@@ -129,28 +133,66 @@ defmodule LambdaEthereumConsensus.Beacon.StoreSetup do
...
+      System.halt(1)
+    end
+
+    # We already checked block and state match
+    {:ok, store} = Store.get_forkchoice_store(anchor_state, anchor_block)
+
+    # Save store in DB
+    StoreDb.persist_store(store)
+
+    store
```
rodrigo-o, line 179: I'm not fan of Halts, but given that it was already used in the previous state of the function we could just maintain it here, in any case i think it's better to go for the positive and halt in the else instead of simulating and early return.

```elixir
    if anchor_state.latest_block_header.parent_root == first_block.message.parent_root do
      {:ok, store} = Store.get_forkchoice_store(anchor_state, anchor_block)
  
      # Save store in DB
      StoreDb.persist_store(store)
  
      store
    else
      Logger.error(
        "[Checkpoint sync] Root mismatch when comparing latest finalized block with downloaded state"
      )

      Logger.flush()
      System.halt(1)
    end
```

rodrigo-o (approved):
Just a small comment

lib/lambda_ethereum_consensus/beacon/store_setup.ex
LeanSerra: Done [here](https://github.com/lambdaclass/lambda_ethereum_consensus/pull/1411/commits/e092fd140d226ed049fa7702039ee5c23a938ad4)

PR #1416, opened 2025-04-03
fix: change tools versions with a correct version of go

rodrigo-o:
**Motivation**

In a previous PR the go version was set to 1.22, but it lacked the patch version resulting in a 404 while trying to install. Protobuff was also updated.


LeanSerra (approved):
Nice catch!

PR #1423, opened 2025-04-09
docs: add an electra implementation gap doc

rodrigo-o:
**Motivation**

Create an initial md where we could track the ongoing effort

**Description**

I brought in the electra spec mds from https://github.com/ethereum/consensus-specs/tree/dev/specs/electra. Also created the electra-gap.md for tracking the implementation of electra linking both at PRs and the spec, this will probably become an issue as well.


LeanSerra (approved):
Nice! 🚀

PR #1427, opened 2025-04-14
fix: skipped spec tests in electra

rodrigo-o:
**Motivation**

We were skipping 8k spec-test before this PR

**Description**

After some debugging and testing, most of the previously skipped test are not needed to be skipped at all. We are right now at 784 spect-test skipped which is mostly the same number as we are currently skipping in main.

This nonetheless could include some false positives and it's possible tat lots of this test share low-hanging fruit fixes apart from the electra implementation. I created a new issue for tracking that: #1429

Before this PR:
`11370 tests, 206 failures, 8092 skipped`

After this PR:
`11370 tests, 2003 failures, 784 skipped`

### DIFF
--- a/test/spec/runners/bls.ex
+++ b/test/spec/runners/bls.ex
@@ -6,21 +6,11 @@ defmodule BlsTestRunner do
   use ExUnit.CaseTemplate
   use TestRunner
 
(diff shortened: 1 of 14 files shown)


LeanSerra (approved):
Awesome!

PR #1430, opened 2025-04-15
fix: new ssz fixes for the static spec-test

### LINKED ISSUE #1429: Validate the newly un-skipped spec-test in electra
After #1427 we have nearly 1800 new failing spec-test, it's possible that some of them are just false positives (they were failing before electra) or that some small fixed could swipe up lots of them (specially outside the needed electra implementation for requests and withdrawals)

rodrigo-o:
**Motivation**

After #1427 nearly 1500 test were failing due to ssz issues

**Description**

This PR makes a small changing adding the new containers to the schema_match macro that were missing them.

Before this PR: 
`11370 tests, 2003 failures, 784 skipped`

After this PR:
`11370 tests, 979 failures, 784 skipped`

Resolves part of #1429

### DIFF
--- a/native/ssz_nif/src/utils/mod.rs
+++ b/native/ssz_nif/src/utils/mod.rs
@@ -53,6 +53,14 @@ macro_rules! schema_match {
                 Epoch,
                 BlobSidecar<C>,
                 BlobIdentifier,
+                PendingDeposit,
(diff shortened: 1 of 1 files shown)


LeanSerra (approved):
🚀

PR #1431, opened 2025-04-15
feat: electra withdrawals

LeanSerra:
**Motivation**

Implement changes in Electra to withdrawals + fix all spec-tests related to it.

**Description**

- Modified `get_expected_withdrawals` [link](https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/beacon-chain.md#modified-get_expected_withdrawals)
- Modified `process_withdrawals` [link](https://github.com/ethereum/consensus-specs/blob/dev/specs/electra/beacon-chain.md#modified-process_withdrawals)

**Fixed Spec Tests**

```
test/generated/mainnet/electra/operations.exs:
previous: 333 tests, 19 failures
now: 333 tests, 0 failures

test/generated/mainnet/electra/sanity.exs:
previous: 97 tests, 8 failures, 1 skipped
now: 97 tests, 2 failures, 1 skipped

test/generated/minimal/electra/operations.exs:
previous: 366 tests, 38 failures
now: 366 tests, 0 failures

test/generated/minimal/electra/sanity.exs:
previous: 104 tests, 8 failures, 1 skipped
now: 104 tests, 2 failures, 1 skipped

tests fixed: 69
```

**Spec Test Progress**

`11370 tests, 4 failures, 784 skipped`


lib/lambda_ethereum_consensus/state_transition/operations.ex
```diff
@@ -355,49 +362,146 @@ defmodule LambdaEthereumConsensus.StateTransition.Operations do
...
+         withdrawals
+       ) do
+    epoch = Accessors.get_current_epoch(state)
+
+    max_pending_partials_per_withdrawals_sweep =
+      ChainSpec.get("MAX_PENDING_PARTIALS_PER_WITHDRAWALS_SWEEP")
+
+    if withdrawal.withdrawable_epoch > epoch ||
+         processed_partial_withdrawals_count == max_pending_partials_per_withdrawals_sweep do
+      {:halt, {processed_partial_withdrawals_count, withdrawal_index, withdrawals}}
```
rodrigo-o, line 454: This is mostly a question I have, they are ordered right? specially for the `withdrawal.withdrawable_epoch > epoch` because that just make sense to halt if you have them ordered by epoch, or am I missing something?

Btw I checked that this was the correct condition to break the loop in the spec but i was intrigued about the order.

lib/lambda_ethereum_consensus/state_transition/operations.ex
```diff
@@ -1030,7 +1134,7 @@ defmodule LambdaEthereumConsensus.StateTransition.Operations do
 
     is_correct_source_address =
       case validator.withdrawal_credentials do
-        <<_::binary-size(12), rest>> -> rest == address
+        <<_::binary-size(12), validator_address::binary>> -> validator_address == address
```
rodrigo-o, line 1137: 👍

rodrigo-o (approved):
LGTM, just some comments for the future and one question

lib/lambda_ethereum_consensus/state_transition/operations.ex
LeanSerra: That's a good question, I don't explicitly order them but I've yet to see a case in which they aren't not in the spec tests and not while syncing Sepolia. I can add that or maybe leave a comment so that if we run into this issue later in testing we can fix it

rodrigo-o: Maybe there is no way to get them unordered but yes, I think we should left at least a comment for the future (but probably where we actually add the pending widthrawal instead of here)

PR #1432, opened 2025-04-15
fix: outstanding ssz issues after #1430

rodrigo-o:
**Motivation**

There are still 492 ssz related test failing in the spec test suite after #1430.

**Description**

- [x] 123 from BlobSidecar: The issue was related to a config in the minimal preset (`KZG_COMMITMENT_INCLUSION_PROOF_DEPTH`) which was out of sync since deneb.
- [x] 123 from BeaconBlockBody: The issue was related to a config in minimal also (`MAX_BLOB_COMMITMENTS_PER_BLOCK`) which was out of sync since deneb.
- [x] 123 from BeaconBlock: This failed because of BeaconBlockBody
- [x] 123 from SignedBeaconBlock: This also failed because of BeaconBlockBody

A couple of other test have been also fixed by this changes

Before this PR:
`11370 tests, 979 failures, 784 skipped`

After this PR:
`11370 tests, 424 failures, 784 skipped`

Closes #1429

PR #1433, opened 2025-04-16
fix: fixing the CI updating ubuntu and go

rodrigo-o:
**Motivation**

Most of CI steps where failing unable to fetch the protobuf-compiler

**Description**

All go version has been updated to 1.24 and all Ubuntu 22.04 to 24.04 (LTS)

### DIFF
--- a/test/unit/validator/block_builder_test.exs
+++ b/test/unit/validator/block_builder_test.exs
@@ -85,7 +85,7 @@ defmodule Unit.Validator.BlockBuilderTest do
     commitment_root = SszEx.hash_tree_root!(commitment, TypeAliases.kzg_commitment())
 
     # Manually computed generalized index of the commitment in the body
-    index = 0b101100000
+    index = 0b1011000000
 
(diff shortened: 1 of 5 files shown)


test/unit/validator/block_builder_test.exs
```diff
@@ -80,7 +80,7 @@ defmodule Unit.Validator.BlockBuilderTest do
 
     [proof] = BlockBuilder.compute_inclusion_proofs(body)
 
-    assert length(proof) == 9
+    assert length(proof) == 10
```
rodrigo-o, line 83: This was changed in #1432, the change was modifying `KZG_COMMITMENT_INCLUSION_PROOF_DEPTH` from9 to 10 as per the [updated spec](https://github.com/ethereum/consensus-specs/blob/8e92b52369fe66b01a8096dc2fb24835885dd494/presets/minimal/deneb.yaml#L11)

LeanSerra (approved):

PR #1443, opened 2025-04-23
fix PendingBlocks.process_blobs returns {:ok, store} we only need to …

LeanSerra:
…return store

**Motivation**

**Description**

Closes #issue_number


lib/lambda_ethereum_consensus/p2p/gossip/blob_sidecar.ex
```diff
@@ -19,7 +19,7 @@ defmodule LambdaEthereumConsensus.P2P.Gossip.BlobSideCar do
       Logger.debug("[Gossip] Blob sidecar received, with index #{blob_index}")
       Libp2pPort.validate_message(msg_id, :accept)
       # TODO: (#1406) Enhance the API to reduce unnecessary wrappers (:ok + list)
-      PendingBlocks.process_blobs(store, {:ok, [blob]})
+      PendingBlocks.process_blobs(store, {:ok, [blob]}) |> then(&elem(&1, 1))
```
rodrigo-o, line 22: Couldn't this return an `{:error, reason}` in which case we'll be returning the reason here?

rodrigo-o (approved):

LeanSerra:
Left #1444 as an issue  to add tests for this so we don't have regressions again

PR #1449, opened 2025-04-25
do not merge: Electra sync forced with Sepolia

LeanSerra:
**Motivation**

Sync Sepolia 

**Description**

- Comment out the blob validations
- When we get a block with missing blobs we schedule the download but process the block anyways
- We were constantly penalizing peers so that was also commented out

**Findings**

- The peers are constantly being penalized resulting in running out of peers
- After checkpoint-sync when downloading blocks the chain wouldn't advance because the next block was stuck in :download_blobs
- After commenting the blob validation I run into an issue only once in several tries:
   - The chain was advancing but one block failed because it was from "the future" which resulted in that block having the status of :invalid when it wasn't the check for ForkChoice.future_slot? in `process_block` tries to fix this but its not really tested


rodrigo-o:
This will be closed until we start delving in Sepolia syncing again.

PR #1460, opened 2025-06-10
Fix broken spec link for get_validator_from_deposit in Electra gap document

VolodymyrBg:
Corrected the documentation link for the "Modified get_validator_from_deposit" function in electra-gap.md to point to the existing docs/specs/electra/beacon-chain.md file and the correct section. This resolves a broken link issue and ensures the reference is accurate for future readers and contributors. No other content was changed.


rodrigo-o:
Sorry, but we don't accept typo contributions

